Issue #9, opened 2015-12-30
OpenBve.csproj on Linux/Mono


JakubVanek:
Hi,
I have this problem: vanilla OpenBve.csproj is not working with xbuild. I'm going to create new PR for solution of other problems but I have problem with this:
```xml
  <PropertyGroup>
    <PreBuildEvent>set textTemplatingPath="%25CommonProgramFiles(x86)%25\Microsoft Shared\TextTemplating\$(VisualStudioVersion)\texttransform.exe"
if %25textTemplatingPath%25=="\Microsoft Shared\TextTemplating\$(VisualStudioVersion)\texttransform.exe" set textTemplatingPath="%25CommonProgramFiles%25\Microsoft Shared\TextTemplating\$(VisualStudioVersion)\texttransform.exe"
%25textTemplatingPath%25 "$(ProjectDir)\Properties\AssemblyInfo.tt"</PreBuildEvent>
  </PropertyGroup>
```
TextTransform.exe is provided by MonoDevelop. It *usually* resides at ```/usr/lib/monodevelop/AddIns/MonoDevelop.TextTemplating/TextTransform.exe``` and I don't know how to make the path configurable. Switching between OSes shouldn't be a problem, see [Mono Wiki](http://www.mono-project.com/archived/porting_msbuild_projects_to_xbuild/). Could you please help me?
Thanks

JakubVanek:
Solved with global property.

PR #10, opened 2015-12-30
Fix OpenBve.csproj on xbuild

JakubVanek:
Related to issue #9 (path isn't configurable yet). Please test it on Windows/Visual Studio because I'm afraid it could discard this hack.

### DIFF
--- a/openBVE/OpenBve/OpenBve.csproj
+++ b/openBVE/OpenBve/OpenBve.csproj
@@ -12,7 +12,7 @@
     <AssemblyName>OpenBve</AssemblyName>
     <NoStdLib>False</NoStdLib>
     <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
-    <AllowUnsafeBlocks>False</AllowUnsafeBlocks>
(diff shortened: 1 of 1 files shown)


leezer3:
In some ways, the xbuild text-transform event might actually want to be dumped altogether.

The only thing this text-transform does is to neatly bump the version number whenever a build is made, which saves me a little housekeeping.
Keeping it isn't the end of the world, it's just neat :)

Both of your changes work successfully in VS2013.

leezer3:
Also, I'd have thought that TextTransform.exe should be findable via $PATH under *Nix systems.

It's a standard enough VS component that I'm a little surprised that the Monodevelop package didn't add it to $PATH itself.

I suppose it'd be possible to script finding it/ adding to $PATH, but a simple instruction in the readme would probably suffice.

JakubVanek:
I solved it by introducing new global property. Now xbuild will try $(TextTransformMono) if it's defined, if not, it tries the default path. Property can be set by simply appending /p:TextTransformMono=[path] to xbuild.

Issue #8, opened 2015-12-30
Key release not handled


JakubVanek:
See this video: https://www.youtube.com/watch?v=g7XV7cUfhW8
I just look around using arrow keys. Then I keep the right arrow pressed a bit longer and the camera starts rotating on its own (all physical keys are released). The same problem repeats with F1 at the end of the video (info flood). I don't know whether this is a bug in openBVE, OpenTK or the X11/SDL2 backend but it makes the game unplayable for me.
OS: Xubuntu 14.04.3 LTS
OpenTK version: 1.1.2188.6217
openBVE up-to-date at master branch

JakubVanek:
The same problem exists with OpenTK 1.1-4c (downloaded from releases).

JakubVanek:
Well, I think something like this happened to me too before I moved my repo to SDL2. You must somehow get keyUp and keyDown events working correctly with repeat. From actual source it looks to me that there are two key repeat mechanisms: that inside OpenTK and that from SDL 1.2 era.

JakubVanek:
I hacked MainLoop.keyUpEvent and MainLoop.keyDownEvent to provide debugging info and this is the result. Number in brackets is millisecond offset from first log entry.
```
Key log:
[0] KeyPress:	[Key:Right,IsRepeat:True,Modifiers:0]
[492] KeyPress:	[Key:Right,IsRepeat:True,Modifiers:0]
[535] KeyPress:	[Key:Right,IsRepeat:True,Modifiers:0]
[581] KeyPress:	[Key:Right,IsRepeat:True,Modifiers:0]
[619] KeyPress:	[Key:Right,IsRepeat:True,Modifiers:0]
[699] KeyPress:	[Key:Right,IsRepeat:True,Modifiers:0]
[740] KeyPress:	[Key:Right,IsRepeat:True,Modifiers:0]
[777] KeyPress:	[Key:Right,IsRepeat:True,Modifiers:0]
[820] KeyPress:	[Key:Right,IsRepeat:True,Modifiers:0]
[901] KeyPress:	[Key:Right,IsRepeat:True,Modifiers:0]
[939] KeyPress:	[Key:Right,IsRepeat:True,Modifiers:0]
[982] KeyPress:	[Key:Right,IsRepeat:True,Modifiers:0]
[1066] KeyPress:	[Key:Right,IsRepeat:True,Modifiers:0]
[1103] KeyPress:	[Key:Right,IsRepeat:True,Modifiers:0]
[...]

PR #12, opened 2015-12-31
Fix infinite key repeating

JakubVanek:
Solves issue #8 .


leezer3:
Dumping  # already and using a return statement in the for loop would seem to be better to me?

JakubVanek:
Yeah, sorry, I sometimes complicate things too much.

PR #13, opened 2015-12-31
Fix translation issue in TrainManager

JakubVanek:
Hi,
this issue came to my eyes when I was breaking the speed limit and I crossed the red signal light. Instead of suggesting me to proceed at 25 km/h, it suggested me that I should proceed at 25 [unit].
Before:
screenshot - 31 12 2015 - 15 53 01
After:
screenshot - 31 12 2015 - 15 48 01


leezer3:
Not quite fully fixed.
This commit fixes it fully:
https://github.com/leezer3/OpenBVE/commit/e166423594f25e6d4dc709946ddc26c0dc325634

The specifiable unit of speed also requires multiplication of the base speed value by the conversion factor (Game.SpeedConversionFactor), as otherwise with a custom speed set unit set, the speed displayed will be wrong. 

This also brings up something else not implemented yet- The review screen  doesn't use custom speed units:
https://github.com/leezer3/OpenBVE/issues/14

PR #17, opened 2016-01-16
Depends included intro workdir

v1993:
I created version include all depends of OpenBVE. I also create patch. I want to you create branch with name like 'standalone'. Or, if you don't do it, I can patch all new commits and support my fork.
I can't attach patch, because it's too big.
Signed-off-by: V-Mint <v19930312@gmail.com>


JakubVanek:
What about licenses? Are they required to be distributed with binaries?

leezer3:
IMHO dropping binaries into the main build tree isn't ideal, although having said that I note that the paths in the solution are currently absolute paths for my system, rather than relative to the build tree.
That ought to be changed, will try and deal with that sometime this weekend.

OpenTK version 1.1.2204.6440 (As of this commit: https://github.com/opentk/opentk/commit/b0dbb7de7fb329e44b6a0ba2f4cadc1ea12b69c9 ) is also a 'requirement', as otherwise certain keymapping combinations don't work correctly.

Final thoughts:
The licence files for OpenTK and CS-Script are currently in the root of the build tree. I'd forgotten Sharpcompress (Still playing with file extraction....),  so need to add that.
I think we probably need a window to popup the 3rd party licence texts too.

v1993:
OK, I updated repo from main branch (leezer3 repo), openTK and added SharpCompress license.

leezer3:
The above commit does basically what you were trying to do, but is IMHO a little neater.

We don't require OpenTK.GLControl.dll and the folder structure doesn't really require all the nuget stuff/ version numbers.

v1993:
OK, I removed my fork.

leezer3:
No problems!
It's much easier to find mess-ups when someone else tries to do something you didn't expect.

PR #21, opened 2016-02-15
Improve .gitignore

mgavioli:
Improve .gitignore:
- using a single top-level file
- ignoring some of the auto-generated files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 # git ignore file
+*.userprefs
 bin/
 obj/
--- a/ObjectViewer/.gitignore
+++ b/ObjectViewer/.gitignore
@@ -1,3 +1,0 @@
-# git ignore file
-bin/
-obj/
--- a/RouteViewer/.gitignore
+++ b/RouteViewer/.gitignore
@@ -1,3 +1,0 @@
-# git ignore file
-bin/
-obj/
--- a/TrainEditor/.gitignore
+++ b/TrainEditor/.gitignore
@@ -1,3 +1,0 @@
-# git ignore file
-bin/
-obj/
--- a/openBVE/OpenBve/.gitignore
+++ b/openBVE/OpenBve/.gitignore
@@ -1,3 +1,0 @@
-# git ignore file
-bin/
-obj/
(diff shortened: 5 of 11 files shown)


leezer3:
The single top-level .gitignore file is a good shout.

The AssemblyInfo.cs file is a little bit of a bodge on my part. What I would have liked to do is to increment the version number on merged commit only, but I couldn't get this to work properly, so I settled on incrementing the version automatically every time the solution was built.
This isn't the best idea in the world, as if someone else builds the same set of source code, they'll get a different version number, but it was the best I could think of.....

FYI, the current version number works like this:
1- Base version
4- Release version
36- Daily build (This increments by one on each day on which a build is made)
782- Revision number (Number of minutes since midnight)

mgavioli:
Not sure to understand: AssemblyInfo.cs is an auto-generated file, isn't it?

1) Does the generation of the 'next' iteration of AssemblyInfo.cs takes into account the contents of the 'previous' one?

2) Are you suggesting to remove AssemblyInfo.cs from .gitignore? I noticed this usually generates modified AssemblyInfo.cs in branches, with possible merge conflicts.

For what is worth, with MuseScore (another large open source project I participate to), we attempted several times to have some automatic generation of meaningful version numbers and eventually we gave up, simply using the git branch SHA.

As soon as this is sorted out, I'll fix the merge conflicts.

leezer3:
It's generated by this script, which is called as a pre-build event:
https://github.com/leezer3/OpenBVE/blob/master/openBVE/OpenBve/Properties/AssemblyInfo.tt

You should be able to read it easily enough, but in essence, it reads the first line of the file to determine the current version number, and uses that as a base to generate the version number as above.

I suppose the version number could be changed to use the commit hash instead?
Could I *think* pull that with a Git hook.

mgavioli:
- Removed `AssemblyInfo.cs` from `.gitignore` (i.e. no longer ignored)
- Fixed merge conflicts

leezer3:
Added a few more bits to the .gitignore file here:
https://github.com/leezer3/OpenBVE/commit/9e43e95c0afb8f0d7e50787c3a05cee9944eb862

PR #22, opened 2016-02-21
Synchronize language files.

mgavioli:
- Adds a few missing strings for the in-game menu.
- Makes all files parallel with the same number of lines in the same place, easing future insertions.
- **Ukrainian**: fixed several strings which were split across multiple lines.
- **Italian**: corrected accented vowels, translated untranslated strings, corrected a few typos.


mgavioli:
I noticed that Italian does not use accented vowels, resorting to a very old way of marking accents (of typewriter derivation). German, or more exotic languages like Slavic or oriental, do use non-ASCII characters, encoding the files as UTF-8 with BOM.

Anything against 'upgrading' Italian to 21th century too? Any known problem with Unicode characters?

leezer3:
I've just implemented fallback language choices in this commit:
https://github.com/leezer3/OpenBVE/commit/58637c679dda6f5df139159e8b843400852e71e1

You should be able to remove the English lines from your translation now.

Unicode *should* work- 
The officially supported character encodings are as follows:
**Unicode:**
UTF-16BE, UTF-16LE, UTF-32BE, UTF-32LE 
**Other:**
SHIFT_JIS (Legacy of BVE1/2 Japanese content)

leezer3:
For reference, if you wish to specify specific fallback language(s) in addition to en-US, this should be done by creating a new section in the language as follows:
`[fallback]
language=de-DE`

leezer3:
Final thoughts-
I'd presume that UTF-8 is being upconverted to UTF-16 somewhere.

Accents definitely work in French here (French is the only language that I speak well enough to be confident that everything is correct), so I'd go ahead and alter the Italian language file as you see fit.

mgavioli:
Thanks for the work on languages. Still, I believe having all the language files parallel makes the life easier for translators, so I would leave the changes as they are.

Completed the Italian.

leezer3:
No problems.

Thinking about it, that's probably the best approach, but providing the fallback option means that even if an older language file is used, we should never display an internal string, which is good manners.

PR #23, opened 2016-02-22
Route-train loading screen

mgavioli:
Improves (hopefully!) the loading screen by:

- adding a background image randomly chosen from a pool of built-in images;
- reverting to a sort of 'official' openBVE logo;
- logo texture is provided in 3 sizes, all at powers of 2 dimensions for best rendering;
- displaying actual assembly version;
- temporarily removed the web site URL;
- displaying a progress bar (cumulative for route and train)

An SVG drawing of the logo for creating textures of appropriate sizes is included.

Sample screen shot:
loadingscreen_screenshot_2


leezer3:
There's no 'official' site as such, just at the moment-
There are various places hosting Odakyufan's last build, but to the best of my knowledge, nothing 
significant was ever changed in the source code.

The site you've linked to is Shadow93's-
https://github.com/shadow93/
I did reach out to him when I started doing major work, but never got anything back. The last commit on his OpenBVE repo was by Jakub Vanek, who has contributed some bits to this build.

The site link on the main menu for these builds links back to this Github, but that's it.
That probably ought to change at some stage I suppose.....

mgavioli:
I added an URL simply because one (dead!) was present in the original loading screen and I though an URL from which the commoner can download a ready-to-use executable, although outdated, was useful.

Possibly, the best compromise is simply do not put any URL at all. Commented out.

leezer3:
Very true!

I'd be inclined to leave the loading screen texture as fullscreen, but keep your aspect ratio fixes, which I'll admit I hadn't thought of.
The logo has been through various incarnations, and I *think* this was the last 'official' one supplied:
logo

Some of the different builds have used different logos, and one of my patches was to allow the setting of a custom loading-screen texture on a per-route basis:
https://github.com/leezer3/OpenBVE/commit/a9e657d4af58edcad5d3f0dd8979a534aefee555

mgavioli:
I think the question should be split in two: background proper (the train picture in your example) and logo (the "openbve" 'brand'). I believe they should be separate.

1) **Background texture**
A background texture would be a fine touch. If you have one you like, please tell me.

Not sure about route-dependent image, though: it is likely the image provided by the route designer would be a brightly coloured one and would make the overlaid logo and text harder to read.

2) **Logo**
According to my attempts, it is highly preferable that the logo be of power of 2 dimensions (not necessarily equal, but both width and height at powers of 2) and be of a fixed size, regardless of window(/screen) size. Otherwise OpenGL kicks in with its spells, turning a neat and crisp image into an ugly 'laddered', pixelated, blurred one (even anti-aliasing does not help so much in this case).

(Note: this also applies to the background texture, of course, but assuming it would be a lower detail, subdued image, it will be less evident).

Assuming an approx. ratio of 4 : 1 for the above logo, the texture should be 256 x 64 or 512 x 128, to fit a small (but still reasonable) windowed context.

If this logo is the best approximation of an 'official' one, I will try to turn it into an SVG, which I find very useful for creating crisp images of whatever required size.

3) **Other details**: version display, progress bar... Are they suitable?

Thanks!

leezer3:
The other details are just fine I think.
The loading bar will be a bit inaccurate, as there is no progress reported for the final step (Mesh positioning and optimization), but I think we can live with that.

With regards to background pictures versus logos etc.-
Perhaps a compromise, so that the background picture and overlay logo are shown, unless a custom load screen image is selected- I think the possibility is important as it allows for 'better' branding of routes.
For the stock screen, I'm tempted to do 4 or 5 background pictures, and show a random one.

The power of two rule is definitely important, as OpenGL only performs a basic resize, rather than a resample when resizing any texture supplied, hence the nasty artifacts.
GLU.ScaleImage() should have better quality, and might help in this case, but on the flip side, it's probably more trouble than it's worth.

mgavioli:
Experimenting... The screen shots below are taken from a WIP version using a number of different background images of different ratios and the 'official' logo you posted above.

I made up just two backgrounds (as I know you like steam engines, I stole two steam images from wikipedia, converted to grey and compressed toward the middle of the tonal range), with different ratios to check the results; one is 16:9, another is 1:1 (square). The image is scaled to fit the screen, but not stretched, to preserve its proportions: black padding below or at the sides is added to fill the screen. This also emulates what would happen with user route images, which may have any ratio.

The logo is overlaid on the background image and it is kept at fixed dimensions (currently 512x128 pixels).

The good new is that the OpenGL curses on the background go mostly unnoticed, as the images are rather subdued, while the logo keeps its crispness, being fixed at power of 2 sizes..

The bad new is that I have no idea what would happen with custom route images. Is there a route sporting one?

As you are probably more documented than I am, I would appreciate if you could locate 4 - 5 (or whatever) images you consider representative to use as background. I can process and add them. The number of backgrounds to choose from would be hard-coded at the moment.

Not shown in the screen shots, I plan to replace the cyan of the progress bar with the same yellow of the dot in the logo, to add a bright note. The same yellow could be used in the menu, as a 'branding' element.

Let me know if the whole makes sense to you!

16:9 background:
loadingscreen_screenshot_0

Square background:
loadingscreen_screenshot_1

leezer3:
Makes perfect sense.

I don't think that there's anything that's actually used custom route images or any of the new features at the moment-
I've very much been adding stuff that has been worked around to varying degrees of sucess in the past, and stuff that I want to make my life easier. 
I'm planning to post a version of my route demonstrating the new features at some stage, but that hasn't come to fruition yet.

I'd probably be inclined to change the black borders on images not matching the screen resolution to a light grey, assuming that greyscale images were what was supplied.
If you wanted to get fancy, you could workout the average color of the background image, and use that instead, but that's overkill!

Images-wise, I think what this wants is 4-5 different country/ engine types.
The first image above is a Japanese EMU of some description, which is very appropriate considering BVE's Japanese origins.
I've got plenty of British images ( https://www.flickr.com/photos/53871940@N08/  for a lot of the better ones), although I haven't got around to uploading anything recently......
Possibly wants something British, plus European and USA based?

Recognisable trains for the general public would probably be better, but exactly what those are is another thing entirely!

leezer3:
A sample of the type of custom loadscreen I was thinking of when I added the feature:
loadscreen

My intention is to use era-specific variants (Steam/ diesel)

mgavioli:
- Changed uncovered screen areas to medium grey
- Logo texture size depending upon screen width (3 sizes)
- Provided 7 background images of my like (just raided Wikipedia!); feel free to replace them or to point me to specific images you think more appropriate.

leezer3:
Looking really good, although I'd say center the image, as opposed to being rooted to the top?

Other than that (Not sure on it mind), I'd say this should be ready to merge?

mgavioli:
Not sure about centring the image vertically; I think to prefer it 'hanging' (vertical axis is psycologically different from horizontal axis, in which asymmetry needs strong reasons usually). I believe this to be easy to change later if users comment about.

Just FYI: the `loadingbkg_2.png` image, which would probably puzzle several (non-Italian) users, was the work-horse of the Italian Railways during my childhood; I clearly remember them occurring everywhere. So, it is a kind of memorabilia for me...

mgavioli:
Well, it was perhaps not very clear, but me too I consider this PR ready to merge, if there are no other concerns.

leezer3:
Merged.

I'm working on a basic XML based package format just at the minute, so haven't got quite as far as doing a build yet.....
https://github.com/leezer3/OpenBVE/tree/PackageInstaller

PR #24, opened 2016-02-22
In-game menu

mgavioli:
The in-game menu code, which was scattered among several places and mostly globally available static, is refactored into a separate class, as an attempt to encapsulate all the menu management within an appropriate container.

**Logic**
- Menu logic is moved from `ProcessControls.cs` to `Menu.cs`
- Only the relevant menu is built at each step, rather than the whole menu system at each invocation
- The concept of sub-menu is removed
- The `InterfaceType.CustomiseControl` enum item is removed and the in-game control customization is merged into the `Menu` class itself
- As a consequence, control customization code in `OldCode/MainLoop.cs` and in `System/Input/Keyboard.cs` is greatly reduced.

**Drawing**
- Menu drawing is also moved from `Overlays.cs` to `Menu.cs`
- Menus with commands are centred and menus with lists (stations, controls) are left-aligned
- Added a background under the menu as a whole
- The size of the menu font varies according to the screen height
- The currently selected menu item has an 'openBVE yellow' background highlight with black text
- Captions are slightly dimmed (no longer cyan!)
- The menu for customization of a single control is more similar to the other menus.


leezer3:
Not entirely convinced by the solid black highlight around the text- Looks rather odd on a 2048px wide screen. I'd be inclined to leave it out, but perhaps add a further highlight around the menu itself, which it meets at the edge?

Another, separate thought on the colors-
It shouldn't be too difficult to implement color loading from the user options file. 
It's probably not worth providing an option for this, but a changeable font option in the main menu would be nice- Will add this to my list of things to think about.

mgavioli:
Although nice in principle, I would postpone **skin configurability** for the moment (and I have reserves on the skin configurability in itself). The menu uses whatever font is provided by the programme and would automatically benefit from it, once implemented. Colours are hard-coded, but factored out at the beginning of the source file, so they would be easy to adapt.

1) About the **black highlight**: possibly, it is a matter of taste, but if it looks odd to you, it would likely look odd to others too. My idea was to replicate the concept of highlighted list item common in many programmes, using a contrasting background and text colours (human eye is more sensible to contrast than to any other colour parameter). Black+yellow is the most contrasting pair and a rather sober choice, given the context.

1a) As the colours are declared at the beginning of the file, it would be relatively easy to experiment with different colour pairs: black on yellow, white on dark blue, black on light blue, whatever. If you have suggestions, they are welcome

1b) The black highlight currently uses the total menu width (+ some padding); in some cases the caption is definitely wider and may make the highlight unexpectedly wide. Skipping captions from the width reckoning would make the highlight as wide as the real menu options. I would avoid using an highlight as wide as the currently selected item: changing the highlight width from item to item would look odd to me; so I would still use an highlight as wide as the widest option (captions excluded).

2) "**a further highlight around the menu itself**": not sure to understand. Do you propose to add a background rectangle under the menu as a whole, as wide and high as the whole menu (+ some padding)?

For instance: a solid black menu background, a yellow selected text on a dark grey highlight (or black text on yellow highlight), white text for other items. Or whichever other colour combination makes sense / looks good.

In this case, I would also experiment with the alpha of overlay (the whole screen darkening of the scenery underneath): a less dark overlay might make the menu more prominently separated from the scenery, while still marking the latter as inactive.

3) **Everything else**: except for these visual elements, are the general concepts of the new code fitting the context?

Thanks!

leezer3:
The font family should be whichever generic sans-serif font family is provided by the system.

I think, although I'd have to check, that this is different to the default font provided by the Winforms menu implementation.

**a further highlight around the menu itself** 
That's the idea I had in mind. That way, on a large screen, the borders of the menu are clearly demarcated.

sample

This image has been downsized from 2560x1440, but you should be able to see that the width of the menu text is ~1/4 of the screen, and it looks a little odd.
Adding a highlight around the menu border gives a nice visual cue.

mgavioli:
Changes:

- background (black) under the menu as a whole
- font size now depends on screen height (choosing among the 5 sizes already provided by the programme and defined in `Graphics/Fonts.cs`)
- lighter overlay on whole screen (makes the menu more evident)
- selected item highlight with the 'openBVE yellow' (and black text)
- several code clean-ups.

A sample:
screenshot_menu_0

mgavioli:
Does anything remain open in this PR? Do you have more suggestions?

leezer3:
Now that's much better!

Will test and merge a little later.

Issue #29, opened 2016-02-26
Error on starting OpenBVE


v1993:
MonoDevelop sayed on line 136 in Program.cs "System.IO.DirectoryNotFoundException".
I pressed 'Show details', unchecked 'Only my code', 'Copy', and CTRL+V here:

System.IO.DirectoryNotFoundException: Directory '/home/v/compile/OpenBVE/openBVE/OpenBve/bin/Debug/Data/Languages' not found.
  at System.IO.Directory.ValidateDirectoryListing (System.String path, System.String searchPattern, System.Boolean& stop) [0x00081] in /tmp/buildd/mono-4.2.2.30/mcs/class/corlib/System.IO/Directory.cs:470
  at System.IO.Directory.GetFileSystemEntries (System.String path, System.String searchPattern, FileAttributes mask, FileAttributes attrs) [0x00023] in /tmp/buildd/mono-4.2.2.30/mcs/class/corlib/System.IO/Directory.cs:489
  at System.IO.Directory.GetFiles (System.String path, System.String searchPattern) [0x00000] in /tmp/buildd/mono-4.2.2.30/mcs/class/corlib/System.IO/Directory.cs:296
  at OpenBve.Program.Main (System.String[] args) [0x00239] in /home/v/compile/OpenBVE/openBVE/OpenBve/System/Program.cs:136

JakubVanek:
If you have used xbuild it's strange because #10 should fix that. Just copy https://github.com/leezer3/OpenBVE/tree/master/openBVE/OpenBve/Data to the Debug folder.

leezer3:
Agreed.

What distro are you using, and which and language are you set to (I'm presuming ru-RU for the langauge, but presumptions are never good)?

v1993:
I'm on Linux Mint 17.3, LANG vaule is ru_RU.UTF-8.

v1993:
Mono repo is "deb http://download.mono-project.com/repo/debian wheezy main".

mgavioli:
Just FYI, I am on Linux Mint 17.3 too, using MonoDevelop 5.10 (runtime Mono 4.2.2) which I got from the same repo you quote: no problem with `Data/Languages` files here.

My language setting is different (en-UK in openBVE and English-Ireland as system setting), but this should not matter, as in your case, the programme is failing to locate the whole language directory, not just this or that specific language file.

Does the `Data/Languages` directory exist within the directory your executable is run from?

JakubVanek:
[...]

PR #33, opened 2016-03-04
Fix Data folder copying bug

JakubVanek:
This fixes issue #29. I suspect that the latter PostBuildEvent has overridden the previous one.

### DIFF
--- a/openBVE/OpenBve/OpenBve.csproj
+++ b/openBVE/OpenBve/OpenBve.csproj
@@ -318,8 +318,7 @@ if %25textTemplatingPath%25=="\Microsoft Shared\TextTemplating\$(VisualStudioVer
   </PropertyGroup>
   <PropertyGroup>
     <PostBuildEvent Condition=" '$(OS)' != 'Unix' ">xcopy "$(ProjectDir)Data\*.*" "$(TargetDir)Data" /Y /I /E /C</PostBuildEvent>
-    <PostBuildEvent Condition=" '$(OS)' == 'Unix' ">cp -rf "$(ProjectDir)/Data" "$(TargetDir)"</PostBuildEvent>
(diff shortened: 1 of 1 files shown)


openBVE/OpenBve/OpenBve.csproj
```diff
@@ -318,8 +318,7 @@ if %25textTemplatingPath%25=="\Microsoft Shared\TextTemplating\$(VisualStudioVer
   </PropertyGroup>
   <PropertyGroup>
     <PostBuildEvent Condition=" '$(OS)' != 'Unix' ">xcopy "$(ProjectDir)Data\*.*" "$(TargetDir)Data" /Y /I /E /C</PostBuildEvent>
-    <PostBuildEvent Condition=" '$(OS)' == 'Unix' ">cp -rf "$(ProjectDir)/Data" "$(TargetDir)"</PostBuildEvent>
-    <PostBuildEvent Condition=" '$(OS)' == 'Unix' ">cp -f "$(ProjectDir)../../Dependencies/OpenTK.dll.config" "$(TargetDir)"</PostBuildEvent>
+    <PostBuildEvent Condition=" '$(OS)' == 'Unix' ">cp -rf "$(ProjectDir)Data" "$(ProjectDir)../../Dependencies/OpenTK.dll.config" "$(TargetDir)"</PostBuildEvent>
```
mgavioli: This line just copies one file; isn't the `r` option redundant? Harmless, of course, but someone might be puzzled by it or imagine that `OpenTK.dll.config` is or should be a directory, rather than a single file.

JakubVanek: 1) AFAIK cp accepts multiple sources. (http://www.cyberciti.biz/faq/copy-command/)
2) It copies both OpenTK.dll.config _file_ and the Data _directory_.

mgavioli: Aha! Correct. I didn't focused that both sources go into the same level of the same destination directory. Sorry for the noise!

Issue #38, opened 2016-03-08
Updated Build for Ubuntu & Further Development


JonnyTech:
Continued from [Updated Build for Ubuntu & Further Development ](https://github.com/sladen/openbve/issues/1) from @sladen's repo:

> Daily auto-generated builds available from here as of now-
> http://vps.bvecornwall.co.uk/OpenBVE/Builds/

On my son's computer (running Linux Mint 17.3 with Nvidia graphics - I think it's the nvidia-304 branch), OpenBVE-2016-03-07.tar.gz does not run correctly in full screen mode. I get a small windows top left of the screen with graphics, the other 70% of the screen is black. Trying to exit full-screen generates an error message which I shall post when I can get to his to his system. Windowed mode is working well.

Are there any logs that may be useful to you?

leezer3:
Don't think I've changed anything that should break fullscreen mode recently....
It's working on Windows at the moment.
You might try installing (Or conversely uninstalling) SDL2, but that's a sticking plaster, not the solution to the issue.

Under the UserData\Settings folder, it ought to have produced a file named something like **OpenBVE Crash- 2016.3.08[12:59].log**
Hopefully that will give a better idea of what's going on.

If not, the error message text may be of help.

JonnyTech:
It's running Linux Mint 17.3 64bit with Cinnamon. Previous OpenBVE's were working fine. I won't be at his computer for many hours yet but I'll do further tests then. (He's the train fan, I'm his Linux tech support!).

It works well for me, I'm running Debian Jessie Cinnamon with an Intel graphics card, I am guessing that it is an Nvidia issue but I'll report back later.

I am more than happy to do any further Linux/Ubuntu tests as required to keep the project going.

leezer3:
To note, there are a few potentially relevant changes in the most recent build for today.
Whether they help or not, they should make the build a little more stable.

JonnyTech:
OK, tried removing SDL, no change, reinstalled it again, still the same.

The message is: "An error occured whilst attempting to switch to fullscreen mode. Please check your fullscreen resolution settings."

![screenshot from 2016-03-08 21 07 01](https://cloud.githubusercontent.com/assets/3584732/13617092/cd2ca9f2-e574-11e5-97bd-3e214dd76df8.png)

[...]

PR #39, opened 2016-03-08
Fix crash for unavailable fullscreen resolutions

JakubVanek:
Window is manipulated before it's nullity is checked. This may do something with #38. I've simplified some code sections too and I replaced spaces with tabulators in this function.

### DIFF
--- a/openBVE/OpenBve/OldCode/MainLoop.cs
+++ b/openBVE/OpenBve/OldCode/MainLoop.cs
@@ -39,63 +39,54 @@ namespace OpenBve {
 
 		internal static void StartLoopEx(formMain.MainDialogResult result)
 		{
-            Sounds.Initialize();
-		    currentResult = result;   
-		    if (OpenTKWindow == false)
-		    {
-                GraphicsMode currentGraphicsMode = new GraphicsMode(new ColorFormat(8, 8, 8, 8), 24, 8, Interface.CurrentOptions.AntiAliasingLevel);
+			Sounds.Initialize();
+			currentResult = result;
+			if (OpenTKWindow == false)
+			{
+				GraphicsMode currentGraphicsMode = new GraphicsMode(new ColorFormat(8, 8, 8, 8), 24, 8, Interface.CurrentOptions.AntiAliasingLevel);
 
-                 /*
-                  * TODO: This should be able to be moved back into the screen initialisation file
-                  */
-                
-		        if (Interface.CurrentOptions.FullscreenMode)
-		        {
-		            IList<DisplayResolution> resolutions = OpenTK.DisplayDevice.Default.AvailableResolutions;
+				 /*
+				  * TODO: This should be able to be moved back into the screen initialisation file
+				  */
 
-		            for (int i = 0; i < resolutions.Count; i++)
-		            {
-		                //Test each resolution
-		                if (resolutions[i].Width == Interface.CurrentOptions.FullscreenWidth &&
-		                    resolutions[i].Height == Interface.CurrentOptions.FullscreenHeight &&
-		                    resolutions[i].BitsPerPixel == Interface.CurrentOptions.FullscreenBits)
-		                {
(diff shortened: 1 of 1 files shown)


leezer3:
That's interesting- Something's a little funny in that function looking at it.....

The original intent was to test the fullscreen resolutions against the list:
https://github.com/leezer3/OpenBVE/blob/master/openBVE/OpenBve/OldCode/MainLoop.cs#L58
If this failed, Program.currentGameWindow should be null, and then you get the crash you're seeing.

On Windows though, it seems to be falling back to windowed mode. Can't explain that at the minute, and I don't much like that.

Thinking about resolutions, it's probably a good idea for the crash handler to log the current screen resolution settings as well as everything else it catches, will add that to my list.

JakubVanek:
There are more issues with crash handler. I've modified code in StartLoopEx to throw a NullReferenceException, this is the result.
```
$ mono OpenBve.exe

Unhandled Exception:
System.NullReferenceException: Object reference not set to an instance of an object.
  at OpenBve.MainLoop.StartLoopEx (MainDialogResult result) <0x41da7760 + 0x001af> in <filename unknown>:0 
  at OpenBve.Program.Main (System.String[] args) <0x41bae9c0 + 0x0130b> in <filename unknown>:0 
[ERROR] FATAL UNHANDLED EXCEPTION: System.NullReferenceException: Object reference not set to an instance of an object.
  at OpenBve.MainLoop.StartLoopEx (MainDialogResult result) <0x41da7760 + 0x001af> in <filename unknown>:0 
  at OpenBve.Program.Main (System.String[] args) <0x41bae9c0 + 0x0130b> in <filename unknown>:0 
AL lib: ReleaseALC: 1 device not closed
```

```
$ mono --version
Mono JIT compiler version 4.2.2 (Stable 4.2.2.30/996df3c Mon Feb 15 17:30:30 UTC 2016)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           __thread
	SIGSEGV:       altstack
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	LLVM:          supported, not enabled.
	GC:            sgen
```

leezer3:
Hmm, I sorta know what's happening here, but you're not getting quite the same results as me.....

If you chuck in a **throw new NullReferenceException();** into the **StartLoopEx** function this will be caught by an exception handler added by Michelle here:
https://github.com/leezer3/OpenBVE/blob/master/openBVE/OpenBve/System/Program.cs#L243

At present, the program will silently vanish at this point, generating no error log, which isn't what should be happening.

How are you triggering the exception please?

leezer3:
I'm not sure the commit above will fix your crash, but it adds logging of caught exceptions whilst loading to the crash handler.

The next phase of the crash-handling is going to be adding specific exceptions for each crash, rather than the default Windows exception.
I've done this for one case at the moment (train.dat of zero-length- https://github.com/leezer3/OpenBVE/commit/8c455694ebb460a3599fac707a2944f97401f620) simply because I ran into that this morning.....

JakubVanek:
I'll try the patch, I've thrown it this way:
```
	    internal static AppDomain RouteInfoFormDomain;

		internal static void StartLoopEx(formMain.MainDialogResult result)
		{
            Sounds.Initialize();
		    currentResult = result;   
		    throw new NullReferenceException();
		    if (OpenTKWindow == false)
		    {
                GraphicsMode currentGraphicsMode = new GraphicsMode(new ColorFormat(8, 8, 8, 8), 24, 8, Interface.CurrentOptions.AntiAliasingLevel);
                 /*
                  * TODO: This should be able to be moved back into the screen initialisation file
                  */
               
```

JakubVanek:
Yes, the patch works. In Debug build, exception is not catched, in Release build a messagebox appears. Great work!

leezer3:
The last commit fixes this fully.
I've added a second error message explaining that the graphics card driver doesn't support the selected resolution, and fixed the second check, relegating the error message here to an 'unspecified error'

PR #44, opened 2016-03-09
Fix load error deadlock

JakubVanek:
How to reproduce bug: throw exception inside LoadEverythingThreaded
What happens during the bug: Loading.Cancel becomes true, LoadingScreenLoop ends, but the window **and it's loop** are still there. Loop is generating ~85% CPU on my dualcore Intel P8600.
What is causing the deadlock: Well, technically, it might not be a deadlock but it looks it is. Draw() is returning immediately because firstFrame is the wrong value because Loading.Complete is false. The CPU load is just Update().
What this fix does: When Loading.Cancel becomes true, LoadingScreenLoops calls Close() on the window. This fixes the deadlock.
PS: do you know where to find trainscript train plugin? OdakyUfan's web is down and google didn't find anything useful at the first glance. Alstom Metropolis requires it and I don't have it, so it throws an exception.

### DIFF
--- a/openBVE/OpenBve/OldCode/NewCode/GameWindow.cs
+++ b/openBVE/OpenBve/OldCode/NewCode/GameWindow.cs
@@ -627,6 +627,8 @@ namespace OpenBve
                 GL.PopMatrix();
                 GL.MatrixMode(MatrixMode.Projection);
                 SetupSimulation();
+            } else {
+                this.Close();
             }
         }
 


leezer3:
That seems highly reasonable, merged.

There's still a deadlock *somewhere* in the loading routine, which triggers randomly, which I haven't managed to find yet.....
I think that one is related to Windows based train plugins failing to load correctly, but we're running them through a proxy .DLL, which shields them.

Almost wants some sort of sanity check for failed loading, but I can't think what off the top of my head.

OT:
The Alstom Metopolis (Here: http://joeyfoo.com/openbve/ ) would appear to include the plugin under the Plugin directory.
Is this version not running, or what?

JakubVanek:
Yes, this version is running (thanks!), I was running some third party version without the plugin. It was probably a backup from old managed content architecture and there was the plugin separate.

leezer3:
No problem.

Another thing on my list of things to twiddle with is to provide a central plugin database, where trains can requests a Plugin XXX, version 1.0 or greater etc.
I've also got half a thought to a add a subset of BVEC_ATS to the main source, and automatically try upgrading OS_ATS based trains on non-Windows architecture, as AFAIK it now actually supports **all** features of that automatically.
Don't really want to start including too much into the main tree though.....

PR #48, opened 2016-03-10
Route map/gradient images as in-game overlay

mgavioli:
The route map and gradient profile (F8), currently displayed in an external window, are moved into the simulation window as overlay.

- Partially transparent (background about 20%, drawings about 50%).
- Fixed size (500 x 500 as before). The image is not scaled, to avoid visual artifacts and worsened legibility.
- Colour scheme changed to match the dark background.
- F8 now cycles through None | Map | Gradient | None again.
- A moving red dot (map) | white vert. line (gradient) marks the current position of the train.
- The 'old' image generation (for 'system' windows, with original colour scheme) is left in place and a new parameter allows to select the kind of image to generate.
- All colours and sizes and factored-out and put at the beginning as constants.

Sample screen shots:
ingamemap_00
ingamegradient_00


mgavioli:
1) Not sure this is the better way to implement it: should it build upon the `Interface.CurrentHudElements` support used for most of the current HUD elements in `Overlays.cs`?

2) It would be nice to trim the map if it does not use the whole width / height (like, in the sample, the empty band on the right side).

3) it would also be nice to size the bitmap to some ratio of the actual screen sizes (for instance half of the shortest screen dimension, but not over some maximum) but, in order to avoid scaling (with its artifacts), this requires rebuilding the map if the screen size changes (for instance passing to /from full screen mode) with a possible visible simulation stutter.

4) The files for the old separate map window can be removed, but I refrained from doing it, not being entirely sure of which they are.

5) Similarly, code sections in other files dealing the that window (`OldCode/MainLoop.cs` and `SystemInput/ProcessControls.cs`) have been commented out, but not removed (yet?) to allow an easier inspection of the changes.

leezer3:
Looking really nice!

I've played with semi-transparent windows in the past, but they only work to a certain extent, and aren't nearly as predictable on Mono.
I hadn't actually thought of doing it as an overlay image on the OpenGL surface, so that's a good call!

Files for the information window should only be formRouteInformation.cs and the associated designer file IIRC.

Your track position method seems reasonable enough.
A couple of extension/ tweak thoughts-
We can also attach a camera-follower to the track (Key F3)
Not convinced by this, but adding an ?option? to use this follower might be nice.

Similarly, you've posted the idea of a Novice HUD-
The position of other trains might be of interest in that.

Trouble is overcomplication though.....

Final, rather off-topic note:
I recognise that train as being one fitted with Stefano Zilocchi's OS_SZ_ATS for SCMT safety features- The ALN-501?
I've actually ported a lot of OS_SZ_ATS into into BVEC_ATS for cross-platform usage, although it's by no means finished. (IIRC, the entire startup sequence and speed selection works, but I hadn't got as far as trying to test the SCMT properly)
 I'll go back to that at some stage, but if you're interested, try building a copy of the latest master branch from source.
(The most recent development has been done on the BlinkStick branch, but nothing on OS_SZ_ATS should be in there)

leezer3:
As we're thinking about route maps, there is an issue with overlapping labels....
Take this example image:
overlap

This issue can be seen on most urban (LU tube for example) and longer routes.
I can't see any good way around this, other than implementing a mouseover tooltip on station names, which seems rather a lot of work for very little gain....

The branch I've created above makes a small tweak:
If overlapping labels are detected, labels are not rendered for stations where the train does not stop.
This helps some cases, but there are plenty more where it's not clever.

An alternative approach would be to render a completely linear route map, something like that on the Birmingham X-City timetables:
timetable

Should be achievable, but I think that it'd have a different problem with long routes- Requiring scrolling of some description....

Tempted to try building a generator for this and seeing how bad it is.

Any thoughts?

mgavioli:
**About the PR itself**:
- There is some misalignment between the current rain position in the map and in the gradient profile. I'll check it again, but this might be the best I can do by myself.
- F3: do you propose to add to the map another marker with the current position of the track camera?
- Other trains: should not be over complex, as there is a handy array of trains, isn't it?

For the rest, do you see things obviously needing some correction?

**About other layouts**:
- Not sure hiding the non-stop stations really helps in the majority of cases.
- I think the linear layout does not convey any info not already in the time table while lacking the timings. It seems to me to defeat a fair part of the map purpose (and would make complex and not so expressive interpolating the current train position).

Some thoughts:
- alternating the side when overlap is detected (if the side the label is on is linked to the door opening side, this can be conveyed by, for instance, the shape of the station marker); would not solve any possible conflicts, but should help; no idea what to do when the line is mainly in the horizontal direction, though;
- creating a larger map and then automatically scrolling the relevant part into the visible map area (tricky, but possible doable); make sense for the in-game map, where the train position can act as pivot, does not necessarily make sense for the main menu (where however, adding and managing scroll bars is easier);
- simply ignoring the issue, for the moment being! ;-)

**About the.DDL**: yes, that is an Italian FS Aln-501 with OS_SZ_ATS1.dll. I always build from current master (+ occasional local dev branches), but I have not understood what BVEC_ATS is, not which .dll I should drop where, nor I could find track of the BlickStick branch... Something is definitely escaping me...

leezer3:
Completely agree that hiding the non-stop stations doesn't really help that much.
It was the best I managed after considerably fiddling around though- Tried shifting things up, left and right etc. and it either looked odd, or something just ended up colliding with something else.

The only advantage of the linear map over the exisiting timetable would be providing the position of the train, and a relative sense of distance.
Not ideal I agree though.

F3-
Don't know!
Mostly just a half-formed idea; Both MSTS and Rail Simulator allow zoom on the map, and have a completely separate camera icon. Obviously this is only of major importance if there are multiple trains though.


BVEC_ATS:
https://github.com/leezer3/bvec_ats
The most recent 'public' built version is this:
http://bvecornwall.co.uk/downloads/beta/BVEC_ATS.zip
(Last September)

Started life as a C# port of OS_ATS.dll, and has rather spiraled out of control since then!
It (should) fully support any train running an unmodified copy of the original OS_ATS.dll, and has had a lot of considerably more complex functionality bolted on later.

It was my first serious dabbling inside OpenBVE, and a lot of the reason why I'm doing this now- I kept running into stuff I couldn't do with the current stuff.

mgavioli:
- Added to route map markers for the other trains (other trains are red and user train is green)
- Fixed the position of the current position vertical bar in the gradient profile (I didn't notice that the profile may discard some 'unused' part of the track at the beginning and at the end)
- Added comments to gradient profile bitmap creation

I am thinking about improving marker look (icons?) and trimming unused areas of the map rectangle. But, if it seems worth and no bug pops up, the PR looks already functional to me, as nice-to-have details can always be added later.

About label collision, I need to think a little more...

leezer3:
Fair enough, seems reasonable to me.

A further proposal:
Some routes use 'dummy' stations for stuff such as approach controlled signalling (The only way to hold a signal on red until the train approaches is to use a station at which the train must stop)
These always show up in the timetable, and on the map, and in messages.
These are the current options for ArrivalTime/ DepartureTime in a Track.Sta command:

> Available options for ArrivalTime:
time: The train is expected to arrive at this particular time.
omitted: The train may arrive at any time.
P or L: All trains are expected to pass this station.
B: The player's train is expected to pass this station, while all other trains are expected to stop.
S: The player's train is expected to stop at this station, while all other trains are expected to pass.
S:time: The player's train is expected to arrive at this particular time, while all other trains are expected to pass.
▸ Available options for DepartureTime:
time: The train is expected to depart at this particular time.
omitted: The train may depart at any time.
T or =: This is the terminal station. If ForcedRedSignal is set to 1, the departure signal will be held at red indefinately.
T:time: This is the terminal station. If ForcedRedSignal is set to 1, the departure signal will still switch to green before the specified time as if this was a regular station.
C: This is a station at which to "change ends". See the description below.
C:time: This is a station at which to "change ends". Changing ends will take place at the specified time unless StopDuration interferes. See the description below. 

Add the parameter **D** to departure time as follows:

> D: This station is a 'dummy' station, and will not be shown on the map, or in in-game messages.

Otherwise, do you want me to merge?

mgavioli:
1) As far as I am concerned, the PR can be merged: I removed the form files no longer used, while an attempt to improve the marker look resulted in artifacts and issues; so this is probably going to take longer than expected and it is probably not worth waiting.

2) About the file format change you suggest: I am far from having your familiarity with openBve internals (and with openBVE itself, for that matter!). However, from 'outside', an _ad hoc_ solution to counteract an undocumented trick used to by-pass a shortcoming of the signal system does not strikes me as a very solid path... It would not improve old routes, until someone modifies them and would create a version cross-compatibility just for a cosmetic improvement.

My feeling is that addressing (one or more of) the signal system limitations as such would be a better medium term path:
- It would still improve only new (or updated) routes, but with a real, structural gain.
- If this introduces cross-version compatibility issues, again it is for a rather significant gain.
- Old (or unmodified) routes will suffer from the extra station labels in time tables and maps, as they would anyway with any solution: tough luck!

So, by myself, I would incline toward leaving things as they are, until some structural solution(s) can be implemented. Of course, about that solution, it might be possible to start speculating / thinking / dreaming in the meantime, without necessarily delegating it to a distant future...

3) Same thing for the overlapping labels (which has something in common with the above, but it is not exactly the same). This is happening since quite long (possibly since ever); until a 'real' solution can be found (a better algorithm, a different layout of the map, a different kind of display, whatever), I would not spend time and resources on _ad hoc_ patches. Again, this does not mean that we should stop thinking about it!

leezer3:
Right, no problems, will merge this.

Signal limitations:
I fully agree with your assement of it it being an *ad-hoc* solution, although the dummy station concept is used for a whole bunch of other stuff too; Signalling was the most obvious and easily explained example.
The flip side is of course that constraining ourselves to keeping 100% backwards compatibility seems nuts.

As you've probably noted, I've added some of what seemed to me to be the most obvious oversights in animated object files and the plugin interface ( https://github.com/leezer3/OpenBVE/wiki/Compatibility-Notes ), and this feels to me to be another one of those.
If a routefile using this .Sta format is loaded on an older version of OpenBVE, the departure time is just set to the minimum stop time after arrival and an error is generated, so I think it's probably OK.

The overlapping labels was a off-the top of my head tweak requiring a few lines of code, and definitely far less important :)

I'll think some more about both cases more before making any changes.

PR #52, opened 2016-03-13
Commit developer documentation to the repository.

leezer3:
Creating this as a pull request for discussion, rather than just dumping a lump of this size into the repository.

### The Current Situation

At present, the source tree includes no documentation of any description, which IMHO is less than ideal.
Whilst this is currently available on the internet, I think it should probably be included with the main source repository as well.

### This Commit
I've done a little cleaning, but essentially this is a verbatim copy of the original developer documentation.
Unfortunately, it's not the easiest reading (Michelle wasn't a fan of simplification or neat explanations), but it should be technically correct.

### Future
As noted, this could do with considerable re-writing to a more friendly format.
Assuming that others think this concept to be a generally good idea, I think it probably ought to be expanded to include commented examples of things such as simple animations.


mgavioli:
Whatever the contents is (I just glimpsed a little of it), given the author, I am confident it is a valuable addition to the repository, which makes sense to keep together with the sources. Additions, explanations, perhaps a little reformatting, can come after, according to available resources and knowledge.

leezer3:
Yeah, but with other people in the repository now, I'm a little wary of dropping 100+ new files without a little warning :+1: 

It's simply a saved copy of the original developer documentation, but with tidying.
I figured that a **Documentation** folder in the root of the main source tree was probably the best place for it.

I've also got copies of Odakyufan's demo route (Japanese safety systems- ATS, ATC. etc.) and the OS_ATS demo stuff I could stick in there.
Thoughts on those?

leezer3:
Sorry, another thing:
Can someone on Linux please test the links in the right-hand sidebar?

Just clicking on each and seeing that the page loads should be fine, I've just thought that I saved these pages on a Windows machine, and hence the filesystem isn't case-sensitive.....

MarcRiera:
Just tested this on Linux: the menu links work.

However, there is another problem, but only in "Developer Documentation.html": backslashes. Linux is unable to find menu.html or title.html, so there is a "Page not found" error. I think Windows doesn't really care about backslashes or slashes in paths.

mgavioli:
Yes, I didn't check **all** the files and links again but, as far as I can see, only the references:

    <frame src="HTML\menu.html" name="menu" frameborder="0" />
    <frame src="HTML\title.html" name="main" frameborder="0" />

in `Developer Documentation.html` need to be changed to:

    <frame src="HTML/menu.html" name="menu" frameborder="0" />
    <frame src="HTML/title.html" name="main" frameborder="0" />

(I think that either HTML or HTTP **requires** forward slashes, but I am not sure).

I would suggest to remove the space from `Developer Documentation.html`, as file names (or URLs) with blanks might need escaping or quoting in some contexts.

leezer3:
Urp, fixed.

I've skimmed through them on the VM now as well, and I can't see any broken links or missing images.

Will merge in a moment.

PR #54, opened 2016-03-15
Move cd from csproj to AssemblyInfo script

JakubVanek:
- remove temporary file
 - use "/usr/bin/env bash" instead of "/bin/bash"


leezer3:
Didn't think of that whilst I was fighting with BASH.......

Must have been creating the file somewhere completely stupid!

PR #58, opened 2016-03-22
Improved exterior view camera

MarcRiera:
The exterior view camera now cycles through all the cars in the train.
Each time F2 is pressed the camera jumps to a new car, from the front to
the back of the train.

### DIFF
--- a/openBVE/OpenBve/OldCode/NewCode/GameWindow.cs
+++ b/openBVE/OpenBve/OldCode/NewCode/GameWindow.cs
@@ -81,9 +81,15 @@ namespace OpenBve
             //Not doing this means that the camera doesn't move
             // update in one piece
             ObjectManager.UpdateAnimatedWorldObjects(TimeElapsed, false);
-            if (World.CameraMode == World.CameraViewMode.Interior | World.CameraMode == World.CameraViewMode.InteriorLookAhead | World.CameraMode == World.CameraViewMode.Exterior)
+            if (World.CameraMode == World.CameraViewMode.Interior | World.CameraMode == World.CameraViewMode.InteriorLookAhead)
             {
-                TrainManager.UpdateCamera(TrainManager.PlayerTrain);
+                // Cab camera must be placed in the driver's car
+                TrainManager.UpdateCamera(TrainManager.PlayerTrain, TrainManager.PlayerTrain.DriverCar);
+            }
+            else if (World.CameraMode == World.CameraViewMode.Exterior)
+            {
+                // Enables cycling through cars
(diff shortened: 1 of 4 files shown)


leezer3:
Whilst I see the basic logic in this, a couple of thoughts:
The first is directly related to this-
Would it be better to use a 'new' camera mode, or possibly just key assignment for this one?

Second is marginally related-
Another thing on my 'list' to look at, is adding views to extra cars in the train, so that passenger views don't need a separate train configuration.
This should be reasonably straightforward (Famous last words again.....), as most of the necessary framework is already there, and I think it'd just require setting of a few appropriate variables.

MSTS and Rail Simulator both have a dedicated passenger view key, and also have popups windows which allow you to move the camera anchor point between cars in the train (The coupler window in both IIRC)

Undecided on both :)
Anyone else with any interesting thoughts?

MarcRiera:
Having played MSTS, I think that passenger view cameras can be great, but that we should not simply stick to defined presets. In MSTS, the passenger camera was preset and it was impossible to move around, which is what most people would like. Some people prefer sitting on the seat next to the window, while others prefer the corridor (just examples).

Combining both ideas into one doesn't seem crazy. We could add a new camera mode that uses these preset positions (similar to route POIs) and at the same time allow the user to use the exterior (F2) camera mode to move freely between cars and choose their desired spot. The F2 camera would also allow a simple passenger view option for old trains without the new settings.

As for the way the car is selected, I just added it this way because it was the easiest one, but it wouldn't be difficult to improve it and use the route POI keys to go forward or backwards through cars. There would be no need for additional panels.

leezer3:
Good idea!
If we change the POI keys to cycle through the available view variants, we end up with a logic list a little like this:
* Internal view- Cab ==> Passenger ==> Cab
* External view- Loco ==> Moving down train, cycling at the end
* POI- Cycle through POIs.

Most of this needs to wait until we have the replacement train configuration files into place, so I think that changing the car switch to be triggered by POI will probably do for the minute.

Will try and get to that soon......

leezer3:
To answer your MSTS point:
I think all it wants is allowing a view matching the current behaviour of the cab to be attached to additional cars.

Much of the work is therefore in place already, as we have the parsers already written for us.
3D cabs allow free motion of the camera within the cab, so I think that probably allows for your moving seats analogy. I suppose you could create a set of preset 'seats' within the coach, but that seems an awful lot of work for not much gain.

leezer3:
Only just got round to sorting this, as I've been chasing some odd bugs.

That caused a merge conflict, and it was easier just to implement this and my proposed POI key change in a single commit, rather than fixing, merging and adding another commit to change the key assignment (Sorry....)

PR #59, opened 2016-03-31
TGA image loader

leezer3:
Done a little mangling on a TGA image loader:
http://www.codeproject.com/Articles/31702/NET-Targa-Image-Reader

This now works and loads images into openBVE quite nicely.

Any thoughts?
Biggest concern is that this licence is GPL/ FSF incompatible ( http://www.gnu.org/licenses/license-list.en.html#cpol )
Whilst we're public domain, this would probably make the plugin non-available for use in any potential Debian repository builds.

Problem looking for a solution?


leezer3:
Notes:
Email has been sent to the loader author.

leezer3:
Massive cleanup of the original code.
Removed massive amounts of junk property setters and assorted functions that served little purpose.....

leezer3:
Some more cleanup of the code, not happy with it yet though.

leezer3:
Cleaned this up and merged- TGA is a not uncommon format for X files, and the two new parsers should give us much better support.

PR #60, opened 2016-04-02
Build AtsPluginProxy.dll for local builds on Windows

buttercookie42:
I came across this because I wanted to see why the plugin DLL of the F92 underground train wouldn't seem to work for me in OpenBVE [1] - you mentioned that [you couldn't get a recompiled version of the plugin proxy to work](https://github.com/leezer3/OpenBVE/commit/7619d58e8ff19dfdfae2f937e91f177be509e99b). I don't know whether it's VS2015 that made the difference, or something else (did you specify AtsPluginProxy.def as a module-definition file in the linker settings?), but for me, the locally built DLL is working fine.
I don't know what's the best solution to deal with the automation, though - just manually copy a new version into the dependencies folder if necessary for the time being? 

[1] Strangely enough, it looks like the call to LoadLibrary fails. BVE4 on the other hand has some other problems on my computer (only starts in windowed mode and the transparency in the panel image isn't working, either, so I'm not seeing anything of the outside world), however the plugin itself loads fine. The kcrwr_seltrac.dll plugin which was included with DLR_B92_SELTRAC fails in a similar manner, UKMUt.dll and the Ginza line plugin (from trta01_ver2) on the other hand work fine.


leezer3:
From memory, I don't think I added AtsPluginProxy.def to the linker (Not too much experience with C++), but I wouldn't be certain- I basically just created a new C++ project and added the files/ refs and hit the build button.
This produced a file that openBVE loaded/ called the imports from just fine, but that none of my Windows based trains recognised (Mostly UKDT / UKSPT based, not sure if I tried the F92)

VS2015 definitely has a new compiler version though, which may have helped things, although Michelle's readme states that she built using Visual C++ 6.0

The F92 plugin appears to load OK on my machine, although I haven't done much in the way of digging as to what precisely is happening.

leezer3:
Interesting thought-
What operating system and user privileges are you running under?
I do wonder whether we're hitting an administrator vs. restricted privileges issue with some of these (If the plugin is expecting to have write access to train folder)

Would also be interested to know whether they load under an unmodified version of openBVE or not (I'm assuming not, but I'd like confirmation of that)

buttercookie42:
I'll double-check tomorrow, but I think there wasn't any difference between the prebuilt and the locally built version (except that debugging the latter is easier :-) ).
I'll open a separate issue for the plugin problem, but as I said, it's the call to LoadLibraryA/W () that's failing, i.e. Windows can't even load the DLL. The weird thing is that at a first glance, BVE4 doesn't seem to do anything radically different.

leezer3:
Sorry, by unmodified I meant an Odaykufan build (e.g 1.4.3.0)
It should make absolutely no difference, but these builds have had some quite significant changes made in terms of back-end libraries, which means I don't want to rule anything out without testing it first.

Trouble is that bugs that only reproduce on a specific subset of machines are very hard to find.

If you want a copy of the OS_ATS source code in C++ I can stuff one up here, and you could always try building that yourself, and see if that runs- Should give you another point of reference.

leezer3:
OK, so a copy built from your project seems to work OK at this end.
Not sure why my initial build copy didn't work, at the time I didn't really get as far as digging into it.

Trains tested:
F92 - Plugin loads OK, but key definitions are broken, will need to fix this somehow....
Class 47 (UKDT)- Loads OK, plugin seems to be working.
OS_ATS test trains- Plugin loads OK, and all works as expected.

I've seen a couple of complaints about trains not loading properly 100% of the time (The LU tube train which comes with the Central Line seems particularly bad for it), but these issues have always been with Odakyufan builds, not these.

Final note- I'm not thrilled by the amount of extra cruft that's appeared in the .gitignore file- There's stuff from completely irrelevant toolsets and stuff which to me shouldn't really be there.
Will think about that aspect of things some more before merging.

buttercookie42:
Ah, I understand. I did indeed originally notice this with Odaykufan's last build.

I just copied the standard VS .gitignore from Github, but I can trim it back down tomorrow.

buttercookie42:
I've scaled back the .gitignore to only include what is necessary, i.e. the Debug/Release directories used in building AtsPluginProxy, the .sdf IntelliSense database and the .opendb lock file that seems to exist while Visual Studio is open.

One thing that has come to my mind: Are Mono/SharpDevelop users on Windows relevant? AtsPluginProxy won't compile for them, but the post build event differentiates based only on the OS, so on Windows it will always try to copy the file from the build directory.

leezer3:
Better I think.

Don't know about SharpDevelop users on Windows, but I'd actually be inclined to leave the AtsPluginProxy as an entirely separate project, rather than building it into the main solution.

If anyone really wants to build it/ debug with, they can do but otherwise just copy across the prebuilt binary.

Can't get OS_ATS to build at the moment either.....
Works OK until I specify the defs file in VS, but then it throws a wobbly, complaining it can't find any of the exports listed in there.
Last time I compiled this, it was in Bloodshed Dev C++.....

buttercookie42:
Okay, I'll change it into  a separate project.

leezer3:
Just as a FYI:
I'm aware of ~6 C++ DLL files (Not including variants such as OS_SZ_ATS and the OS_ATS variant that the F92 uses) and 3 C# DLL files, so you're a little on your own playing in there.

buttercookie42:
No trouble. It looks like those problematic plugins might have some bug in their DllMain routine, which only manifests itself under some circumstances (on my computer when loaded by OpenBVE, however not when loaded by BVE4 and also not on other people's computers, such as yours) and causes an unhandled exception during initialisation, which means LoadLibrary() fails [1]. Unfortunately my assembler debugging skills are near zero, so I've got no real idea what's actually going wrong. Maybe I can find something by comparing what the plugin does in DllMain when it's loaded by BVE4.

Although if maybe the basic OS_ATS plugin is exhibiting the same bug, it might be a bit easier to find out what's going wrong if I can build it from source. Hmm...

I might also try whether completely eliminating the plugin proxy might get those plugins working on my machine - I think .NET Framework 1.1 only supported the plain DLLImport attribute, which of course only takes a fixed file name, so mackoy had to use the intermediate native DLL (AtsAgent.dll), and michelle possibly simply copied his approach. From what I've seen though, since .NET 2.0 it's possible to call LoadLibrary directly from C# code as well. I'll let you know if I had any success, although it might be a while...

[1] GetLastError() returns ERROR_NOACCESS (0x3E6) - Invalid access to memory location.

leezer3:
No problem.

Untouched copy of OS_ATS:
http://web.archive.org/web/20060903055828/http://koti.mbnet.fi/~lopomo/trainsoft/OS_ATS1_3.zip

(Mine seems to be covered in code notes and assorted junk)

Incidentally, Paul Sladen (He was the Debian maintainer, although he appears to have gone uncontactable at the minute) seemed to think that it'd be relatively trivial to call Windows based DLLs on other platforms. Don't have the details to hand, will dig them out if you're interested.

leezer3:
http://stackoverflow.com/posts/8836228

I think that's essentially what you're getting at?

Would need to do some fiddling.....

buttercookie42:
The link's not quite working, but I found the page and yes, that's exactly what I meant. It's of course entirely possible that this won't help my problem at all, but I'll try...

Regarding getting native plugins to work on other platforms, my initial guess would be to somehow involve Wine (you have to do the work of the Windows loader and then that DLL might make who knows what calls to other DLLs and all the Windows APIs), but in any case I'll pass on that - it's a bit too much outside my area of expertise.

leezer3:
I wouldn't expect it to, although I'd have to do some serious digging with a debugger attached on a problem machine before I could give you any reasonable suggestions.....

First thing I'd try is loading one of the problem DLLs into memory, but without using the Load() call. Then try querying the plugin version and see whether that crashes- I suspect it probably will, as it's doing nothing interesting at all, and this suggests a fundamental issue somewhere:
```
// Function called when this plug-in is loaded

ATS_API void WINAPI Load(void)
{
	for(int i = 0; i < 256; i++)
	{
		memPanelReserve[i] = 0;
		memSoundReserve[i] = ATS_SOUND_STOP;
	}
	gPanel = &memPanelReserve[0];
	gSound = &memSoundReserve[0];
}
```

You might also try completely removing/ disabling antivirus software- OS_ATS variants have a nasty habit of causing false positives.

Haven't got the DLR Seltrac stuff to hand, so I'd have to go and find a copy, but I'd expect it to work on this machine, as everything else does.

Will try running the F92 in the debugger and see whether it does anything 'interesting' , but this may take a few days

leezer3:
Done a little further debugging on this.
I can't reproduce your issues at the minute, but I have now discovered that a copy of AtsPluginProxy built with VS2013 doesn't work on Windows XP.

No idea why, I've tried the C++ runtimes, but no dice, it's just giving me this:
`System.DllNotFoundException: Unable to load DLL 'AtsPluginProxy.dll': The specified module could not be found. (Exception from HRESULT: 0x8007007E)`
Not planning to try and install Visual Studio on this XP VM, but something's funky- All dumpbin shows it depending on is MSCVR120.dll

Michelle's version was built with MingW, and works OK.

leezer3:
Another thought-
The copy of OS_ATS1.dll which the F92 uses imports two things; kernel32.dll and mscvrt.dll
UKSpt.dll only imports kernel32.dll

You might want to try the Visual C++ runtimes.
In *theory* something ought to have installed them already, but no harm in reinstalling.

buttercookie42:
I'll give the rebuilt plugin proxy a test run on my XP VM too, then, although one more argument not to rebuild it by default for the time being. I'll update the pull request when I get around to it.

Re your comment about [how to debug this](https://github.com/leezer3/OpenBVE/pull/60#issuecomment-205075600), that was approximately my intended course of action, but just to clarify, what is failing for me is the LoadLibrary call to Windows. When OpenBVE asks the plugin proxy to load a DLL, the first thing that happens is that the plugin proxy tries loading the DLL via LoadLibraryA/LoadLibraryW and if that succeeds, getting function pointers for all the exported functions. Only then are any BVE specific functions of the plugin (such as Load()) called.

Re your last comment, it might be something, although the weird thing is that even on the same computer BVE4 is loading that plugin fine. The source code for the OS_ATS1 doesn't even include any definition for DllMain, and looking at the F92 plugin what seems to be the user-defined DllMain is the bog-standard very simple version which unconditionally returns true. For normal C++ DLLs, before that user-defined DllMain is called, though, the complier will automagically insert it's own stuff which should initialise all static/global variables. I suppose this step is what is failing for unknown reasons. I'll try if the version of OS_ATS1 that is included in that source code archive is failing, too and if it is, I'll try building it from source - might have to dig up a copy of C++-Dev to produce a hopefully identical binary, though, which I could then debug a little bit more easily.

buttercookie42:
Oh, and regarding that:
> Can't get OS_ATS to build at the moment either.....
> Works OK until I specify the defs file in VS, but then it throws a wobbly, complaining it can't find any of the exports listed in there.
The def file seems to refer to mangled function names, so if MSVC is using a different mangling scheme, it can't find the functions.

buttercookie42:
It seems the OpenBVE 1995 stock is using the same OS_ATS1.dll as the one in the source code archive (found a copy of that on my hard disk as well), and that one has no problems loading in OpenBVE on my computer despite also importing from msvcrt.dll.
The rebuilt AtsPluginProxy.dll works under XP SP3 after installing the [VC 2015 Update 1 C++ Runtime](https://www.microsoft.com/en-gb/download/details.aspx?id=49984).

buttercookie42:
I think I know what's going on: With a small test application, the F92 plugin loads without problems through AtsPluginProxy.dll and I noticed that sometimes even OpenBVE managed to load the plugin. I had already noticed - and then forgotten again - that compared to e.g. UKMUt.dll both the F92 and the Seltrac plugin have a relatively high base address, if I remember correctly something like 0x6B8.....
It turns out that apparently the F92 DLL doesn't like being rebased. I don't know how far ASLR might be involved, but at least today, in those cases where the plugin didn't want to load, the address space around 0x6B8.... was being occupied by atioglxx.dll, i.e. some part of the graphics driver.
I might try whether the DLL survives permanently rebasing it by only a few 100K if I can find a reliable hole in the address space nearby, so I could move it out of the way, although I fear ALSR might make that a bit difficult. As for why the plugin loading sometimes succeeds, maybe there is some sort of race-condition going on which DLL loads first, plus debugging might also introduce further timing variations.

I haven't got around to moving the plugin proxy into its project, hopefully I manage it tomorrow - the rebuilt version is slightly smaller (only 11K :-) ), but it's of course not really worth forcing everybody to install the latest C++ runtimes just for that.

leezer3:
`68B00000 image base (68B00000 to 68B1EFFF)`

I thought ASLR was only enabled for DLL files which had the appropriate linker flag set though?
Smells of a compiler bug that probably only affected certain builds, but we're never going to get to the bottom of exactly what I don't think.
I don't do much C++ (Other than building really) so you're relatively on your own with this one.

buttercookie42:
Not the plugin DLL, but the graphics driver DLL might have ASLR enabled. I'm no C++ guru either, but I'll try to get by :-)

buttercookie42:
I've updated the pull request to only add a new solution and project for the plugin proxy, but not build it by default.

I tried rebasing the F92 DLL to 0x6810000, but apparently that's already too much, or possibly the value is really hard-coded somewhere.
atioglxx.dll's image base is supposed to be at 0x69030000, so it shouldn't conflict, but because of ASLR it often loads slightly earlier. This also means that there is no race condition - atioglxx.dll gets loaded when the loading window appears, i.e. long before any ATS plugin is loaded, it's all down to where ASLR loads that DLL. I guess as a local workaround I'll try deactivating ASLR for OpenBVE only using [EMET](https://technet.microsoft.com/en-us/security/jj653751).

leezer3:
Looks reasonable, do you want me to merge?

We could possibly try loading plugins before the gamewindow initializes at all, but I'm not sure that will help, as I would suspect that OpenTK will request a graphics handle when Toolkit.Init() is called upon program launch. (Need to do this in order to have joystick configuration working in the main menu)

buttercookie42:
Yep, if you're happy, you can go ahead.

I'll need to check again, but I think I've seen Visual Studio trying to load symbols for atioglxx.dll only after route loading has started, so initialising the plugin earlier might indeed help in avoiding having to relocate the DLL. I don't know whether it's worth the effort though - did anybody else ever report similar plugin problems? And I think I've mostly satisfied my F92 nostalgia for the time being :-)

leezer3:
It's very hard to tell!
I know of several reports of plugins failing to load after multiple driving sessions, which smells very related, and I've seen one or two other reports of plugins failing to load, but no-one has really dug into it. 
Certainly not common though- I've got two ATI systems which don't do this. (Mobility Radeon X1200 and HD6870)

FYI: I've added a bunch of the F92's functions to BVEC_ATS https://github.com/leezer3/bvec_ats/commit/b1c5022445367050ec573ad25a7b925c84005952 , but this isn't ready at the minute, and won't support all features, as I don't have the source. Am playing with it though.
Eventually, I'd quite like to add BVEC_ATS and UKTrainSys to the main source tree to automatically replace older plugins on non-Windows systems, and in cases like these, but that's a wishlist item more than anything.

buttercookie42:
Oh, a managed drop-in replacement. Sounds like an interesting idea.
When I find some time, I can look into improving the logging in case of a plugin load failure - currently it looks like any failures are simply silently swallowed.

leezer3:
Not sure off the top of my head. I think it adds an entry to the main simulation log, but doesn't specifically prompt the user. That's probably a reasonable thing, but if we can detect the specific failure cause and log it internally that's good.

Michelle removed any in-game error display (Other than the initial display of 10 warnings etc.) at some stage; It did used to have a popup window similar to Route Viewer.

You can normally tell when a plugin is missing altogether, as the default ATS overlays popup in the bottom right, but I haven't checked whether this is the case with failures to load, mostly because I don't have any machines exhibiting this behaviour.....

buttercookie42:
I wasn't referring to the in-game display, but to the log file itself (the one in the settings directory). It doesn't contain any entries for failed plugin loads (at least not for Win32-plugins, I haven't tested whether managed plugins are treated any differently).
Yes, if plugin loading fails, the built-in security system takes over, but if you don't expect it, it's easy to overlook in the first moment.

leezer3:
Added a specific on-screen error that the plugin failed to load for 5 seconds, which on reflection appears to be good manners.

I've also tweaked the Win32 load failures very slightly to log a final catch-all error at the point the default plugin is loaded, but I'll note that the logging appeared to work before I did this.
Basically, I think this is related to your AtsPluginProxy troubles, in that the plugin appears 100% valid before it raises some sort of exception somewhere.
If you can see where the error is being caught, then a specific error should be easy enough to catch.

buttercookie42:
When LoadLibrary fails, it simply returns NULL, which is detected by AtsPluginProxy which in turn reports back 0 [to OpenBVE](https://github.com/leezer3/OpenBVE/blob/4293c37649ceb4c810702a7133a529801bd22a4f/openBVE/OpenBve/OldCode/LegacyPlugin.cs#L154). There are no exceptions involved, however you can get an error code for logging by calling GetLastError(). I've [implemented that here](https://github.com/buttercookie42/OpenBVE/commit/9dc8d5c2c565381fed3b3ca2dcef9970b8f7d36f), but I don't quite know yet how to best feed that back to the plugin manager so it ends up in the logfile.

leezer3:
The way the logfile is handled is somewhat of a mess....

Adding a call to _Interface.AddMessage_ after line 157 should add the error to the log:
`Interface.AddMessage(Interface.MessageType.Error, true, "The Win32 plugin error returned was: " + errorMessage);`

The trouble is that this (Or the log-file for that matter....) doesn't actually work if Errors and Warnings are turned off.
Relocating the save to log file statement to the start of the _Interface.AddMessage_ method (Before the error/ warning conditions are checked) would probably work, but I haven't really dug whether that would cause any more issues at the minute. 

Can't see that it would, but I don't like to make changes without looking for implications first.

PR #65, opened 2016-04-20
Main form tweaked

MarcRiera:
The main form could become stuck in the background and would not close correctly when starting a route under Mono/Linux.

### DIFF
--- a/openBVE/OpenBve/UserInterface/formMain.Start.cs
+++ b/openBVE/OpenBve/UserInterface/formMain.Start.cs
@@ -435,6 +435,7 @@ namespace OpenBve {
 				if (System.IO.File.Exists(Result.RouteFile) & System.IO.Directory.Exists(Result.TrainFolder)) {
 					Result.Start = true;
 					this.Close();
+					this.Dispose();
 				}
 			} else {
 				System.Media.SystemSounds.Exclamation.Play();


leezer3:
Interesting.

Calling **Dispose()** is as I understand it somewhat of a brute-force approach.

Calling **Close()** in the manner we do *should* request the runtime to unload all resources, then when this has completed and then when the form has closed the **Dispose()** method is called- This suggests that something is still holding a resource handle at the time the form has closed, and your patch is effectively brute-forcing it.

Something is obviously holding onto a handle or failing to terminate somewhere. I'd bet this is related to the grey tabs somewhere along the lines.

Will do a little digging.

leezer3:
A thought:
Might be better to wrap it in a Mono only wrapper, as this bug only affects Linux/ OS-X, hence:
```if (CurrentlyRunningOnMono)
{
this.Dispose();
}```

Need to test on Windows too and see if anything complains.

MarcRiera:
Thank you, this bug is a mystery.

I will do some research on the code starting the simulation, because I think something is wrong there. I tried adding a MessageBox.Show() after this.Close() and this effectively closes the form, just like it happened when the route info was displayed in a separate window.

leezer3:
In that case, it looks like the new window is hijacking the message pump
for the main form, which it shouldn't be-
The message box is a modal dialog, and must be shown before the new window
initialises.

A **Thread.Sleep()** might help, will do some playing.

leezer3:
Right, I think I've got a better solution.

Adding an **Application.DoEvents()** call directly after **this.Close()** forces the runtime to process all pending messages in the window event pump, which would appear to clear the issue.

It's still somewhat of a hack, but I think a little better than simply dumping dispose, and hoping that the GC cleans up whatever was hanging the close call in the first place.

Can you test at your end please?
Have tried in my Ubuntu & XUbuntu virtual machines, but I don't have any physical Linux boxes to hand.

MarcRiera:
Works perfectly as expected.

It is indeed a hack, but at least we have avoided a Dispose(). Another option would be to call this.Hide() right before this.Close() to make it invisible, but I prefer your solution rather than keeping it in the background potentially eating up some resources.

PR #68, opened 2016-05-23
Update ca-ES.cfg

MarcRiera:
Package Management translations added.


leezer3:
Another 15 missing strings:
https://github.com/leezer3/OpenBVE/commit/4761421afab7b03d89abab321d3aaa3011af76e6

I *think* that's just about everything in there at the minute, although there may be a couple of buttons not translated.

Do you want to update the pull, or should I merge manually?

MarcRiera:
I will update it myself, no problem. Thank you!

leezer3:
Merged.

I hope packages are reasonably stable at this point, and so I've added a packages page to the site:
http://openbve-project.net/packages/

Phontanka has also tried them with reasonable success (Other than finding a Sharpcompress bug!), so this is getting a lot closer to being stable.

PR #70, opened 2016-06-23
Fix: Add translation for additional strings

MarcRiera:
Around 30 strings were not available for translation because they were hardcoded in English.


leezer3:
Merged into the RC branch after fixing the travis build failure.

Will do some more testing & checking of my own this weekend before starting to think about RC3.
There's one more fix in my pile that I know about (Disable display lists when unloading textures), plus I'm sure more that I've missed.....

Issue #71, opened 2016-07-01
A couple of relatively critical OS X issues


mdiamond14:
Cross posted from: http://bveworldwide.unlimitedboard.com/t1229p400-new-openbve-build-testers-please#14393

Heads up, the Mac download link for openBVE RC4 on http://openbve-project.net/ leads to a 404 error page.

Also, im getting this error when i try to make a package:

http://i.imgur.com/w56BoiR.png

Additionally, packages seem to be installing in the wrong place. Aren't they supposed to install to either the directory where one has previously been loading routes/trains from (for existing installations)? This directory is in a hidden folder and is different from the legacycontent directory inside of the .App.

http://i.imgur.com/mzX7dV1.png

Perhaps a solution to this issue would be to have separate "Content Packages" and "Manually added content" selection tabs on the "start new game" dialog. 

Thanks for all your help!!

leezer3:
Arrgh...

TravisCI (Which generates the Mac builds automagically for me) has had a major outage. Something still isn't working right at their end, I'll package the current version manually in a minute.

Package creation hasn't been tested on the Mac, but will do that as something is obviously wrong.
The website label shouldn't be showing through the form either, doesn't do that on Windows or Linux.

The Mac installation directory is also something on the list of stuff I need to really think about.
Michelle/ Odakyufan never provided an official Mac app, so there's no framework in place to handle things in a 'Mac' manner. Installing content inside the application seems to me to be a *seriously* bad idea, especially as I've come across people who just run apps from a mounted dmg.
I decided not to use the existing routes/ trains folder for packaged content (At least by default), as the installation/ uninstallation stack still isn't nearly as robust as I'd like it to be. 
Additionally, if you can avoid it, it's a bad idea to mix managed and unmanaged content.

PR #72, opened 2016-07-02
Package management improvements

MarcRiera:
I have been working on the package management user interface to make it look more robust.

Changes:
- Window contents are now placed correctly, even after the window is resized.
- Two additional translatable strings related to packages have been added.
- Package lists are now locked and cannot be sorted (temporary measure to avoid a glitch).
- Packages added as dependancies/recommendations are now listed (image below).
- The website label glitch *might* be fixed (#71). 

dependancies

To do:
- Add a button to remove packages from the "selected packages" list.
- Improve the package lists code to avoid uninstalling/replacing an incorrect package if list items have been sorted.
- Fix some strings that still cannot be translated.
- Fix various internal bugs that make the package system unstable when using dependancies.


leezer3:
Thanks, looks good, although I need to have a proper dig....
I'd probably change your new list as follows:
* Keep the name, minimum version and maximum version columns, but change the author/ website column for a single type of  'Depends' or 'Recommends' 

Other stuff:
* I think there's probably an issue somewhere with the way the background worker is handled- Whilst I haven't had a proper chance to dig into this, this is the most likely thing to be wrong. Threading is always a pain....
* Sorting the datagridview: I think this needs a hidden column to be used as the key for selecting items within lists. Probably GUID, will do some fiddling.

leezer3:
OK, this commit lets us sort the datagrid views:
https://github.com/leezer3/OpenBVE/commit/340d7860bab030136b4057420b135f08efe01c91
Using the hidden column idea from above, best I can think of at the minute.

I've also added a **while** loop in an attempt to let the background worker thread terminate if it hasn't already, but if the background worker thread freezes, this may cause more problems than it solves....
(It shouldn't ever be reset before it terminates, but I'm wondering if a couple of the odd exceptions seen are because the thread is reset forcefully before it's ready....)

You should be able to add a 'remove dependancy' button using the same key method I've done above-
Did you want to do this/ any other tweaks before I merge, or should I just merge and sort out?

MarcRiera:
Thanks, I will add a 'remove dependancy' button using this. I would leave this open at least until then; I have merged the two recent commits so I am able to use the new column.

I have also been able to fix the bug preventing packages with dependancies from being installed. There was a loop in the function checking dependancies that was not broken properly and would lead to an "Unhandled Windows Forms Exception". It was also related to the cosmetic change to the 'Version' key in dependancies, which would delete it from the installed package too and leave it without a version number.

I will commit later and keep working on the interface translations, so maybe it is better not to merge yet until I finish them.

leezer3:
No problem, will leave it for the minute.

Another thought- I'd be tempted to remove the packages from the main list if they're added as a dependency. Thinking along these lines, I wouldn't be surprised if adding the same dependency twice with differing version requirements caused all sorts of trouble.....
Will do some testing along those lines tomorrow probably.

MarcRiera:
I agree on both.

I remember adding the same package as a dependancy twice in previous builds, and the entry would be recorded twice in the XML file. Now that GUID is part of the package lists it should not be difficult to hide/show rows as desired.

Adding generic next/back buttons at the bottom would also look better than now. Currently it is difficult for the user to know when is package actually installed or uninstalled. If we change the big button for a pair that says "Install" in the very last step the user will feel he/she has more control over the process, not to mention a "Back" button produces the same effect and is always welcome.

MarcRiera:
Finally managed to kill that annoying bug that prevented the installation of packages with dependancies.

Other fixes:
- Packages of type 'Other' would not be installed.
- Packages were not correctly removed from the database after uninstalling.
- Dependancies can now be removed using a new button.
- Package lists can be sorted again.

To do list:
- Add some strings to the language files.
- Replace the big button with a back/next button pair.

MarcRiera:
The latest commit finally fixes everything that was left, so now you can merge. GitHub says there are conflicts to be resolved before merging, probably because you added three new comits that affect files I have been editing. Do not worry about it, though, I have manually added the changes to the code while working (so merging the pull request will not revert anything).

Changes:

- All the strings in package management can be translated.
- Tweaked the position and size of some buttons.
- Replaced the big buttons with a back/next pair (example below).

buttonpair

leezer3:
Done, thanks :)

(Don't worry too much about manually merging changes, I can always cherrypick from both)

PR #78, opened 2016-07-07
Main interface content position fixes

MarcRiera:
Changes to fix this: http://bveworldwide.unlimitedboard.com/t1297-openbve-1-5-0-rc1-bugs#14405

- Buttons and labels in the main form (specially in the options panel) are now placed correctly.
- The 'Unload unused textures' checkbox has been moved below to allow for more space.
- Added a colon to 'Timetable mode'.

Except for the past changes, this commit is identical to my previous one (two days ago). This is because my code was not merged correctly and old buggy code was re-added instead, leading to bugs such as the progress screen mess. You will notice that a lot of lines are changed, but do not feel scared, all the features are there.

I would also like to point out that there is no need for a specific code to center the progress text, as Windows Forms provides native support for that (this is why that part of the code is gone). Elements in a form can be anchored to one or multiple sides so they resize or move easily, but removing anchors also has a purpose: it makes elements become centered in the part of the screen where they are.

Regards,

Marc


leezer3:
Sorry, been away for a few days....

I can't review this absolutely properly and test on anything other than the Windows laptop until the start of next week, but will do so as soon as I can.
Just about all of the changes I can see in your new commit are whitespace conflicts, and a couple of functions which seem to have changed places; I'm not quite sure how they've managed to get into it.....
The only broken issue arising from the actual merge I can see is the non-centered label in the progress dialog: This must have got broken when merging, as I'm sure it was working before the merge?


I also found that the function to keep the label centered was absolutely required, although I'll try your changes and see if they work :)
Whilst I don't understand some of the vagaries of WinForms as well as I'd like, this is what I observed whilst trying to get this to work (Windows, although I then tested my changes to make sure I hadn't broken anything on both Mac and Linux VMs): 
* A measurement of the original string length and font size is taken and stored somewhere internally when the form's constructor is run.
* When the string length or font size property is altered dynamically in code, the measurement used by the anchor points doesn't actually seemt to be being updated. 
* What this then means, is that whilst the string position remains anchored in the center of the form, but it's centering is wrong as follows:

1. If our original string has the dimensions of 120px wide by 20 px high, the center point will be calculated as 60px , 10px
2. Our string then changes to a dimensions of 200px by 20px
3. The anchored center point then remains at 60px , 10px
4. This means that the string then overhangs to the left (IIRC) by 40px

This effect isn't actually all that noticable in the current incarnation of the progress screen, as most long names tend to overhang off either end of the screen, and hence this effect is invisible.

Issue #81, opened 2016-08-02
[Suggestion] Cycle option for rail types


MarcRiera:
After using BVE and OpenBVE for a long time and developing routes, I have always wondered why the .Cycle option is limited to .Ground objects. It is a good option to avoid repeating the same grass texture each block, but extending it to .Rail objects would basically add the possibility of ignoring the legacy block sytem while retaining compatibility with existing addons. Let me explain the concept.

The legacy BVE block system limits the placement of track-dependant objects to positions that must be a multiple of 25. OpenBVE took a major step forward by allowing the user to choose a custom block length; short blocks can be used to achieve accurate track geometry in a similar way to what BVE5 does. However, a smaller repeating distance means that rails and track-dependant objects (walls, dikes, etc.) must be shorter, so long and realistic textures are impossible. Also, if the user wants to place an object every 20 metres with a block length of 5 metres, for example, the only solution is to type hundreds of lines of .FreeObj commands or to use the .Pole command (which has its own limitations).

Enabling the option to cycle through rail objects not only would make development easier and faster, but better results could be achieved. For example, let's say an addon developer wants to repeat a long track ballast texture every 25 metres while using 5-metre blocks. With a .Cycle option, the developer could split a 25-metre rail object into five 5-metre objects, each with one part of the texture, and make OpenBVE cycle through them to get the same result. Needless to say, it would be extremely simple to create a curve with these textures without the need to create a specific curved object.

However, the possibilities could expand even more if we defined the length and the span of the cycle. This would produce interesting results in curved tracks. 'Length' refers to the actual object length, and 'span' to the repetition distance. If a cycle had a 'span' of 100, each new object would be placed 100 metres after the previous one (think of the current .Pole command). 'Length' would be used to apply rotation to the object so that its end is correctly centered on the track position where the next object of the cycle is placed (the same rotation already used to connect blocks in curves but expanded beyond the current block length). This new option would be on par with the .Repeater command used in BVE5. Of course, we all know 'rails' in BVE are used for more things than just tracks (overhead wires, trees, etc).

The current CSV route format has a specific namespace for .Cycle structures that already handles ground cycles, but it is only used for Cycle.Ground. 'Cycle.Rail' could be easily defined without breaking compatibility with existing addons. A simple cycle option without 'length' or 'span' would be pretty much straightforward, but the two additional option would not be too difficult to add either (it is just some additional calculations). A rail cycle definition could look like this:

[...]

PR #82, opened 2016-08-04
Cycle.Rail support

MarcRiera:
Added support in both OpenBVE and RouteViewer for the Cycle.Rail command in CSV routes. It has the same functionality as Cycle.Ground but applied to rail objects. If the Railtype structure index is the same as a Cycle.Rail index, the cycle will be used. If not, it will look for a standard rail object.

More information at #81


leezer3:
Thanks, looks good, but I'll have a play with it tomorrow before merging.

PR #84, opened 2016-08-08
[BVE5] Character encoding tweaks and more

MarcRiera:
- Improved character encoding detection
- Route image is now displayed in route selection menu
- Less strict file version check
- Better handling of comments

The scenario file detector code for the main menu is not included, but I will upload it once I am sure it is stable.

### DIFF
--- a/openBVE/OpenBve/OldCode/Loading.cs
+++ b/openBVE/OpenBve/OldCode/Loading.cs
@@ -136,7 +136,7 @@ namespace OpenBve {
 			bool IsBve5 = Bve5ScenarioParser.IsBve5(CurrentRouteFile);
 			if (IsBve5)
 			{
-				Encoding enc = Bve5ScenarioParser.DetermineFileEncoding(CurrentRouteFile, CurrentRouteEncoding);
+				Encoding enc = Bve5ScenarioParser.DetermineFileEncoding(CurrentRouteFile);
 				Bve5ScenarioParser.ParseRoute(CurrentRouteFile, IsRW, enc, CurrentTrainFolder, ObjectFolder, SoundFolder, false);
 			}
 			else
--- a/openBVE/OpenBve/OpenBve.csproj
+++ b/openBVE/OpenBve/OpenBve.csproj
@@ -16,7 +16,6 @@
     <TargetFrameworkProfile>
     </TargetFrameworkProfile>
     <SignAssembly>False</SignAssembly>
-    <DelaySign>False</DelaySign>
     <RunPostBuildEvent>OnBuildSuccess</RunPostBuildEvent>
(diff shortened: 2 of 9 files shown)


leezer3:
Thanks.

Some thoughts:
The file version check being fixed was semi-deliberate- I've got no documentation on the prior format, other than that the textual name of a couple of commands appears to have been altered. 
Decompiling the Map 1.0 to 2.0 convertor suggests that the single quotes enclosing keys were also added.

Your character encoding solution is neater than mine, but breaks if we encounter a colon somewhere in the string, other than separating the charset declaration from the string. (Not that this matters, as BVE5 doesn't use the colon anywhere else in the string)

Good work on the comments- Hadn't really done anything other than to fix those which were in the converted route I'm playing with.

MarcRiera:
I completely understand why the initial version check requires a very specific file version. I still do not know which changes there are between versions, but because they will surely be minor changes, I suggest a 'double check' system. The checks we have so far detect unsupported versions and prevent the route loading. This would be the first check. Deeper in the code, however, we would add another check for commands (such as keys) that change depending on the version.

The character encoding check should not be problematic if a file with an invalid header (even with extra colons) is checked. In that case, the check would fail, UTF-8 would be returned and the file would fail to load when the content parser detects that it is invalid.

MarcRiera:
Scenario detection is now added, it should work for most BVE5 routes. Some of them fail to load but it is related to the 'Include' command (which is not yet supported). Also, there was a command conflict between Signal.Put and Signal.Speedlimit, so I have fixed that too.

Feel free to merge and close the pull request.

leezer3:
A thought before I merge- Would the comment stripping be better placed in an extension method to the string class?
http://stackoverflow.com/a/4910143

Seems a little of a waste to declare it for each file we read in.

MarcRiera:
Sounds like a good idea, it will make the code simpler. The empty line check will still be required, though.

I will implement it and commit.

MarcRiera:
The string extensions are found in the Extensions.cs file (the same where the image extensions are stored). So far the only additional command is String.TrimBVE5Comment(), which removes '#' and '//' comments. The extensions are available from anywhere in the code, so it will be easy to add other functions related to any of the parsers.

I also forgot to upload two files in the previous commit; now the Travis CI build check should not fail.

leezer3:
A note:
An extension method must be called on a string, e.g.
``Lines[i] = Lines[i].TrimBve5Comments();``
That works :)

``Lines[i].TrimBve5Comments();``
This doesn't (You're effectively calling a method and immediately discarding the results)

Should have noticed that before I merged, but oh well....

MarcRiera:
Ooops, sorry, that was a silly mistake on my part, I guess I should not be coding that late at night...

PR #101, opened 2016-09-18
Implement HorizontalGauge and VerticalGauge for panel2.cfg

leezer3:
This implements a request via email, allowing a gauge with a horizontal or vertical needle.
verticalgauge

How this works:
This creates a new panel layer, which uses the texture shift function to move the needle across the gauge.

New Panel2.cfg segments:
~~__HorizontalGauge__
__VerticalGauge__~~
These both work essentially like any other panel2.cfg entry, but take one further parameter: _Width_
This represents the width in pixels of the needle, and is needed to stop the needle running right back around again.

Sample panel2.cfg:
```
[HorizontalGauge]
Subject=sec
Location=493,513
DaytimeImage=..\81xx_2DCab\Panel2\needle2.png
NighttimeImage=..\81xx_2DCab\Panel2\needle2.png
TransparentColor=#0000FF
Layer=10
Minimum=0
Maximum=1
Width=32
```

Needle image (The red is just for illustration purposes whilst testing, in reality the whole of the rest of the image would be transparent and an appropriate masking layer used) :
needle2

Anyone care to comment or test please?


leezer3:
OK, revised the commit a tad, as I decided that adding two new gauge types was a mistake.
It now uses the single type __LinearGauge__
This has one additional parameter: _Direction_
Direction should be two comma separated numbers representing the direction in which to translate the texture. 
The default for this is: `Direction = 1,0` which will shift the texture RIGHT.
`Direction = 0,1` will shift the texture DOWN etc.

leezer3:
Updated the included documentation to handle the linear gauge.
A sample for the image above now looks like this:
```
[LinearGauge]
Subject=kmph
Location=493,513
DaytimeImage=..\81xx_2DCab\Panel2\needle2.png
NighttimeImage=..\81xx_2DCab\Panel2\needle2.png
TransparentColor=#0000FF
Layer=10
Minimum=0
Maximum=100
Width=32
```

Issue #90, opened 2016-08-26
Lighting relative to time of day


leezer3:
Some stuff I've been thinking about, marginally related to #89 

Lighting at the moment is defined by the **Route.DirectionalLight** and **Route.AmbientLight** commands, and is fixed for the duration of a run without using tricks.
Whilst it can be altered to an extent by using the **Brightness** command, this isn't good for changing between radically different lighting. (A true night- dawn- day route for example)

What this requires, is to link the brightness to the time of day.
A brief synopsis of how this should work:
Brightness definitions need to be moved into a completely separate file.

This requires the following data structure:
A time, ambient light and directional light value must be defined for the following datum points:

```
Solar midnight
Dawn
Solar noon
Dusk
```

A sample XML:

``` XML
<?xml version="1.0" encoding="utf-8"?>
<openBVE xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
  <Brightness>
    <Midnight>
        <Time>12:00</Time>
        <AmbientLight>100,100,100</AmbientLight>
        <DirectionalLight>55,55,55</DirectionalLight>
        <LightDirection>40,230</LightDirection>
    </Midnight>
  </Brightness>
</openBVE>
```

The current lighting condition can then be relatively easily interpolated on a per-frame basis.

Routefile commands to override the ambient and directional light values would also be desirable, but this increases the compatibility divergance.

leezer3:
A further thought:
Lighting should be affected by the weather,
This potentially means that the brightness values could be enclosed within a further structure, using a format something like this:
```XML
<Clear>
<Brightness>
    <Midnight>
        <Time>12:00</Time>
        <AmbientLight>150,150,150</AmbientLight>
        <DirectionalLight>55,55,55</DirectionalLight>
        <LightDirection>40,230</LightDirection>
    </Midnight>
</Brightness>
<Cloudy>
<Brightness>
    <Midnight>
        <Time>12:00</Time>
        <AmbientLight>100,100,100</AmbientLight>
        <DirectionalLight>55,55,55</DirectionalLight>
        <LightDirection>40,230</LightDirection>
    </Midnight>
</Brightness>
</Cloudy>
```

leezer3:
So, the branch above implements a first basic attempt at doing this.
Sample XML which cycles between 100% dark and default brightness every 6 hours:
```XML
<?xml version="1.0" encoding="utf-8"?>
[...]

PR #103, opened 2016-09-24
Implement basic dynamic lighting model

leezer3:
This requires a new command: **Route.DynamicLight(_Light.xml_)**
_Light.xml_ points to an XML descriptor file describing the brightness at different times of day.
https://github.com/leezer3/OpenBVE/issues/90

Overrides Route.AmbientLight, Route.DirectionalLight and Route.LightDirection

Sample lighting XML file:
```
  <Brightness>
        <Time>00.00</Time>
        <AmbientLight>0,0,0</AmbientLight>
        <DirectionalLight>0,0,0</DirectionalLight>
        <LightDirection>0.223606797749979, 0.86602540378444, -0.447213595499958</LightDirection>
  </Brightness>
  <Brightness>
        <Time>06.00</Time>
        <AmbientLight>160,160,160</AmbientLight>
        <DirectionalLight>160,160,160</DirectionalLight>
        <LightDirection>0.223606797749979, 0.86602540378444, -0.447213595499958</LightDirection>
  </Brightness>
  <Brightness>
        <Time>12.00</Time>
        <AmbientLight>0,0,0</AmbientLight>
        <DirectionalLight>0,0,0</DirectionalLight>
        <LightDirection>0.223606797749979, 0.86602540378444, -0.447213595499958</LightDirection>
  </Brightness>
  <Brightness>
        <Time>18.00</Time>
        <AmbientLight>160,160,160</AmbientLight>
        <DirectionalLight>160,160,160</DirectionalLight>
        <LightDirection>0.223606797749979, 0.86602540378444, -0.447213595499958</LightDirection>
  </Brightness>
</openBVE>
```

### DIFF
--- a/openBVE/OpenBve/Graphics/Renderer/Functions.cs
+++ b/openBVE/OpenBve/Graphics/Renderer/Functions.cs
@@ -149,28 +149,6 @@ namespace OpenBve
         }
 
         /// <summary>
-        /// Initializes the lighting
-        /// </summary>
-        internal static void InitializeLighting()
-        {
-            GL.Light(LightName.Light0, LightParameter.Ambient, new float[] { inv255 * (float)OptionAmbientColor.R, inv255 * (float)OptionAmbientColor.G, inv255 * (float)OptionAmbientColor.B, 1.0f });
-            GL.Light(LightName.Light0, LightParameter.Diffuse, new float[] { inv255 * (float)OptionDiffuseColor.R, inv255 * (float)OptionDiffuseColor.G, inv255 * (float)OptionDiffuseColor.B, 1.0f });
-            GL.LightModel(LightModelParameter.LightModelAmbient, new float[] { 0.0f, 0.0f, 0.0f, 1.0f });
-            GL.CullFace(CullFaceMode.Front); CullEnabled = true; // possibly undocumented, but required for correct lighting
-            GL.Enable(EnableCap.Light0);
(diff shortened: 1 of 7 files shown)


leezer3:
One point which needs addressing:

* Cab brightness is unaffected by this (As set by the _**Track.Brightness**_ command. This means that at 'night' the daylight cab is shown. This can probably be fixed by adding a `<CabBrightness>` section to each XML node. If this value is set, show the smaller of this and the _**Track.Brightness**_ value, otherwise show the _**Track.Brightness**_ value.

leezer3:
Should have finished the cab lighting sooner, but it's now been done!
A Dynamic Lighting XML file will now take the parameter `<CabLighting>`
This should be an openBVE brightness value from 0 to 255, representing the maximum lighting level in the cab at this time.

**Video (Click):**
[Everything Is AWESOME](https://www.youtube.com/embed/gXSlsZycaQQ)

**Routefile implementation:**
This much simplifies lighting in tunnels, as you no longer need to think about the time of day when.
Simply assume that the exterior brightness is a constant 255, and set your desired brightness values in tunnels etc. accordingly.
Return the brightness to 255 when exiting the tunnel.

The renderer will then interpolate the appropriate brightness value for the current time.

The _**SMALLEST**_ value of your set brightness and the time-based brightness will then be used.
Hence, if the time of day is darker than your tunnel, the cab brightness will not change.

PR #115, opened 2016-11-22
Specify starting station via command line

leezer3:
Ran across this elsewhere:
https://github.com/shadow93/openBVE/issues/6

Seemed like a good idea, so have implemented a first take on it.

Example command-line:
```
OpenBve.exe /route="D:\Program Files\BVE\Railway\Route\Plymouth- St Austell Demo\1980s\Class 104 (Stopping)- Liskeard [c].csv" /station=Devonport
```

**Details:**
If a station name is supplied, we attempt to find it first.
The stop status of this station is ignored (Assume the player knows what he's doing!)
If this station name is not found, we then fall back to the original method.

Very slightly hacky code used to implement this, but I can't see a better idea off the top of my head.
Will think about it a little more before merging, unless anyone else has a better idea.


leezer3:
Couple more useful little switches:
/time : Sets the start time of the simulation.
/ai : If set to true or 1, the simulation starts with AI driver enabled.

Some more thoughts for ones which might be useful:
/mode : Change simulation mode between standard expert etc.
/w : Set screen width
/h : Set screen height
/fullscreen : Set fullscreen mode

leezer3:
New switches:
**/width** - Integer screen width
**/height** - Integer screen height
**/fullscreen** - 1 or true

**Other changes:**
* Improve detection of encodings for command-line launches
* Split encodings into separate class
* Split command-line arguments into separate class
* Cleaned up some old code in the Screen class

PR #119, opened 2016-12-11
Unify build system into a single makefile and single solution

Sirflankalot:
These rather drastic series of changes reorganizes the project so that everything can be built at once and the project is flatter.

**Changes:**
- All projects are now a top level folder inside source/ instead of being nested inside each other. 
- Debug builds get built into bin_debug/ and release builds get built into bin_release/.
- There is both a makefile and a solution so that building on linux is less finicky.
- Capitalization schemes were made consistent so that linux (which has a case-sensitive filesystem) plays nice with various components
- Travis-ci uses linux makefile so output is easier to dissect.
- Makefile uses higher warning level so that mono catches more bugs. 
- One or two small code changes were necessary for it to compile on linux. Errors were caused by not specifying the full name of a type and mono getting confused, or a minor problem with the wrong function being called when specifying the perspective matrix.

I have tested this and all programs build and run fine. Should pass travis-ci checks. You might have to check the builds that were happening after success to make sure it works well with the new build system. It *should* based on what I can ascertain, but I can't really test it. 

**Edit:** Fix description of errors on linux.


leezer3:
Observations so far:
* AssemblyInfo.cs for the main program is auto-generated via AssemblyInfo.tt (Win) or AssemblyInfo.cs (*Nix) Your makefile isn't taking this into account, and it should be excluded from the repo as a result.
* The makefile probably needs an option to build just the main program and dependancies.
* I'd be inclined to copy the data last in the makefile, simply because if the build fails you'll have copied a bunch of stuff unnecessarily.
* I'd probably use the -u switch on the copy command too in the makefile. I'll need to check whether the Windows data copy does this too, as it should if it isn't.

I'll try building on all three platforms later.

Sirflankalot:
Regarding your comments:

- That's true forgot about that, I'll look into implementing auto generation of that. 
- Can be done
- That's a good idea, I'll prep directories first and then copy data.
- It already does, it's just made more complicated because OS X's implimentation of cp doesn't have a -u flag. See 22a38d0 for the necessary changes in the makefile and [here](http://ss64.com/osx/cp.html) for OSX's cp manpage.

Sirflankalot:
I've taken all you advice into account with these new commits. I can't find a portable way to have AssemblyInfo.cs to be auto-generated by MSVS during the build. It is auto-generated by the makefile however. You can double click on the text template and an AssemblyInfo*.cs will be generated which compiles just fine. 

Here are the new targets for the makefile:

- _default_ (alias for openbve-debug)
- debug (alias for openbve-debug)
- release (alias for openbve-release)
- openbve (alias for openbve-debug)
- openbve-debug (builds only openbve with debug settings)
- openbve-release (builds only openbve with release settings)
- all (alias for all-debug)
- all-debug (builds everything with debug settings)
- all-release (builds everything with release settings)

I have adjusted the makefile to do the copying at the end.

leezer3:
See the two pre-build events here for how the assemblyinfo.cs is being generated at present:
https://github.com/leezer3/OpenBVE/blob/master/openBVE/OpenBve/OpenBve.csproj#L330

Should work, although the events need adding to the csproj file manually (Or using Xamarin studio)

Sirflankalot:
Thanks for the hint, I've put it in the appropriate places.

leezer3:
Some test results:
VS2013 (Win)- Builds OK.
Xamarin Studio (Win)- Builds OK.
Monodevelop (Ubuntu 15.04)- 
https://github.com/Sirflankalot/OpenBVE/blob/master/source/OpenBveApi/OpenBveApi.csproj#L97
Needs to be cross-platform, as xcopy isn't available on Linux.

Makefile:
```
error CS0006: Metadata file `cscompmgd.dll' could not be found
Compilation failed: 1 error(s), 0 warnings
makefile:189: recipe for target 'bin_debug/OpenBveApi.dll' failed
make: *** [bin_debug/OpenBveApi.dll] Error 1
```

Change the definition for CSC to:
```CSC := gmcs -pkg:dotnet -lib:/usr/lib/mono/4.0```

Just using mcs doesn't work, as for some reason it just searches the defined lib folder (Rather than backwards looking to the 2.0 folder, which is where cscompgd.dll lives...)

Haven't tried a Mac build yet.

**Comment by Sirflankalot** [comment]
Alright, thanks for testing that.

I should have addressed the cross platform issue in ddd127e it appears to work on my machine.

I tried to test with monodevelop myself, but I'm having an unrelated issue and haven't been able to get xbuild to not issue the /nostdlib command yet on any C# project. I'll see if I can look into fixing it.

With regards to the makefile problem, I have adjusted the command. It compiles cleanly on Ubuntu 16.04. I've encountered this issue while compiling on a raspberry pi before. When I get home I will spin up a 15.04 VM to test it, but if you could test this new change as well that'd be great. What version of mono are you running? For reference I'm using 4.2.1.

**Comment by leezer3** [comment]
The Travis error looks like it's looking for the method in the wrong .Net version, which is why I dumped mcs on 15.04 in favour of gmcs
This means that I think there's a bug in mcs somewhere with dependancy resolution, when the assembly only exists in one .Net version....

gmcs certainly works on 15.04, will have to spin up some more VMs for actually supported versions, as that doesn't really help.
Should only have mono-complete (2.0 and 4.0) installed, but I'll have to check that, as this machine is somewhat of a test dump.

If you want gmcs the package is mono-gmcs (Part of mono-complete)

**Comment by Sirflankalot** [comment]
That's really bizarre. 

gmcs isn't available anywhere in the 16.04's mono-complete as I already had it installed. What's bizzare is that travis-ci is failing even though my version is compiling just fine. I know travis is running on 12.04 LTS so that may be part of it.

**Comment by leezer3** [comment]
Getting better :)
Now builds on my test boxes OK.

Some more issues:
* Changing the target via the makefile doesn't work unless make clean is done first.
* macbundle.tgz is copied from the dependencies on all machine types. TBQH, the Mac app build should be done in the makefile now, rather than hacked in as a travis command. (I stored the blank app in a tgz because it was easier keeping the permissions and structure intact that way from a Windows box, although that's rather yuck)
* Minor- if the output directory is inaccessible (permissions), the make command doesn't complain and appears to complete with no errors.

**Comment by Sirflankalot** [comment]
Yay! Great to hear!

Regarding the issues.

- I honestly haven't a clue how to fix that. I may be able to do something where if it detects a change in a sentinel file and make it touch every file to force a restart, or I could have it call itself with --always-make. I'll see what I can do, but don't expect anything soon, nor anything fancy.
- Alright, I'll look at the commands being used and put together a mac app build. If you have a sec could you just give me a rundown of what happens? I've never actually done mac building before at all (plus I don't one). I will look at what is there to try to implement something but it will take some trial and error without some help. 
- Should be easy to fix.

I am adding some commits to increase the amount of tests we do on travis-ci to test more edge cases (including one that uses gmcs to build!).

**Comment by leezer3** [comment]
Short answer is not much to a Mac app.
They're godawful programmatically, but that applies to a lot on a Mac!


They're just a folder with some 'magic' files inside.
The main executable within the app is  just a shell script which checks for Mono & launches it.

**Comment by Sirflankalot** [comment]
Yay! Sounds great! I love godawful things. I'll start implementing it as soon as I can.

**Comment by Sirflankalot** [comment]
It turns out that I lied, I have found an elegant way to deal with both debug and release builds. The solution was to have multiple targets both do the same thing allowing it to evaluate target doneness separately. It even handles building both debug and release at the same time! The solution is rather complicated and you can see the details in full in the commit here 944a4ad, but I'm pretty proud of it. When you get a chance, would you try both this set of commands and how you're planning on using it?

```
make debug
make release
make clean-all
make debug release
make clean-all
make all-debug 
make all-release
make clean-all
make all-debug all-release
```

I'll work on getting the app building into the makefile tomorrow.

Sirflankalot:
Alright, I believe I have solved all the problems you brought forth.

- As mentioned above I fixed the debug/release build thing.
- I have moved the building of the tar.gz or .dmg into the makefile, with all that travis-ci is doing now is renaming the output to the proper thing and uploading it.
- I've tested this by chowning bin_release/ to root and running the makefile. The makefile dies almost immediately as it tries to do various things that aren't allowed. 

Travis-ci is having a lot of backlog on their mac builds, so expect it to be an hour or so until it's finished, though it built successfully on my branch.

leezer3:
Seems to work on my VMs, so that's good.

Personally, I'd **always** build the app when on a Mac, but I don't know that's necessarily a good idea.

Further small points:
* Please exclude the temp files (linuxbuild / macbuild) from GIT.
* nuniversalchardetct's licence is missing, which is my fault though. Techincally, we're complying with the LGPL and MPL (And GPL too....), but for the purposes of this project, I think MPL is strongly preferable as the least restrictive.
* Again, somewhat of an oversight on my part, but the **Data** folder ( /source/OpenBVE/Data ) should probably be in the root folder. Seeing as we're doing a re-arrange, this probably should be moved now :)

When that's done, I think this is relatively ready for merge.
Any chance of you squashing all these changes into a single commit relative to the current master, and either re-creating your pull branch or opening a separate pull?
I can do that if necessary, but 65 commits is a little OTT :P

Sirflankalot:
As soon as I'm back on my computer I shall make the final changes. I think having a non app version is useful for debugging in a debugger IMO. 
 
I'll add a separate pull for all of these changes as 65 is crazy indeed! 
 
I'll move the folder to the appropriate place and rename it to something like assets (though once built it will still be data. 
 
Glad to hear it's almost ready :) 
 
Expect to see changes in around 3 hours.

leezer3:
One small final one:
sshpass is not installed by default in the Mac environment, and you've the brew install for it from the travis YML file.

Technically ought to change to keypairs for the push, but I was having trouble getting travis to validate self-signed certs properly.

Sirflankalot:
It's actually already there, just in the include matrix instead of the after_sucess section.

leezer3:
Sorry, you're entirely right (Missed that....), will try and merge the single commit PR tomorrow.

I'll have a play with a local Travis-CI worker and try and get keypairs working too, but that can wait.

Sirflankalot:
No worries, that's fine. I actually have to work out some weird dependency crap with the makefile tonight (overnight your time). There's a situation where if you have really fast compilation, slow copies, and multiprocessor compilation it might be possible to hit the linking of OpenBveApi.dll before the other .dlls have finished copying, throwing an error. This is because the current version of the makefile doesn't have OpenBveApi.dll depending on the copied .dlls. The version that I put in the single commit build does have the dependency work right, but it depends on a .PHONY causing recompilation every time. I'm about half way done with implementing a version where it depends on things properly and doesn't have to depend on a .PHONY. I don't expect this to break any builds as I'm not actually editing the commands at all, merely changing the dependency order. I will push these commits into this PR and then make a second combined commit for the single (or now double) commit pull request.

As an aside, how do you get a local Travis-CI worker‽ That would be very useful to not have to commit every single time I want to test something on travis.

Sirflankalot:
I have added the necessary bug fixes to make the makefile work correctly in all situations. Commits have been submitted in the appropriate places. There has been no real change in the building, only in dependency resolution. This should compile as cleanly as it did before.

leezer3:
PR merged, and has built correctly :)

Travis-CI itself is a slightly interesting setup.
The workers themselves are really just a base Ubuntu image with some preconfigured packages per-language and a whitelist for whatever else you want to install.
travis-build compiles the YML into a standard .sh script, which is then just uploaded to the VM.
https://github.com/travis-ci/travis-build

Linked from there is also a ruby-based management interface, which works in approximately the same way the web one does, but that's not especially necessary if you've got the script.
(Haven't tried the web interface...)

Sirflankalot:
Yahoo!

That looks pretty interesting, I'll look into getting some kid of setup together when I get a bunch of time.

PR #122, opened 2016-12-15
Initial implementation of the Compatibility Object Manager

leezer3:
This implements an initial WIP for the Compatibility Object Manager.

The _Data\Compatibility\CompatibilityObjects.xml_ file provides a list of available objects for replacement, and a mapping to their on-disk file.
**Sample:**
```XML
  <Compatibility>
  <!-- Uchibo Route Objects -->
	<Object>
		<Name>Ballast.csv</Name>
		<Path>Uchibo\Ballast.csv</Path>
		<Message></Message>
	</Object>
</Compatibility>
</openBVE>
```
**Name** represents the path of the original object within the Railway\Object\ folder.
**Path** represents the relative path to the compatibility replacement within the Compatibility folder.
The string in the _Message_ tag will be added to the log when an object is used, if it is not empty.

**TODO:**
* Improve logging when compatibility objects are used.
* ~~Allow the base XML file to point to additional XML files.~~
* ~~Allow one base object entry to point to multiple replacements.~~
* Add 'override' tag to allow an object to override the originals.
* Implement in Route Viewer (Raise errors in Route Viewer only- Using these is probably bad practice...)
* TEST!!!


leezer3:
OK, so this has been tested, and is reasonably functional.
The only thing missing from the Uchibo base objects (As opposed to the route-specific objects) is the people, which is due to a lack of appropriately licenced images. (Decent cutout people are a lot harder to take than just plain textures)

Have changed the base XML to a list of subordinate package XMLs, although it'll still take object entries too.

Still requires adding to RouteViewer, but that's a relatively quick process.

Sirflankalot:
Is this patch intended to solve #124 at this point? I just tested it and still fails. An exception is thrown at source/OpenBVE/OldCode/FunctionScripts.cs:1377 if that's of any help. If this isn't intended to do that, disregard.

leezer3:
It does on my Windows box, but I haven't specifically tested on Linux/ Mac.

FunctionScripts.cs suggests a broken animated file, and is something else than anything this was intended to fix.
I'll take a closer look when I'm home, but there's the possibility it's coming from one of those 135 files I'm missing....

PR #125, opened 2016-12-16
Fix or Silence all Warnings

Sirflankalot:
In this pull I have all of the different warnings being either fixed or silenced. There were huge sections of problems on MSVC#, with some files exceeding 300 variable unused warnings. I have silenced them on mass and have removed unnecessary variable declarations. I have made sure that any function that had side effects still gets called, the return value is just thrown away. I have left the variable name in a comment next to it so that if we need this information later it would be easy to put back in. Most of the problems were in the ObjectViewer/RouteViewer, but there were a few things in the OpenBVE source.

### DIFF
--- a/source/OpenBVE/UserInterface/formMain.Packages.cs
+++ b/source/OpenBVE/UserInterface/formMain.Packages.cs
@@ -765,7 +765,7 @@ namespace OpenBve
 				if (!Program.CurrentlyRunningOnMono)
 				{
 					//Mono doesn't support System.Security.AccessControl, so this doesn't work....
-					System.Security.AccessControl.DirectorySecurity ds = Directory.GetAccessControl(directory);
+					// System.Security.AccessControl.DirectorySecurity ds = Directory.GetAccessControl(directory);
 				}
 				using (FileStream fs = File.OpenWrite(currentPackage.FileName))
 				{
--- a/source/OpenBVE/UserInterface/formMain.Packages.cs
+++ b/source/OpenBVE/UserInterface/formMain.Packages.cs
@@ -765,7 +765,7 @@ namespace OpenBve
 				if (!Program.CurrentlyRunningOnMono)
 				{
 					//Mono doesn't support System.Security.AccessControl, so this doesn't work....
-					// System.Security.AccessControl.DirectorySecurity ds = Directory.GetAccessControl(directory);
+					Directory.GetAccessControl(directory);
 				}
 				using (FileStream fs = File.OpenWrite(currentPackage.FileName))
 				{
--- a/source/OpenBVE/UserInterface/formMain.cs
+++ b/source/OpenBVE/UserInterface/formMain.cs
@@ -158,6 +158,7 @@ namespace OpenBve {
 			Image Logo = LoadImage(MenuFolder, "logo.png");
 			if (Logo != null) pictureboxLogo.Image = Logo;
 			string flagsFolder = Program.FileSystem.GetDataFolder("Flags");
+	#pragma warning disable 0219
 			string[] flags = new string[] { };
 			try
 			{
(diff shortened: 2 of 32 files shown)


source/OpenBVE/UserInterface/formMain.Packages.cs
```diff
@@ -765,7 +765,7 @@ private void buttonCreatePackage_Click(object sender, EventArgs e)
 				if (!Program.CurrentlyRunningOnMono)
 				{
 					//Mono doesn't support System.Security.AccessControl, so this doesn't work....
-					System.Security.AccessControl.DirectorySecurity ds = Directory.GetAccessControl(directory);
+					// System.Security.AccessControl.DirectorySecurity ds = Directory.GetAccessControl(directory);
```
Sirflankalot: Would it work if we called the function but just didn't assign it to a variable?

source/OpenBVE/UserInterface/formMain.cs
```diff
@@ -158,6 +158,7 @@ private void formMain_Load(object sender, EventArgs e)
 			Image Logo = LoadImage(MenuFolder, "logo.png");
 			if (Logo != null) pictureboxLogo.Image = Logo;
 			string flagsFolder = Program.FileSystem.GetDataFolder("Flags");
+	#pragma warning disable 0219
 			string[] flags = new string[] { };
```
Sirflankalot: Alright, should we do anything about them at this point?

source/Sound.Flac/Decoder.cs
```diff
@@ -179,7 +179,7 @@ internal static class Decoder {
 					 * block may be shorter than the stream blocksize; its starting sample number will be calculated as the frame number 
 					 * times the previous frame's blocksize, or zero if it is the first frame). 
 					 */
-					ulong sampleOrFrameNumber = (ulong)reader.ReadUTF8EncodedInteger();
+					reader.ReadUTF8EncodedInteger(); // ulong sampleOrFrameNumber = (ulong)
```
leezer3: We don't support variable frame lengths in the flack decoder plugin.

I suspect it'll fall over nastily if these are used, but haven't ever got as far as trying to debug it properly. (Know nothing about FLACs...)

source/OpenBVE/UserInterface/formMain.Packages.cs
leezer3: Not sure, would have to check that.

Sirflankalot: I don't see why it wouldn't as variable assignment can't throw.

leezer3:
Some observations above on the purpose of a few chunks of code (to the best of my understanding anyways)

No idea what to do with them, but still!

source/Sound.Flac/Decoder.cs
Sirflankalot: Yup, but at least until we can fix it, we still have to read the value and discard it so that everything else is in the right place.

Sirflankalot:
I've added a commit that removed the problems with the exception throwing function. A lot of the parsers need to be rebuilt from the ground up, but that's not in scale for this commit. My two specialties are compilers and graphics, so I can help rewrite the parsers to work better after I've dealt with graphics.

Sirflankalot:
I've added todos regarding what we've talked about as notes for the future.

Sirflankalot:
I'm using this pull to also add a very minor problem in the makefile where the data is properly copied when compiling both debug and release at the same time.

leezer3:
Cheers.

I'll try and get a list together of all the incomplete/ broken features that I'm aware of, and start an issue or something, as I suspect that would be helpful.
As you look through the code, you'll find a **lot** of todo messages (mostly) left by Michelle, but some of them are mine too....

Sirflankalot:
How does a wiki page sound to you?

PR #129, opened 2016-12-19
Fix Monodevelop on Fresh Build and Add Travis-ci Tests

Sirflankalot:
I've fixed problems with dependency copying on xbuild, as well as adding travis-ci tests on all platforms using xbuild.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,16 +9,16 @@ matrix:
     - os: linux
       dist: trusty
       mono: latest
-      script: make publish
+      script: make all-release
     - os: osx
       mono: latest
       before_install: brew install hudochenkov/sshpass/sshpass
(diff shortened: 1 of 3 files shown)


leezer3:
Minor thought:
Calling make publish on all Linux configurations could potentially cause a race condition when uploading.

Probably solution is to change all Linux configurations other than the latest to simply run make?

Unfortunately, I don't think travis will let us run a command within a build VM after all other builds have passed. They've got a webhook after all successful builds. but I think that building something along these fronts is OTT.

Sirflankalot:
That would be a smart idea, I'll adjust accordingly. 

I'll change the rest to run `make all-release` which is what publish depends upon. 

Building that just sounds like a PITA.

PR #135, opened 2017-01-01
Switch daily builds to use .zip for a wider audience

Sirflankalot:
When I was linking people to the daily builds, they often didn't know what to do, as they had never seen .tar.gz before (obviously non-programmers).

I think, while we don't have a automated windows build, we should have the linux build (which works just fine on windows) push a .zip. This way people can just click the zip if they have windows/linux, and the .dmg if they have mac. Obviously the full linux build will be .tar.gz but that can be done easily manually.

**Edit:**  There is a disadvantage however, becasue .zip isn't a solid archive, we go from ~7Mb to ~13Mb

**Edit 2:** We could also do .7z which is familiar and gets us some of the best compression, sitting at 5.9Mb

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,8 +15,8 @@
 bin_debug/
 bin_release/
 obj/
-linuxbuild.tar.gz
-macbuild.dmg
+linuxbuild.*
+macbuild.*
 mac/
 
 # Misc. Visual Studio files
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,6 +64,7 @@ addons:
   apt:
     packages:
       - sshpass
+      - p7zip-full
 env:
   global:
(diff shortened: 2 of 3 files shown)


leezer3:
Hmm, possible alternative is to use 7z to create the zip?

This gets us a 6.5mb zip:
```7z a -tzip -mm=lzma -mx=9```

Sirflankalot:
I mean it is, but zips compressed with anything except deflate aren't going to open with native tools. I know for Windows you have to use deflate and I'm fairly positive it's the same on mac. 

It really sucks but we'd have to switch to .7z if we want to use something other than deflate.

leezer3:
Windows explorer definitely supports LZMA zip files, and from a quick test BZIP2 and PPMD seem to work as well.
If I recall, this was patched several years ago.

Archive utility doesn't support anything other than deflate, but as DMG files are being built as well I don't really see that as too big a deal.

Sirflankalot:
Holy crap, I had no idea they had fixed that, that makes our lives so much easier. And to believe they back ported the fix as well... 

Pushed a commit with the changes we talked about.

leezer3:
I'm going away to hit my head on a wall with this one......

Just had a report that LZMA zips don't extract with the default Linux unzip, and to make matters worse, p7zip doesn't want to extract SHIFT_JIS standard encoded zips properly either. (Messes up the charset....)

Going to roll back to deflate for the minute.....

Sirflankalot:
Oy vey. If only we lived in a world where everything supported everything they should, but whatever. Sounds a plan.

PR #137, opened 2017-01-11
Compatibility: Framework for replacing Win32 plugins with a suitable alternative

leezer3:
This pull introduces a database of compatible replacement plugins for use on non Win32 systems.

Currently, one plugin is supported: BVEC_ATS, which provides support for several basic variants of OS_ATS.

**Usage:**

- A Win32 plugin may be assigned a compatible replacement via __Data\PluginDatabase\CompatibleReplacements.xml__  The file length, filename and MD5 must match that stored in the database.

- A faulty plugin may also be blacklisted on all operating systems, using the file __Data\PluginDatabase\BlackList.xml__

**Other changes:**

- Fixes a bug whereby if no plugin was loaded, the controls would be stuck.

- If a .Net plugin crashes in-game, the plugin will be cleanly unloaded and a crash log created.


leezer3:
Current issues with this pull:

- No handling of situations where this is no write access to the train directory. (BVEC_ATS needs an in-memory configuration option if this is the case)

- Needs to support copying across a new configuration file, assuming that we have write access. Doing this will allow UKTrainSys to be used.

- Further to the above, requires a whitelist option for specific train/ plugin combos. UKTrainSys only supports a small subset of trains, with specific configuration files, wheras BVEC_ATS has been designed to work with all basic OS_ATS variants.

leezer3:
The csproj build on Linux will spit out an error, due to the following commit being missing:
https://github.com/leezer3/OpenBVE/commit/19960af474ded0e8e01335329e6d7e6ea8c9ac77

Ignore this, will be OK once merged into master....

leezer3:
Some more work, aimed at allowing compatibility plugins to be installed as a package, rather than by default.
This helps with potential licensing issues & reduces the download size, as it's only Linux users who'll want this sort of thing.

TODO:
* Finish the routine that allows us to write out the blacklist DB and compatibility DB.
* Add interface for querying the openBVE website when a train with an incompatible plugin is loaded. (?? load factor ??)

Have also received a request from BVE4 Trains *not* to include his stuff in any default replacement DB.

leezer3:
Packages now write the appropriate blacklist/ replacement database files.

TODO:

- Uninstall- The files are uninstalled, but the XML is not.
- Packages listed in the 'other' section do not uninstall correctly.

leezer3:
Some more work, and have squashed the branch into a single commit.
TODO:
- Only the blacklist DB is currently updated on package install. This needs to be extended to the replacement DB and the compatibility objects DB.
- Still a little buggy, as the database can acquire duplicate entries when installing a package multiple times.
- Chuck the current replacement plugins out of the main source and into a package.

leezer3:
Couple more bugs with packages....

leezer3:
Have tidied the PR up (again.....)
Current TODO:
- The default database path for these should probably be under UserData somewhere, as if installed under Program Files, we won't have write access here by default.
- Fix BVEC_ATS so that it doesn't barf when there is no write access to the train directory (See above)

A sample package, which implements BVEC_ATS for OS_ATS version 1.0, 1.1, 1.2 and 1.3, and UKTrainsys for the 153 & 158 (original TSC versions)
[Replacement Plugins 1.0.zip](https://github.com/leezer3/OpenBVE/files/801253/Replacement.Plugins.1.0.zip)

leezer3:
https://bveworldwide.forumotion.com/t2093-bvec_ats-dll-failed-to-load#20619

Need to revisit this with 64-bit / Linux.
Database package in hindsight was probably OTT, can just update with new builds.
Easiest way to do things could well be to have a modified version of BVEC_ATS which will read an OS_ATS config file directly, as opposed to a regen of the config file.

Possibly hook in more ATS components to the main game / inbuilts too. Would it be worth doing a basic port of the AWS to OpenBveAts??

PR #138, opened 2017-01-16
Tidied up 2 things

Gudbrandr:
* Removed all copies of the icon except the one in /assets and pointed all projects there.
* Prevented warning when compiling the release builds by documenting some parameters in Colors.cs.


Gudbrandr:
It seemed silly to have 6 copies of the icon now the applications are all bundled together.

Gudbrandr:
Fixes #139. Fixes #140.
The package manager fixes were supposed to be in a separate branch but they got caught up in trying to merge your versions of my changes somehow. Sorry.

leezer3:
Hmm, your merge procedures could use a little work, this is one confused little pull request :)

Try doing it this way:
Add this repository as a remote.
Checkout my master branch.
Create your own new branch, and name appropriately.
Work on this branch.
Finally, create a pull relative to my master.

(Forget that master ever exists in your own repository.....)

In essence, there should only ever be ONE master, and all work should be done relative to it.
Also, remember that GIT will merge the two branches, so unless there's a commit you specifically need which isn't in your branch, never create a pull to your repository.
In 90% of cases where you want something like this, you'll find that stashing your current branch, checking out a fresh copy of master and popping your stash will avoid even this pull.

Please don't take this as a criticism, just trying to keep down on the commit noise.

leezer3:
https://github.com/leezer3/OpenBVE/commit/24308e290df32471818208ddd025c419fa3c40e4

I think that fixes those :)

Gudbrandr:
No problem at all. I'm fairly new to Git and have little experience with branching and merging because all the companies I've worked for work in trunk (!), so it's something I'm trying to learn.

I'll try your suggestion with a small change as a test.

Thanks.

PR #141, opened 2017-01-20
New: Sounds can be attached to a TrackFollower

leezer3:
This branch has the initial work for attaching sounds to a Track Following animated object.

Sample animated file to start with:
```
[Object]
position = -4,0,0
States = 16xx.b3d
TrackFollowerFunction = if[value == 0, value + 1000, if[trackdistance > -200,if[value < 0, value, value - 10 * delta], value]]
Sound=Steamer.wav
```

The sound for this is played looped at a constant 100% volume & pitch when the object is visible (Stopped if the state is -1), emanating from the object's front axle.
The sound radii at present is a constant 30m (As per other trains within the simulation), but before merging, this should be changed to allow the user to set the radii appropriately.

Bugs:
* ~~Something not quite right with the sound attachments at the minute, occasionally managing to get the wrong type set for a sound....~~

### DIFF
--- a/source/ObjectBender/ObjectBender.csproj
+++ b/source/ObjectBender/ObjectBender.csproj
@@ -46,7 +46,7 @@
     <PlatformTarget>x86</PlatformTarget>
     <FileAlignment>4096</FileAlignment>
   </PropertyGroup>
-  <Import Project="$(MSBuildBinPath)\Microsoft.CSharp.Targets" />
+  <Import Project="$(MSBuildBinPath)\Microsoft.CSharp.targets" />
   <ItemGroup>
     <Reference Include="System" />
     <Reference Include="System.Drawing" />
--- a/source/ObjectViewer/ObjectViewer.csproj
+++ b/source/ObjectViewer/ObjectViewer.csproj
@@ -65,7 +65,7 @@
     <Reference Include="System" />
     <Reference Include="System.Drawing" />
     <Reference Include="System.Windows.Forms" />
-    <Reference Include="System.XML" />
+    <Reference Include="System.Xml" />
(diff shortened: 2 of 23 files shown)


leezer3:
Video here:
https://youtu.be/9LNKSVEPem4

**TODO:**
* Allow sound radius to be changed. (30 is the default for train based sounds, but it's IMHO way too low)
* Allow sounds to be attached to any animated object.
* Investigate changing pitch/ volume via a function. Probably not for this merge, as I suspect it'll require a bunch of nasty function related work....

leezer3:
So, the first draft of an XML for loading sounds has been implemented.
```XML
  <Sound>
	<File>Steamer.wav</File>
	<Radius>60.0</Radius>
	<VolumeFunction>if[trackdistance &gt; -100, 4,1]</VolumeFunction>
  </Sound>
</openBVE>
```

This has only been done for animated files at present, but has one flaw: XML files don't support entering the **<** and **>** symbols directly.
Thinking about this.....

leezer3:
A new and revised version.

This allows an arbritary sound to be added into an animated object.
Sample:
```
[sound]
filename = steamer.wav
trackfollowerfunction = -mod[10 * time, 1000] + 5
volume = 0.5
```

Supported Attributes:
* TrackFollowerFunction - Moves along the path of Rail 0
* VolumeFunction - Alters the volume
* PitchFunction - Alters the position
* Position - Sets the base position of the sound

leezer3:
Not quite sure why the MSBuild workers barfed at the minute, the log suggests that it couldn't find the targets file though.
Used to work, so probably a Mono regression, but at the same time the case is wrong.....

PR #148, opened 2017-03-01
Use built in tools for .tt translation instead of custom build step.

Sirflankalot:
Sorry about not being as active lately, some things came up in my life that kept me away from development for a long while. 

In this interim I installed VS2017RC and I found that the .tt translation commands we had been using weren't working. I did some research and found there is actually a way to have them autotranslate themselves by adding some stuff to the .csproj file. This will allow greater portability as it doesn't tie to any specific version of VS as well as not needing to debug the commands we were using. 

On unix, the .sh file is called as normal. 

Note: S3 is having some problems so Travis-CI is too. I've tested on both windows and linux and this works as expected.

### DIFF
--- a/source/OpenBVE/OpenBve.csproj
+++ b/source/OpenBVE/OpenBve.csproj
@@ -344,7 +344,7 @@
     <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
   </PropertyGroup>
   <!-- This is the important line: -->
-  <Import Project="$(VSToolsPath)\TextTemplating\Microsoft.TextTemplating.targets" />
+  <Import Condition=" '$(OS)' != 'Unix' " Project="$(VSToolsPath)\TextTemplating\Microsoft.TextTemplating.targets" />
   <!-- Details about translation -->
   <PropertyGroup>
     <TransformOnBuild>true</TransformOnBuild>
@@ -354,6 +354,9 @@
   </PropertyGroup>
   <PropertyGroup>
     <TransformOutOfDateOnly>false</TransformOutOfDateOnly>
+  </PropertyGroup>
+  <PropertyGroup>
+    <PreBuildEvent Condition=" '$(OS)' == 'Unix' ">bash $(ProjectDir)Properties/AssemblyInfo.sh</PreBuildEvent>
   </PropertyGroup>
   <PropertyGroup>
     <!-- Copy in dependancies and child projects -->
--- a/source/OpenBVE/OpenBve.csproj
+++ b/source/OpenBVE/OpenBve.csproj
@@ -328,12 +328,32 @@
   <ItemGroup>
     <Service Include="{508349B6-6B84-4DF5-91F0-309BEEBAD82D}" />
   </ItemGroup>
+  <ItemGroup>
+    <Content Include="Properties\AssemblyInfo.tt">
+      <Generator>TextTemplatingFileGenerator</Generator>
+      <LastGenOutput>AssemblyInfo.cs</LastGenOutput>
+    </Content>
+  </ItemGroup>
   <Import Project="$(MSBuildBinPath)\Microsoft.CSharp.targets" />
+  <!-- AUTOMATICALLY TRANSLATE TEXT TEMPLATES -->
+  <!-- Optionally make the import portable across VS versions -->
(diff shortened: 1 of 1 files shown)


leezer3:
No trouble.

Whatever is contributed is welcome, and I'm not doing anything related to graphics at the minute, so it's not stopping me doing anything :P

leezer3:
Interesting.

Your solution seems to fail on systems without the VS Modelling SDK installed for the current VS version.

I've done a little hacking, which seems to work for me, by testing for the existence of the VS modelling SDK project first, then falls back to the older code if this isn't found:
```<Import Project="$(MSBuildBinPath)\Microsoft.CSharp.targets" />
  <!-- AUTOMATICALLY TRANSLATE TEXT TEMPLATES -->
  <!-- Optionally make the import portable across VS versions -->
  <PropertyGroup>
    <!-- Get the Visual Studio version – defaults to 10: -->
    <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">10.0</VisualStudioVersion>
    <!-- Keep the next element all on one line: -->
    <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
  </PropertyGroup>
  <!-- This is the important line: -->
  <Import Condition=" '$(OS)' != 'Unix' and Exists('$(VSToolsPath)\TextTemplating\Microsoft.TextTemplating.targets')" Project="$(VSToolsPath)\TextTemplating\Microsoft.TextTemplating.targets" />
  <!-- Details about translation -->
  <PropertyGroup>
    <TransformOnBuild>true</TransformOnBuild>
  </PropertyGroup>
  <PropertyGroup>
    <OverwriteReadOnlyOutputFiles>true</OverwriteReadOnlyOutputFiles>
  </PropertyGroup>
  <PropertyGroup>
    <TransformOutOfDateOnly>false</TransformOutOfDateOnly>
  </PropertyGroup>
  <PropertyGroup>
	<PreBuildEvent Condition=" '$(OS)' != 'Unix' and !Exists('$(VSToolsPath)\TextTemplating\Microsoft.TextTemplating.targets')">set textTemplatingPath="%25CommonProgramFiles(x86)%25\Microsoft Shared\TextTemplating\$(VisualStudioVersion)\texttransform.exe" -if %25textTemplatingPath%25=="\Microsoft Shared\TextTemplating\$(VisualStudioVersion)\texttransform.exe" set textTemplatingPath="%25CommonProgramFiles%25\Microsoft Shared\TextTemplating\$(VisualStudioVersion)\texttransform.exe" -%25textTemplatingPath%25 "$(ProjectDir)\Properties\AssemblyInfo.tt"</PreBuildEvent>
    <PreBuildEvent Condition=" '$(OS)' == 'Unix' ">bash $(ProjectDir)Properties/AssemblyInfo.sh</PreBuildEvent>
  </PropertyGroup>```

Can you test with 2017 at all?

leezer3:
Another couple of potential solutions:
- Bundle TextTransform.exe with the source- The mono version is MIT/X11, so that would work just fine.
- Write a bat / Windows script of some description to do it without reference to VS.
- Write our own generator for the text transform. About 5 mins work.

Sirflankalot:
Yeah I noticed that too, I had to install it for it to work on 2015. Unfortunately my 2017 RC install is messed up (I actually had to [file a bug](https://developercommunity.visualstudio.com/content/problem/23183/installation-failing-because-of-invalid-sourcedest.html) about it) and like half the profiling tools didn't install at all. As far as I know is that .tt translation is included in the default 2017 install. Ill test it with my installs and see if it works. 

I think however, the best option for the job is going to be to write our own translation script as either a batch script or some c# that gets executed one off. I can work on writing a batch script once I'm off my linux machine and I'm back home. Budling TextTransform.exe probably isn't allowed, and it's also clunky

leezer3:
We can bundle the Mono texttransform (MIT/X11), but I'd presume not the Microsoft one.

Sirflankalot:
Either way, I'd perfer not to bundle in _more_ crap than we have to. On a slightly tangential note, shouldn't we increment the version number whenever a release comes out?

Sirflankalot:
Your solution worked perfectly fine on VS2015 and VS2017 RC. I'm going to say we stick with that and move on. I started to write a batch script, but got stuck on all the shinangians you have to go through to even get the date in seconds. It's a might ridiculous.

PR #149, opened 2017-03-02
Fix problems with building without the VS Modelling SDK.

Sirflankalot:
Also fixed sh formatting. See #148


Sirflankalot:
Checks failed due to no fault of our own.

PR #152, opened 2017-03-05
New: Allow horns to have start and end sounds

leezer3:
Not quite convinced this one is perfect yet, holding off for a few days before committing.

**Summary:**
Allows the horns to have a start and stop sound if appropriate:
```
[Horn]
PrimaryStart = PrimaryHornStart.wav
Primary = PrimaryHorn.wav
PrimaryEnd = PrimaryEnd.wav
```

An alternate spelling of PrimaryLoop, SecondaryLoop and MusicLoop is accepted.

**Techincal:**
Moves the horn processing from the main control loop, to a function within the horn struct.


leezer3:
Have now separated out the Horn into it's own file, still thinking about other tweaks....

leezer3:
I knew there was something!

Added the following subjects to those available for animation:
**PrimaryKlaxon** : Returns 1 if the primary horn is playing, 0 otherwise.
**SecondaryKlaxon** : Returns 1 if the secondary horn is playing, 0 otherwise.
**MusicKlaxon** : Returns 1 if the music horn is playing, 0 otherwise.

All klaxon subjects are also now available for panel2.cfg based trains.

_TODO:_ 
* Update documentation
* ???

leezer3:
I think this is ready to merge.

I would bump the version of the sounds.cfg file as new parameters have been added, but this causes 1.4.3 to not load any sounds, whereas extra parameters do not.

XML sound configuration is next on the list.

MarcRiera:
I'm getting an "Object reference not set to an instance of an object" error when using "PrimaryStart" but not defining "Primary" and "PrimaryEnd". It happens only when I hold down the horn key after the sound has ended, not if I release it before it ends, so I suspect only "Primary" is broken.

leezer3:
One always sneaks through, entirely my fault....
Was too busy testing the loop start code to notice the missing null check.

It's the start sound not being defined that does it.
https://github.com/leezer3/OpenBVE/commit/d278797267e620518917fbb045e066da5d5edc38

Should fix as soon as things rebuild.

PR #153, opened 2017-03-07
Refactor the Brake System

leezer3:
This pull starts the process of massively refactoring the brake system.

The brake system has been separated out into appropriate classes, and the methods have been disentangled.

**TODO:**
- The air brake components should have their own root class, not live in the TrainManager
- Missing documentation coverage on a lot of values
- Can the brake system class be non-static?


Sirflankalot:
This looks like good work! I'm definitely moving away from having everything static so that we have more choices as to what to do in the future. I don't see the difference right now between having a single break system instance in the train manager and having it be static. With the benefits that non-static-ness gives us, I think we should strive for it. 

> The air brake components should have their own root class, not live in the TrainManager

Why do you have this in mind? I don't really know that section of the simulator that well right now (spending all my time in the graphics module) so I'm curious.

leezer3:
The eventual aim would be to move brake systems into their own independent assemblies.

Thus, the train manager would call the brake implementation for each car, each adding to the final deceleration figure.
Haven't got this quite plotted out yet, but separating into discrete components is the first step.

Sirflankalot:
Ah, didn't think about that, that's a very good idea.

leezer3:
Getting closer....

This moves the compressor and air sounds into the brake system namespace & FFWDs the source to the current base. (Will squash to a single commit before any eventual merge)

**Blocking Issues:**
* Sounds are not a shared resource (as per textures), but are stored on within the class. These need to be refactored into a sound manager, which simply calls the index. (Need to check whether this was done for an openAL specific reason, multiple copies of the sound playing at different time indicies???)
* Brake handle movements should possibly be moved into the class as well.  Not sure on this one, but it makes reasonable logical sense for the brake system assembly to handle this & the associated sounds.

PR #154, opened 2017-03-19
BVE5: Initial route parser

leezer3:
This is first attempt at a reasonably functional route parser for BVE5.
(Manually rebased, some more of the functions this needs have now been implemented)

Working OK:

- Most functions of legacy routes converted via BVETS Convertor

Not working

- Interpolation of curves, cant etc.
- ~~Some X files aren't interpreted correctly~~
- Repeaters with a length less than the current block length are broken on curves
- Various commands not implemented yet

Aln Valley

Keisei

Random route
[ci skip]


leezer3:
Currently in the process of merging the changes from master into this branch.

https://github.com/leezer3/OpenBVE/commit/7318a5ae95d60f9d7ec4c74b0a31dbb180f6f9e6
bve54

This gets a lot more X-objects to actually work successfully

leezer3:
bve55

Improvements from today's work above.
The crossing poles and most of the platform are now at least semi-working.

The positioning of various objects is now much better too.

leezer3:
Oops, not fixed the makefile to take account of the new directories....

Must do that :/

leezer3:
Have picked the X parser improvements and fast-forwarded this branch relative to main again.

Potentially needs merging behind a hidden option just to keep the codebase from diverging....

leezer3:
More issues fixed, primarily typos.

Repeater Type 0 also appears to still be partially broken.

leezer3:
Another FFWD relative to master, plus initial works towards getting curve radius parsed on secondary tracks.

Need to investigate exactly how this is supposed to work,...

leezer3:
Next job this one needs is for the Jerk and delay values to be replaced with arrays assigned to the power notch, so that we can have separate values on a per-notch basis.

These will need to be hooked into the train.xml parser at some stage, unfortunately it's not possible to extend the train.dat without breaking backwards compatibility in this case......

leezer3:
The most recent commit now should parse a lot of the basic data required from the BVE5 train formats.

### Missing:
* Panel configurations- These are a 90% variant on Panel2.cfg, I believe just a few additional subjects for the ammeters.
* Acceleration / brake curves.
* Motor sounds
* Loading sound configurations- Again appears to be a variant on sound.cfg


Not currently tested

leezer3:
This hooks up some more of the missing functions.
We still crash at the minute as the acceleration output appears to be broken.

leezer3:
Getting there.

We now load the train & panel OK. (NOTE: Only one specific train, no variants tested....)
No acceleration implemented, but the brake pressures are working as expected.

leezer3:
Further progress: The base acceleration logic is in place. and a train loaded from BVE5 will now move.

### Issues:
* Need to figure out exactly how to convert between the acceleration in N (BVE5) and km/h/s (openBVE)
* Separate out the deceleration from the acceleration
* Motor sounds not yet implemented

leezer3:
Deceleration logic has now been abstracted.

The primary remaining issue in this area is that we currently assume that our deceleration is proportional to the pressure ratio, as opposed to being able to load a complete deceleration curve.
This may require considerable rehashing, looking into it....

leezer3:
Some more fixes.
Trains converted through the BveTsConvertor now work reasonably as designed.

Native BVE5 trains generally work OK-ish, but the deceleration logic is not yet complete (see above), and the ammeter has not yet been properly tested or plumbed up to the panel logic.

### Other issues
* No documentation on exactly what has changed between file format versions. Some stuff on Mackoy's blog, and there are the **Up** labels in the documentation, but not clear sometimes.
* Suspension sounds have been split into apply and release. Check behaviour.

leezer3:
Another fast-forwards relative to the master branch.

[Doing these regularly might keep this mergable one day....]

Next move will be to cherrypick the abstracted acceleration logic into the main branch; In hindsight, should have done the abstraction on main or a separate branch, but it's so much easier to fiddle with and test this way.....
Following that, need to investigate exactly how deceleration values are calculated and used, and from there how we can provide a decleration value per notch without breaking things.

### Thoughts and Assumptions:
* __BrakeDecelerationAtMaximumServicePressure__ provides the max deceleration at full service brakes (Note not EB)
* This appears to be multiplied with the current proportion of pressure in the BC to produce the deceleration figure.
* BVE5 provides a list of figures for notch & speed (Assume that the final one is used if notch is greater than list of values?), interpolation between them?
* What uses the figure also, and will changing it from the proportions above bug out the brake systems or other physics somewhere? Can't think of anything immediately, we're not changing the amount of air in the BC or anything.
* Would be nice to implement something like this in the new format, investigate exactly how it should be done....

leezer3:
Squelched all the train related commits and fast-forwarded again after the refactoring work on the main branch.

### TODO NEXT:
* Get the brake systems branch merged into mainline, as we need various of those properties to be settable which were fixed before.
* Deal with the brake deceleration, as above.

leezer3:
Motor sounds are going to be a pain.

BVE5 allows an unlimited number of motor sounds.
No documentation on them, so this is backfudged:

### Assumptions:
* Speed is the first entry in the motor sound table ==> Seen a couple with a small negative as the first option; Ignore....
* Speed does not have to go in .5km/h increments (as per BVE2 / BVE4), linear interpolation seems to be being used instead.
* Each value after the first represents either the pitch or the volume of the sound at the given speed setpoint.
* Itinerate backwards through the split line until we meet a non-empty entry.

PR #156, opened 2017-04-28
Implement Alternative Route Messages

leezer3:
This is a WIP framework for implementing messages of all kinds generated by the routefile.

**Changes:**
* Moves markers and game-generated messages into a new class __MessageManager__
* The message Manager contains an abstract __Message__ definition, plus definitions for each type of message.

**TODO:**
* Split the individual message classes into separate child files.
* Implement the new texture / text based messages into the routefile. This will require a new sub-file to hold the early/ late/ on-time textures/ strings.
* Add new command to the routefile __TextMarker__ : Plain text marker, to work as per the original marker images.
* Clean existing code.....


leezer3:
Hmm, need to do some thinking on this one.

Abstracting & combining messages/ markers brings some somewhat interesting problems with regards to the renderer, as they're somewhat disparate objects, and require totally different rendering approaches.

This is easy enough, but the positions are all relative, hence the abstraction idea, which then falls over, as too many properties must be added to the abstract class.....

leezer3:
Not there yet...
Separated a bunch of stuff into some more logical functions within the renderer, but haven't got a solution to the abstraction problem at the minute that I like at the minute.

Best current idea is to create two new lists within the renderer itself:
* Text messages
* Markers/ overlay images

Each of these would contain the item to be rendered, color, alpha etc. but this somewhat feels like duplication.

leezer3:
Have separated into two lists, plus abstracted text and images separately.
(Only decent way I could come up with....)

Next jobs:
* Implement marker text fully into the routefile.
* Implement fade in / out for image markers.
* Implement early / late etc.
* ?? XML marker format ??

leezer3:
Commit above adds the command __Track.TextMarker__ , and gets things running in a basic manner.
This works identically to __Track,Marker__ , and displays the text entered in the first command argument.

TODO:
* Allow changing of the color of the text.
* Early/ late text.
* XML marker loader.
* Fade in/ out of images if selected via XML file.

leezer3:
Getting closer....

Text and image markers now work using the original distance based model, including early & late etc.

TODO:
* Allow a message to be removed via time, rather than distance.
* Allow both time and distance to be considered when removing a message.
* Squash down into a single commit.

leezer3:
OK, reasonably ready to squash into a single commit now.

Let's see what Travis does.

leezer3:
Squished and merged!

PR #160, opened 2017-06-21
Move LBAHeader to DevTools folder; Fix makefile; Fix bash script with windows line endings.

Sirflankalot:
Implement the suggestion I made earlier to move tools into DevTools folder. I also fixed dependency issues with the makefile and fixed AssemblyInfo.sh which had `/r/n` line endings.


Sirflankalot:
Forgot to change OpenBVE.csproj for xbuild. Problem is fixed.

PR #158, opened 2017-06-15
Train refactor

leezer3:
### Nov 17- Updating to note what has been manually done (This PR is completely out of date), will strikeout comments as done

Current changes:
* ~~Separate out most structs into separate files (Will eliminate some later)~~
* ~~Make the axle position a property of the axle itself.~~
* Make exterior loading a property of the train.
* Move car motion into a property of the train. (Possibly ought to be the car itself, but each car is a train!)
* ~~Convert several items to a Vector3 rather than 3 doubles.~~
* ~~Move Horns to be a property of the car~~
* ~~Move point, run and flange sounds to the axle~~
* ~~Check encoding of Extensions.cfg file~~
* ~~Move sounds and properties into the doors~~
* ~~Make disposal a property of the train~~
* ~~Run and motor sound updates removed from the lump and into separate functions~~
* ~~Move car relative camera updates to the car~~
* ~~Physics and speed update methods moved to the train~~

Other notes:
* Some of the logic abstraction work has needed to be done slightly differently. See the BVE5 branch https://github.com/leezer3/OpenBVE/pull/154

### Final Remaining:
Final remaining work to close out the changes in this PR on mainline is to manually merge the BrakeSystems code.
Should be relatively quick to do, and parts of this are required for the BVE5 PR.

Also need to seriously investigate creating a SoundManager root class, but that will probably happen well after the BVE5 initial merge.

PR #162, opened 2017-06-26
Plugins: Extend door button interface

leezer3:
This allows plugins to detect when the door button has been pressed, regardless of what the doors then do.


leezer3:
Small TODO:
http://openbve-project.net/documentation/HTML/train_panel2_cfg.html#subjects

The available subjects for doors are as follows:
door
doorl
door
doorl**n**
doorr**n**

What this needs is the ability to show the button press as well as this status, eg.
doorbuttonl
doorbuttonr

Other interesting ideas (Probably requires #158 merging first & XML configuration files):
* Delay before the door opening process begins.
* Control over door opening times. At the minute, this is controlled by the length of the door sound: https://github.com/leezer3/OpenBVE/blob/master/source/OpenBVE/OldCode/Loading.cs#L235
* Standardize door locking beacons?

MarcRiera:
Thanks for this addition, it will be very useful to improve cab animations.

The other ideas sound great too, especially the latter two. The current sound-based door opening/closing timing is rather cumbersome and precise configuration is practically impossible. Enabling some minor configuration shouldn't make things too complicated, it's one of those things that people don't like because they have a fixed behaviour.

As for the door locking beacons, it is probably a good idea too. I have been using a custom system for a route under development and it would be interesting if the indexes were standardized in order to avoid plugin conflicts. Real-life train safety devices are usually not compatible between each other, but in OpenBVE many people love to ride routes with their favourite trains, so interoperability to some extent would surely be attractive.

leezer3:
First a note:
When I say 'standardizing beacons' , I'm thinking along the lines of a set of dedicated routefile commands, so to take the door interlock as an example:
.DoorInterlock 0;1

e.g. Door interlocking starts from this point, on side 1 (right)

I'm well aware that in 'real' life things very rarely work together (Or for that matter in several cases fit at all), but as you note in a simulation context, allowing things of this nature to work wherever is I think good practice.


The other ideas are in the pipeline, but will take a little longer. Dunno how long, but I need to finish the train refactoring first; At the minute, too many things are stuck in the middle of '00 line functions to give the level of control I'd like.

leezer3:
This tests OK locally and would appear to be ready to merge when Travis has built.

Note: Minor behaviour change, in that the door state no longer bounces when the key is held down, but this seems completely logical to me.

Issue #109, opened 2016-10-16
ObjectViewer_fileopen drag and drop


ginga81:
ObjectViewer is only to open file(s), press F7 key and select file.
I want to open file(s) by drag and drop operation.
Please consider this issue.

leezer3:
Not currently included in openTK.

https://github.com/opentk/opentk/pull/542
Tracking this PR, and when merged will revisit.

ginga81:
I didn't know that the drag and drop is by OpenTK.
I thought that it is by .NET.
I am waiting for that applied the drag and drop version of OpenTK.

By this correspondence, the creator can very easy and smoothly checking at the same time the rail, train and platform etc.
Thank you very much for considering this request.

PR #163, opened 2017-07-05
Add Drag and Drop support for Route Viewer and Object Viewer

leezer3:
Closes https://github.com/leezer3/OpenBVE/issues/109

Untested on anything other than Windows at the minute, but will do so later tomorrow.


source/RouteViewer/RendererR.cs
```diff
@@ -934,7 +934,7 @@ private struct ObjectFace {
 			GL.LoadIdentity();
 			GL.MatrixMode(MatrixMode.Projection);
 			GL.PushMatrix();
-			//GL.LoadIdentity();
+			GL.LoadIdentity();
```
leezer3: No idea why this got commented out, or why it worked until openTK's version got bumped, but there we go....

PR #165, opened 2017-07-20
New: Basic route viewer mode for the main program

leezer3:
This is a basic implementation of a viewer mode for the main openBVE program, eventually designed to replace Route Viewer.

**To Use:**
* Set the CurrentOptions.GameMode to developer. (Currently has to be done in-code, no UI option at the minute.

**Flaws:**
* ~~No event markers.~~
* Game is not paused by default. (Need to figure out the exact controls, so we can essentially just run the sim if people wish to test TOD based effects)
* ~~Requires Route Viewer overlays porting. (Should be a relatively quick job.....)~~

### DIFF
--- a/source/OpenBVE/Graphics/Renderer/Overlays.HUD.cs
+++ b/source/OpenBVE/Graphics/Renderer/Overlays.HUD.cs
@@ -47,6 +47,10 @@ namespace OpenBve
 			switch (Command)
 			{
 				case "reverser":
+					if (Interface.CurrentOptions.GameMode == Interface.GameMode.Developer)
+					{
+						return;
+					}
 					if (TrainManager.PlayerTrain.Handles.Reverser.Driver < 0)
 					{
 						sc = MessageColor.Orange;
@@ -86,7 +90,7 @@ namespace OpenBve
 					Element.TransitionState = 0.0;
 					break;
 				case "power":
-					if (TrainManager.PlayerTrain.Handles.SingleHandle)
+					if (TrainManager.PlayerTrain.Handles.SingleHandle || Interface.CurrentOptions.GameMode == Interface.GameMode.Developer)
 					{
 						return;
 					}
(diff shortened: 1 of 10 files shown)


leezer3:
This now mostly works as-per Route Viewer.

__TODO:__
* Disable unrelated controls / overlays when in developer mode.
* Camera motion speed is not right when using WASD, check.
* Implement developer mode into the dropdown bit.
* Allow hiding of the overlays
* Merge the main branch, and then specifically disable all hacks when in developer mode. (Developers should find and fix the issues, not rely on workarounds)
* Station jump is up the creak when using the numpad, check this.

leezer3:
Another PR which has ended up getting out of sync with master....

Have cherry-picked the commit to render the events, and placed that instead behind a new key option (Set to CTRL+SHIFT+E by default)

I think the next step is to cherry-pick the initial loading commit & then rehack the overlays.

leezer3:
Revisited this today, a little more work done.

Flaws:
* Disabling the overlays is crappy code at the minute.
* Pause needs to be reworked, so that time is paused by default, but we can also change the time, mess with dynamic lighting etc.
* No key etc. overlays
* Viewer mode keyconfig?

adfriz:
So... from what im understand the merged route viewer will be just a game mode now?

For me, idk is this the _right_ way...

If this merged, the flow will be: open launcher -> select route -> change mode to developer -> start. right?

---
My thoughts how about we just merge all viewer and editor like traineditor to the main launcher. Now the launcher is the one stop for all player or route/train developer. The access just new button in the side bar.

Screenshot 2026-08-10 180458

leezer3:
Honestly, dunno. This is an old branch, and it's primarily what I'd term a loose idea at the minute, by no means anything fixed in stone.

The basic problem is however that by maintaining a separate game and viewer. we're duplicating code massively (two copies of the renderer, function scripts etc. etc.) and it's very easy to miss fixing something in one or the other.

The link to the developer tools in your picture is basically what's in the GL main menu. That works and is simple, but still has the duplication problem above.

(To note, it *might* be possible to split the function scripts into their own manager DLL, and combine them that way. Haven't thought about that though....)

PR #166, opened 2017-07-21
Create Loksim3D object parser & reorganise slightly

leezer3:
This gets rid of the OldParsers folder, and moves the existing Parsers folder into a slightly better structure.

Other changes: Some fixes to Loksim animation conditions, and an initial merge of the Loksim code into the main sim.

Merge after next marked release.


leezer3:
Object Viewer now has Loksim's Alpha bitmaps implemented.
Probably a little slow (We perform two resizes and the final texture copy run at the minute), but this should only be used on initial texture loads, and the resulting final bitmap retained in memory.

This can just be copied + pasted into Route Viewer, but will probably require a new API method taking two bitmaps for the main sim.

leezer3:
The package manager now handles .l3dpack files.

TODO:
* Search the installation directory of .l3dpack files from viewers etc.
* Create a separate category within the package manager for these. (Currently under 'Other')

leezer3:
Eliminated a completely unnecessary loop....
We also now handle not found files much better: Previously, the rotation attributes could get applied to the next object in line by mistake.....

leezer3:
Mostly there.

TODO:
* Allow setting of the Loksim package install directory (?)
* Save Loksim package directory in filesystem.cfg
* Version filesystem.cfg (Should have done this sooner.....) so that we don't just junk stuff we don't understand when it's re-saved.
* Pass the Loksim package directory as a search parameter; Have added the parameter already, just needs pushing correctly.

leezer3:
OK, so this is now reasonably complete.
Loksim packages are stored under **UserData\ManagedContent\Loksim3D**
This is not currently changeable, except by editing the __FileSystem.cfg__ file.

They are currently displayed under 'Other' in the package manager tab. Still thinking about whether we need an 'object pack' category or similar.

PR #167, opened 2017-07-25
New: Implement basic MSTS cabview parser

leezer3:
This branch implements an incomplete MSTS cabview parser.

Supported:
* Power & brake handles. (Basic config only- Both engine and loco brake levers work together)
* Ammeter (Uses the BVE power related bodge, not 'real')
* Brake pipe/ brake cylinder gauges

TODO:
* Tweak the current parser, so that subject transform and element config form part of the base loader, rather than being duplicated in each file.
* Add appropriate error message if the power/ brake notches in the CVF don't match that in the train.


leezer3:
Oops, makefile not updated, hence why Travis is falling over.

Will fix tomorrow.

leezer3:
Now train plugins are available, this needs to be moved into it's own plugin, and a parser implemented for the consist files and the ENG files.

All of these follow the same basic principle / token block method, so it's do-able with relatively little code, but check our missing features list above?!

leezer3:
Closing this in favour of https://github.com/leezer3/OpenBVE/pull/684

PR #184, opened 2017-08-24
Handle nested groups in the Loksim3D object parser

leezer3:
This handles nested groups in the Loksim3D object parser, e.g.
http://www.veb-verkehr.de/betrieb/down/U-Bahn_Berlin-Obj-D_v201.l3dpack

Mostly complete, but something isn't quite right with the rotation.
Probably isn't being passed down the levels quite right.

loksim01

Note that doors on the RHS (Viewed from the front of the train) are misplaced, and that the interior bits are sticking out.
I *think* that the rotation is probably 180 degrees out, but need to find out whether this has not been applied, or whether this has been passed down incorrectly.....


leezer3:
That's very odd.....

The rotation is definitely broken, but there would appear to be further issues with the doors and associated bits.
Note the part sticking out after the first door (Windfang.l3dgrp)
```XML
<Object>
	<Props Name="dunkel\Windfang.l3dgrp" Position="11.905;0;1.205" Rotation="180;0;0" BlinkIntervall="0" FixedDynamicVisibility="true" DynamicVisibility="" WeitSichtbar="false" ZoomFaktor="1" />
</Object>
```
This consists of two objects. The 180 degree rotation must be applied to both. 
That's fair enough. However, look at the same part next to door number 2: One of the objects has failed to load completely.

Investigate!

leezer3:
That fixes the multiple group rotation, but not the object related bug.

Still need to investigate this one.

leezer3:
bug3

??z-sorting issue??

Just noticed this one, we seem to be getting odd see-through effects at certain angles.
Need to check which transparency type is in use.

Also have various uses of __5__ as the transparency type for these, check what this should be...

leezer3:
Very odd...

The z-sorting issue only occurs when using ObjectViewer.

leezer3:
Ouch...

It appears possible to apply a function to an entire .ls3dgrp file when it is nested.
I'd guess that this overrides any functions within the child file (e.g. if hidden, these will always be hidden), but openBVE simply doesn't work in this way.

Thinking......

leezer3:
No real work in the last commit, just a total refactor of the ObjectManager (Main sim only) to remove various static functions with an insane number of ref calls....

Fixed bugs as a result of this:
* Cloning an object with night-time textures could crash due to an incorrect check

leezer3:
bug4

That's better, the interior is now in place correctly.

The issue with missing components appears to have been to do with the Z-sorting algorithm misbehaving, with the sheer number of bits involved with this model.
For the minute, we now merge all static parts of the object into a single internal MeshBuilder, as opposed to keeping them as distinct groups, which solves our problem.
The image above is still an issue, but as it's only in ObjectViewer, is low priority.

TODO:
* Transparency type 5 currently generates an error message, but works OK. This should be Use the Alpha channel from the image.

Issue #185, opened 2017-09-06
Using large StructureIdx values produces excessive memory allocation


leezer3:
Take the following FreeObj StructureIdx (from a sample provided by email):
```
48 87 94 31
```

This causes the following array to grow to ~4.9 million members:
https://github.com/leezer3/OpenBVE/blob/master/source/OpenBVE/Parsers/Route/BVE/CsvRwRouteParser.cs#L142

Similar issue for all other structure types.
What we therefore require is to replace the plain array with a dictionary / key structure.
This probably has a very slight performance overhead when looking up an object, but this is going to be overweighed by not allocating stupidly sized arrays.

PR #186, opened 2017-09-06
Change: Improve handling of excessively large StructureIdx values

leezer3:
Closes https://github.com/leezer3/OpenBVE/issues/185

Needs some more testing to make sure nothing has broken.

### Specifically:
* ~~Test what happens when an object indices is declared twice.~~ ==> Broken, see below.
* Double-check routes using _.Crack_ and the other deformable object types.
* ~~Do we port to Route Viewer? https://github.com/leezer3/OpenBVE/pull/165 may negate the need for this, but not sure how far away that is.~~

### DIFF
--- a/source/OpenBVE/OpenBve.csproj
+++ b/source/OpenBVE/OpenBve.csproj
@@ -148,6 +148,7 @@
     <Compile Include="Parsers\Route\BVE\CsvRwRouteParser.ApplyRouteData.cs" />
     <Compile Include="Parsers\Route\BVE\CsvRwRouteParser.CompatibilityObjects.cs" />
     <Compile Include="Parsers\Route\BVE\CsvRwRouteParser.Expression.cs" />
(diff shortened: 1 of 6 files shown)


leezer3:
Fiddle, some of the Travis builds are failing because using as an inline out isn't actually supported until C# 7.0, despite the fact that this is just a syntatic little push....

Will fix that, and probably port to Route Viewer later.

leezer3:
Hmm.
Object indicies declared twice now throw an exception (Object containing the same key has already been added)

This is not logged anywhere either, which isn't good.

### TODO:
* Implement our own dictionary class, allowing us to override the Add method (Probably requires the least amount of work ( https://stackoverflow.com/questions/6250706/override-dictionary-add )
* Log the fact that an object index has been used twice- This may increase error counts slightly on some routes, but it's worth it.
* Check BVE2 & BVE4 behaviour on this front.

PR #203, opened 2017-11-02
New: Start adding some unit tests

leezer3:
[ci skip]

WIP for adding unit tests for our API types, just basic addition, subtraction and negation for Vector3 at the minute.

Whilst (as far as I know....) everything works as expected, once completed this will hopefully stop stupid bugs from creeping in.
Specifically, stuff like some of the more custom string parsing functions (to times etc.) would be useful to be tested, as several oddities have been fixed in this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,7 @@ source/*.xml
 
 #InnoSetup Output Folder
 [Oo]utput/
+
+#NuGet for MS Unit Tests
+packages
+!packages/repositories.config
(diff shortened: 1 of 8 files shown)


leezer3:
This highlights the first issue (Minor, but there we go)....

Vector2 and Vector3 have the == operator defined, but not the Equals.

Michelle didn't implement this one, can't see a good reason why at the minute, but the compiler warning was specifically disabled.

leezer3:
Operators added, and cherry-picked onto the main branch for good measure.

Checking with the debugger, these are unused (presumably why not implemented), but using the default CLR comparisons would produce unintended results.

leezer3:
Same issue with Colors regarding operators, will try and deal with those at some point soonish.

PR #207, opened 2017-12-03
New: Allow a sound to be played on state change

leezer3:
This is the initial prototype for allowing sounds to be played on state change.

TODO:
* List of sounds per state?
* Separate out to a .sound file?
* ==> Related- Allow for radius / pitch to be altered? No easy/ neat way to do this in a .animated file, and XML really doesn't work well for sounds...
* Configure whether the sounds are played on show / hide, or just state change?


leezer3:
Animated format:
```
[StateChangeSound]
Filename = crossing.flac
Pitch = 1.0
Volume = 1.0
PlayOnShow = true
PlayOnHide = true
```

Available options:
- **FileName** : The relative path to the sound _or_  **FileNames** : A comma separated list of sound files for each state (starts with -1 ==> hide)
- **Pitch** : Set the pitch
- **Volume** : Set the volume
- **PlayOnShow** : Controls whether a sound is played when this object is shown. 
- **PlayOnHide** : Controls whether a sound is played when this object is hidden.

NOTE: If not using a hide sound, this must be left blank, e.g.
```
FileNames = ,shown.wav, state1.wav etc.
```

Issue #213, opened 2018-01-11
Freeze on exit


City-busz:
When I try to close the OpenBVE's main window with the close button, it just freezes instead of exit from the application. I can close it with Ctrl+C on command line, but anything else don't work. The result is the same when I build it from the sources and when I use the prebuilt archive. I also tested some earlier versions with the same results.

Distribution: Arch Linux

Package versions:
- mono 5.0.0.100-2 and 5.4.1.7-2
- libgdiplus 5.6-1
- openal 1.18.2-1
- openbve 1.5.3.0-1

Locale: LANG=hu_HU.utf8

leezer3:
Going from past experience, the probable fix is going to be to add an Application.DoEvents() call to the method that the button calls.

Will try and roll an Arch VM tomorrow and take a closer look; I'm sure the form closes alright on Debian though :/

leezer3:
Closes OK for me using XFCE, which suggests a desktop / WM issue.

I note the hack I mentioned is already in place:
https://github.com/leezer3/OpenBVE/blob/master/source/OpenBVE/UserInterface/formMain.cs#L1353

City-busz:
Thanks for testing this.

I experience the same problem in desktop environments where I tested: GNOME Flashback (Metacity), MATE (Marco), LXDE (Openbox).

Some details:
- When I click on the close button on the window decorator, the form freezes.
- When I click on the "Close” button on the bottom left corner of the form, the window disappears, but the 'openbve' process still running in the background.
- When I start a route and try to exit from the simulation, the window closes, but the 'openbve' process still running in the background.

leezer3:
Interesting.

I can reproduce using the button in the top-left. Let me do some digging.

leezer3:
Try the branch here:
https://github.com/leezer3/OpenBVE/pull/214

This explicitly issues an Environment.Exit(0) call after any close request. This appears to get rid of the hanging around issue with no effects I can see on the Arch VM.

I don't think this should do any harm, as all de-initialisation should have been done by this point, but needs more testing on other combos before I merge into mainline.
I think it's really a Mono bug mind, but I don't want to try and chase whatever is locking.

City-busz:
I tried the patch, and I'm able to exit the form, but when I try to start a game, the program exit instead of starting it.

leezer3:
Ah, oops....

Pushed another commit, which I think clears things up, although I've had to hack around the closing logic quite a bit, which I'm  not happy with at all.

[...]

PR #214, opened 2018-01-12
Change: Attempt to force close the window under Mono

leezer3:
https://github.com/leezer3/OpenBVE/issues/213

Some systems seem to keep things hanging around after the game has terminated. Only seen this under Mono, and it appears to be a little random, so probably WM / DE related.

Therefore:
If we are running under Mono, specifically issue an **Environment.Exit(0);** call after the program should have successfully quit.


leezer3:
Closed by #287

PR #219, opened 2018-02-10
MeshVertexColors template in X files

leezer3:
Support for the MeshVertexColors template in .X files.
Also required (well at least declared in) by #242 

Sample from a textual object:
```
MeshVertexColors { 
4; 
0;0.000000; 0.000000; 1.000000; 0.000000;;, 
1;0.000000; 1.000000; 1.000000; 0.000000;;,
2;0.000000; 0.000000; 1.000000; 0.000000;;, 
3;0.000000; 1.000000; 1.000000; 0.000000;;; 
} 
```

### Specification:
* Array count
* Vertex index ==> R,G,B,A [Repeated until count]

### Sample object (zipped):
[rail0.zip](https://github.com/leezer3/OpenBVE/files/1713479/rail0.zip)

### TODO:
* Test Loksim parser, all programs.
* Test Wavefront parser, all programs.
* Test object based backgrounds, main program.
* Compare memory usage on large NYCTA route (Should be OK now the types have been abstracted, but test)

[ci skip]


leezer3:
Binary files fixed.

Need to check what this does to memory usage on large routes in the main program next:
I believe this will add a total of 20 bytes per vertex (16 bytes for storage of the color, plus 4 bytes for the nullable boolean when aligned by the compiler), which on some of the NYCTA routes may be an issue.
Thus:
* Add a VertexColors array to the face somehow, and hook into this when rendering?
* Check how face colors are handled; Can we hook into this? (IIRC they just issue a single GL.Color call before rendering, which may not be of too much use when trying to color single verts???)

leezer3:
https://github.com/leezer3/OpenBVE/commit/e11b7f02da1f911f573c9b1a36d0170956923b6b

Some tangentially related improvements to the parser (Makes models converted to X using Open 3D Model Viewer work)

leezer3:
First stab at abstracting vertices in the commit above.
It also removes the (unused & unfinished) object plugin interface.

Issues:
* Moving vertices to a class rather than a struct means we have to instance them when we create them. Needs thoroughly testing to ensure nothing has broken.
* ??Rename VertexTemplate to something better??
* Animations seem to have broken as a result of this. Can't see why initially, probably instancing where we should not, but needs finding and testing....

leezer3:
OK, that now seems to work OK in Object Viewer.

Next needs porting to Route Viewer and testing in a bigger environment before finally the main sim.

leezer3:
AppVeyor is throwing a wobbly because this PR was instituted before it got added :/

Need to FFWD the branch, but not doing that just at the minute. Presumably that'll be an issue with any of the older PRs.
Irritating that it causes a failure badge for the main branch too......

leezer3:
FFWD branch relative to master and port to Route Viewer.

### TODO:
* Test Loksim and Wavefront parsers
* Add vertex colors to Route Viewer
* Port to main program

leezer3:
Squished and FFWD to the current base.
source/ObjectViewer/RendererS.cs.orig needs removing before merge.

leezer3:
Try again, fixed the main program and removed the junk file.....

PR #222, opened 2018-02-20
Destination event

leezer3:
This PR implements destinations for trains.

### Routefile Code
Train.Destination(__InitialDestination__) 
__InitialDestination__ should be an integer, representing the initial destination displayed by the train at the first stop

---

Track.Destination(__Type__ , __BeaconStructureIndex__ , __NextDestination__ , __PreviousDestination__ , __TriggerOnce__ , __X__ , __Y__ , __Yaw__ , __Pitch__ , __Roll__)
__Type__ should be one of the following: _-1_ : AI trains only _0_ : All trains _1_ : Player train only
__BeaconStructureIndex__ should be the index of any beacon object previously declared, or -1 to show no visible marker.
__NextDestination__ represents the destination to change to when passing over this beacon forwards, or _-1_ to not change.
__PreviousDestination__ represents the destination to change to when passing over this beacon in reverse, or _-1_ to not change.
__X, Y , Z, Pitch , Roll__ control the positioning of the associated object (if in use)

### DIFF
--- a/documentation/HTML/route_csv.html
+++ b/documentation/HTML/route_csv.html
@@ -2136,6 +2136,27 @@ This command places an environmental sound effect at the specified location. The
 
 </font></td></tr></table>
 This command places a bumper. The train can collide with the bumper in both the forward and backward directions. Place this command at the beginning and the end of the route. An object is not automatically created, so use Track.FreeObj to create a visual representation of the bumper if necessary.
+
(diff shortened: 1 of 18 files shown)


leezer3:
Today's work adds the variable Destination to the ElapseData supplied to plugins.
As per the handles, this may be modified, and the value will be passed back to the sim when the plugin returns.

PR #230, opened 2018-03-13
Plugin proxy to allow for x64 builds

leezer3:
This represents a rather major step forwards, allowing us to proxy 32-bit plugins & communicate with them via WCF, allowing 64 bit builds.

Major changes:
- New app: WinPluginProxy. This proxies the plugins and communicates with the main thread via WCF.
- Most runtime attributes set to DataContract / DataMembers.
- New runtime class: ElapseProxy ; This holds the ElapseData and the Panel / Sound integer arrays needed to callback (Change or drop into the interop class in the main program, not runtime???)

TODO:
- Move the Interop into the API (???)
- Tidy things up
- Either set the current setup to be used when on Win32, or remove entirely?
- Can we ask WINE to run the plugin proxy on Linux / Mac?
- Fix makefile to build the new bits
- Pass errors back the other way (Do we need to takedown the plugin as per now, or just log these?)

@Sirflankalot any thoughts?


Sirflankalot:
Holy crap, that is really impressive. I wonder what caused that. It may be better code generation (x64 has SSE2 by default while x86 is technically i386 which is _old_) or could me memory management things. Either way this is a massive step in the right direction!

leezer3:
Vsync turned on for the first, I knew there was something I missed....

leezer3:
Much less laggy though, I can run through Laira without a hitch now at 5x speed.

Something still not quite right, as EB applications from the plugin aren't working.
Hopefully the last commit gets the makefile right as well :/

leezer3:
Arrghhh.....
Makefile works, but something in the build order is funny when mono builds from the sln under Linux, meaning that the dependancy on openBVE.exe is never resolved properly.

Going to work now, will look at this properly tomorrow.

Sirflankalot:
That sucks that it is different, but all improvements are good! Are you using xbuild or msbuild? I have never gotten msbuild to work on my machines.

leezer3:
msbuild

IIRC Mono defaults to msbuild now on modern versions as part of the .Net OSS stuff.

leezer3:
One day I'll murder Windows and the fact that it's not case-sensitive in it's filesystem handling.....

Build now works correctly- I'm now going to play with WINE.

leezer3:
Pity, WINE isn't too happy :(

- Attempting to load directly using a Process.Start() freezes before the server even loads.
- Loading the server directly from the console triggers, but it doesn't take incoming connections & everything just freezes.
- Installing .Net under WINE also freezes, and starts complaining about pointers and stuff ad infinitum. If I'm guessing, it doesn't like the fact that we need to pass specific size structs to the DLLimport.

leezer3:
Interface declarations moved into the API.
This removes the need for the reference to the main program for WinPluginProxy.

Have also implemented basic callback handling from the plugin so that errors are logged.

Current flaws:

- Main program crashes if the plugin server window is closed. Getting a reference to a running service is a PITA, so raising an event isn't really working. Maybe just add try / catch blocks to the plugin in the main program.....
- Plugin server window steals focus. Probably just needs the background attribute setting, possibly hide entirely.....

leezer3:
Redone to work with the current codebase.

Becoming a little more urgent with the Catalina situation ( https://github.com/leezer3/OpenBVE/issues/384 ) and also may reduce pressure elsewhere too with some of the insanely coded stuff....

Unsure if this will be merged at the minute, various issues to work out:
* The RailDriver needs a completely separate DLL import for x32 and x64.....
* Whilst the situation after closing the server window has been improved, the controls are jammed.
* x32 .Net plugins (shouldn't really exist, but I'm sure someone has) could do with running through the proxy.

leezer3:
Now that's interesting.

Latest Mono msbuild seems to handle casing errors in the sln file much better :/
Will fix that tomorrow and see about getting the RailDriver working for both if possible (Will be a PITA!)....

leezer3:
A bunch of rather stupid duplication of code (why on earth PIE can't just detect the IntPtr size internally I really don't know...) but we now have the RailDriver working with x64 builds.

Marginally considering declaring x32 only .Net plugins obsolete.
Unaware of any that actually exist, AFAIK the only three released .Net plugins are all AnyCPU.

leezer3:
Interesting thought-
Corflags will let us mark an AnyCPU executable as 32-bit.
```corflags.exe myapp.exe /32BIT+```

Assuming (!!!) this works, it'd be trivial to automagically provide both 32-bit default and 64-bit with proxy in the builds.

Not going to make it into the current release, but seriously thinking about this, just need to find any more interesting little kinks....

leezer3:
It even builds without blowing up!

Still wants some of the JoystickManager stuff moving about, but somewhat closer to mergable.
Need to test the x64 Winforms under Mono too.....

PR #232, opened 2018-03-22
Appveyor-Ci for Windows Continuous Integration

Sirflankalot:
I've been using appveyor for my windows ci on my project and it is by far the best implementation of CI I've seen thus far. I implemented a basic check for all combinations, setting up automatic generation of installers and such is something that either you have to deal with, or I have to deal with when I'm back at home and know how to call Inno Setup on the command line. They have a similar thing to travis-ci so that you can store your ssh key privately for uploads.

All you would have to do is sign into appveyor, click add project, and add openbve to your builds. Proof that it works... [here](https://ci.appveyor.com/project/Sirflankalot/openbve/build/1.5.3.2-7)


Sirflankalot:
Errored on one mac instance, but shouldn't be an actual problem, all others passed

leezer3:
IIRC Innosetup can be called just by passing the iss file as a parameter to the iss.exe file along with the /q switch.

Really needs a basic cmd or powershell script knocking up to generate the version numbering & upload the final result; I'll see if I can take a look at that tomorrow.

Sirflankalot:
Alright, sounds good. In the mean time, I will add a bit to the appveyor script to see if I can at least build the installer. On appveyor, Inno Setup is at `C:\Program Files (x86)\Inno Setup 5`.

Sirflankalot:
Alright, I just confirmed you can run inno setup fine, only took me 4 tries. [This](https://github.com/Sirflankalot/OpenBVE/blob/6075b13a720b2c4c55161e8cee079b8012633020/appveyor.yml#L28) is the important line.

leezer3:
Took a few attempts, but I *think* AppVeyour should now generate the installer version correctly.

Something I've noticed:
```
The project "OpenBve" is not selected for building in solution configuration "Debug|x86".
The project "ObjectViewer" is not selected for building in solution configuration "Debug|x86".
The project "ObjectBender" is not selected for building in solution configuration "Debug|x86".
The project "LBAHeader" is not selected for building in solution configuration "Debug|x86".
```

Should hopefully be fixed with the last commit.

I'll leave this for a bit before merging and enabling AppVeyor on the main repo in case anything else comes up.

Will have to look into the deployment too, looks like it should support SFTP.

Sirflankalot:
It seems like the debug builds on non-2017 are now failing due to a missing [AssemblyInfo.cs](https://ci.appveyor.com/project/Sirflankalot/openbve/build/1.5.3.2-17/job/a4ilddpgil4wdna4#L255). See [here](https://ci.appveyor.com/project/Sirflankalot/openbve/build/1.5.3.2-17). I don't know particularly what's gone wrong, not on windows right now.

leezer3:
I think it's looking in the wrong place for TextTransform.exe; The vs2013 build I checked was trying the 14.0 folder, which IIRC is 2015

Reason its broke though is because it wasn't building the main program in the first place & now it is :/
Will take a proper look with a 2013 CM tomorrow

Sirflankalot:
Ah nice....

Question, how does the change you make force it to build it?

leezer3:
It's the projects selected to build for each CPU config.
Everything was only selected to build for AnyCPU (Incidentally, I think we need to keep the API as AnyCPU to maintain backwards compatbility.

Solved the texttransform issues by shipping the Mono implementation instead, it's 100kb & MIT licenced, and much easier than fighting trying to find it with VS2010 / 2013 (Not sure, but looks to me like it's actually in the modelling SDK not the minimal install)

Issue #234, opened 2018-03-23
More of a feature request


wizdagamer:
Not so sure how to request a feature but i have some I would like to share. If you can, move it to the correct spot please. 

- For trains that have the Lap position brake (AMUE is what we call them in NY) add a initial release. This would then release the brake pressure slowly. 

- Add a option for cab sway...Sorta like in WoS V4. I noticed in some videos that they crack up the swaying of the train. How would I go about that till you maybe add a slider? 

- Another thing I would like is the Continuous Range Motion. The new technology trains in NYC have this, and it would be especially cool if you have a throttle controller and can simulate it. 

- The ability to put the throttle into deadman with the brake in emergency with Panel2.cfg (Don’t know how to do it if you already can. Separate handles btw)

I believe that is is. If I have anymore I’ll be sure to let you know! Thanks for listening.

leezer3:
```
Add a option for cab sway...Sorta like in WoS V4. I noticed in some videos that they crack up the swaying of the train. How would I go about that till you maybe add a slider?
```
Have you looked into the __Track.Accuracy__ command?
This accepts floating-point values from 0 (No cab sway) ==> 4 (Max cab sway)


Will have a think about the other stuff and see what I can come up with.

leezer3:
```
The ability to put the throttle into deadman with the brake in emergency with Panel2.cfg (Don’t know how to do it if you already can. Separate handles btw)
```
Not currently possible.
However, it's easy enough to do:
https://github.com/leezer3/OpenBVE/pull/236

Will merge this into the main branch in a day or two once I've thought about things a little more and checked there's nothing I've forgotten.

PR #236, opened 2018-03-28
New: Option to control the other handles when EB is applied in Train.dat

leezer3:
https://github.com/leezer3/OpenBVE/issues/234

This adds an option to control the behaviour of the other handles when EB is applied
Available options:
* NoAction
* PowerToNeutral
* ReverserToNeutral
* PowerReverserToNeutral

Whilst I don't always like altering opaque formats like Train.dat, this is a simple change, and is backwards compatible.
(Prior versions of openBVE will silently ignore this. Unsure what BVE4 will do, but it rejects **OPENBVE** format specified train.dat files anyway)

### Potential changes:
* ~~Do we need a newer format specifier (e.g. **OPENBVE1500** or something)- Doing this, and a comparision of the string against the 'current' format specifier would then allow us to detect the fact that we're using a newer train.dat than supported and prompt to upgrade?~~ DONE.


OpenBveTrainSim:
Chris:
Recommend a serious detailed study of typical\standard real world train braking systems, 
ie those on actual DMUs, EMUs, Metro, Locomotive stock and Passenger+Goods Wagons,
both historic and modern systems, so they may be more realisticly simulated in OpenBVE.
Likely speaking here of things you already know well.

Most common air brake system manufacturers:
Knorr, WABCo, Oerlikon, Dako, Mitsubishi, IRAB\India, Schaltbau, Secheron, Gessman, Xian\China...
And the legacy vacuum brake system manufacturers:
Gresham & Craven, IRAB\India...

My 2 pence worth.
OBTS

leezer3:
Sure, I'm well aware of most of the IRL systems.
Unfortunately, simulation of most of these isn't something that's likely to come quickly. The WIP XML train format allows control of some of the fixed parameters within the system, but that's not anywhere close to being ready for general public consumption.

This however is a simple change within the existing framework to allow a small piece of logical & requested behaviour :)

OpenBveTrainSim:
Chris:
Umm.  Then OpenBVE train developers desiring to closely simulate real brakes,
have only limited .Animated logic or cumbersome custom .DLL\ATS functions.
Maybe some day.  Thanx
OBTS

leezer3:
Added an openBVE version parameter to the train.dat file.

This will generate a small (harmless) error if used with prior versions of openBVE, which will then silently ignore any extra parameters.

cwfitzgerald:
Before this gets pushed could there be an updated to the docs regarding the new sections? I can do it if need be.

leezer3:
Its documented in the included docs copy, but not the site, which I'll push when this is ready. Don't want to put a draft feature into the public docs :)
Need to add the train.dat versioning too, but thats minir and doesn't really do anything useful ATM.

cwfitzgerald:
Wonderful! Thanks!

Of course :)

leezer3:
Merged and main documentation updated.

Not really a new section as such, just one additional parameter in the **#HANDLE** section, and an (optional) addition to the **OPENBVE** identifier in case something else like this gets added.

PR #240, opened 2018-04-21
New: Add DDS texture loader plugin

leezer3:
This adds a DDS texture loader plugin.

The code used is a C# conversion of the Nvidia reference implementation, with a few mild cleanups & tweaking to return the openBVE texture type.

Licence:
```
Copyright (C) 1999, 2000 NVIDIA Corporation
This file is provided without support, instruction, or implied warranty of any
kind.  NVIDIA makes no guarantee of its fitness for a particular purpose and is
not liable under any circumstances for any damages or loss whatsoever arising
from the use or inability to use this file or items derived from it.
```

This reference implementation crops up all over the place (including in some GPL stuff), so I think assuming this header remains we're reasonable safe to use.


leezer3:
Missing a reference in the makefile, will deal with that one tomorrow.....

PR #256, opened 2018-08-03
Fix: Issue that ConstantSpeed Device doesn't work

s520:
This PR fixes problems that ConstantSpeed device doesn't work, which occurs after 1.5.2.3.


leezer3:
Ouch, you're right- Thanks :)

PR #259, opened 2018-08-06
Loading refactor

leezer3:
This PR is a WIP refactor of the main loading classes.

### Aims:
* Use openGL based message boxes derived from the current in-game menu class- These work much better on non-Windows systems.
* Separate route / train loading types so that all are independant.


leezer3:
Outdated

PR #262, opened 2018-08-19
Fix: Issue that plugin doesn't be unloaded when OpenBVEGame window is closed

s520:
This PR fixes problems that plugin doesn't be unloaded when OpenBVEGame window is closed.


leezer3:
Thanks, that seems reasonable enough.

Need to keep a minor eye on plugins freezing on unload (and thus blocking close of the window). but I think that's niche enough not to bother adding a timeout for unless we actually run into said issue.

PR #266, opened 2018-08-28
New: Add that show the in-game remain distance of the next station

s520:
This PR add that show the in-game remain distance of the next station.


leezer3:
Thanks.

Might update the English translation a bit later, but it's OK for the minute as-is.

PR #268, opened 2018-09-01
Fix: Build error and reference error

s520:
This PR fix build error and reference error.
- build error due to VS2013 (also Mono 3.2.8) not supporting get-only property
- reference error due to a typo in HintPath on monodevelop


leezer3:
Thanks!

The Mono 3.2.8 error was accidental-
Only Debian OldStable (Jessie) is now stuck on 3.2.8, and I've been testing bumping this, and that managed to creep into the main build as opposed to the branch.

PR #269, opened 2018-09-08
New: Add function that something gets caught in the door

s520:
This PR add function that something gets caught in the door.
It works as shown in the picture below.
stuck-door-ichinoseki
This screenshot was taken by Midnight Express Ginga81(ginga81).

### DIFF
--- a/source/OpenBVE/Parsers/Route/BVE/CsvRwRouteParser.cs
+++ b/source/OpenBVE/Parsers/Route/BVE/CsvRwRouteParser.cs
@@ -3968,6 +3968,50 @@ namespace OpenBve {
 												}
 											}
 										}
+										double reopenDoor = 0.0;
+										if (!PreviewOnly)
+										{
+											if (Arguments.Length >= 13 && Arguments[12].Length > 0 && !NumberFormats.TryParseDoubleVb6(Arguments[12], out reopenDoor)) {
+												Interface.AddMessage(Interface.MessageType.Error, false, "ReopenDoor is invalid in Track.Sta at line " + Expressions[j].Line.ToString(Culture) + ", column " + Expressions[j].Column.ToString(Culture) + " in file " + Expressions[j].File);
+												reopenDoor = 0.0;
+											} else if (reopenDoor < 0.0) {
+												Interface.AddMessage(Interface.MessageType.Error, false, "ReopenDoor is expected to be non-negative in Track.Sta at line " + Expressions[j].Line.ToString(Culture) + ", column " + Expressions[j].Column.ToString(Culture) + " in file " + Expressions[j].File);
+												reopenDoor = 0.0;
+											}
+										}
+										int reopenStationLimit = 5;
+										if(!PreviewOnly)
+										{
+											if (Arguments.Length >= 14 && Arguments[13].Length > 0 && !NumberFormats.TryParseIntVb6(Arguments[13], out reopenStationLimit)) {
+												Interface.AddMessage(Interface.MessageType.Error, false, "ReopenStationLimit is invalid in Track.Sta at line " + Expressions[j].Line.ToString(Culture) + ", column " + Expressions[j].Column.ToString(Culture) + " in file " + Expressions[j].File);
+												reopenStationLimit = 5;
+											} else if (reopenStationLimit < 0) {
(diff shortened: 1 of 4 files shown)


leezer3:
Couple of thoughts:
If you want to implement this on a per-station basis, then it might be better in the XML station file: https://openbve-project.net/documentation/HTML/route_stationxml.html
Adding parameters to the end of existing functions is a messy business, and I can see this hooking some incorrectly entered timetable indicies as the probability :)

Secondly, is some thoughts on improvement / complication on how this currently works.

In 'real' practice, the door closing / interlock process will vary considerably dependant on the stock. 
Older stuff will tolerate a much greater incursion into the door and still give traction interlock.
Now, in order to simulate this, we need the following:
* The proportionate chance of an object in the doors- You've implemented this already.
* It then wants a random roll to determine the size of the object in question. (Somewhere from 10mm [bag strap?] upto 300mm [person etc.])
* Compare this to the max tolerance for the stock.

You'd also want to set the door width in train.dat, which would allow you to get rid of the ratio of obstruction :)

I might be over-thinking this though- Thoughts appreciated!

s520:
Thank you for reviewing my code. This code has functions as shown below.

- Determine whether door-catching occurs in each car.
- Avoid re-opening being repeated for longtime by setting the upper limit of the re-opening per station.
- Able to set the maximum value of obstacles per statin.
- Able to set the minimum value of obstackes that the door can detect.

And, parameter that relate to re-opening is added to the end of command:Track.Sta, we choosed this way. Midnight Express Ginga81(ginga81), F81_tec200 and I discussed and decided it.

Of course, we know that there is a possibility that unexpected problem occurs. But there may be a risk of losing compatibility with older version of openBVE and route format by handling only XML station file, we strongly concern about it.

This comment was translated by F81 tec200. I really appreciate it.

leezer3:
Loosing compatibility with older versions is always a danger when introducing a new feature, but potentially causing issues with already existing stuff concerns me much more on that front.

Setting the door width and tolerance settings via train.dat that way is exactly what I had in mind :)

I can't find anything which breaks with this change made, so that's good. I'll do a little more testing before merging this.

s520:
Thank you very much for praise about my code.
I added my new parameter also to the XML Station Format. Please check it.

I suggest that you should add Track.StationXML command's reference to the Developer's Manual.
And I think that at the example of The XML Station Format, not necessary '<openBVE>' token at line 3.

[Additionally, we are attempt to enable multi-language at TrainEditor.exe.](https://github.com/s520/OpenBVE/commit/25d69e65d6bf8003fd126852f0975f842c63ba50)
traineditor_multilanguage
At the same timing, I think that will be planning to enable to edit these parameters.

This comment was translated by Midnight Express Ginga81(ginga81). I really appreciate it.

leezer3:
> And I think that at the example of The XML Station Format, not necessary '<openBVE>' token at line 3.

That's very deliberate, and is implementing XML best practices.
A large part of the decision to move to XML for major new additions is the fact that it's an existing format with a well-documented structure and list of best practices.

I've spent an awful lot of time fixing issues that stem from the original BVE formats, and the fact that they can be interpreted in many different and inconsistant ways by BVE2, BVE4 and openBVE. Implementing a home-grown format was discarded as an idea, as that just leaves us open to doing exactly the same thing.

> I suggest that you should add Track.StationXML command's reference to the Developer's Manual.
That should be there already.
Documentation\route_stationxml.html
Should be in the menu as the very last item under the CSV route.


Multi-language in the Train Editor would be very helpful too.
It'd probably be better though to move the language loading to an entirely separate DLL file though, so we can just import / maintain the one set of code.

source/OpenBVE/Parsers/Script/StationXMLParser.cs
```diff
@@ -331,6 +331,70 @@ public static Game.Station ReadStationXML(string fileName, bool PreviewOnly, Tex
...
+										{
+											if (interferenceInDoor < 0.0)
+											{
+												Interface.AddMessage(Interface.MessageType.Error, false, "InterferenceInDoor must be non-negative in XML file " + fileName);
+												interferenceInDoor = 0.0;
+											}
+											station.InterferenceInDoor = interferenceInDoor;
+										}
+										break;
+									case "maxinterferencingobjectrate":
```
leezer3: Spelling needs correcting- should be **maxinterferingobjectrate**

PR #270, opened 2018-09-09
New: Add XML sound parser

leezer3:
This is a WIP implementation for this idea:
http://bveworldwide.forumotion.com/t1598-station-announcements#17075

Working:
* Adding sounds via the XML
* Positioning sounds
* Pause between repetitions when sound is looped

TODO:
* Add some time based conditions- Rehack the sound event on the track to add time based triggers.
* ~~How easy is it to add a pause between repetitions?~~ DONE
* Cleanup
* Document

### DIFF
--- a/source/OpenBVE/Audio/Sounds.Update.cs
+++ b/source/OpenBVE/Audio/Sounds.Update.cs
@@ -50,7 +50,7 @@ namespace OpenBve
 			 * Update the sound sources
 			 * */
 			int actuallyPlaying = 0;
-			for (int i = 0; i < SourceCount; i++) {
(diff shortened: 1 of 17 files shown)


leezer3:
Pause added between repetitions. 
Note that the user must ensure that their sound ends at amplitude 0 (e.g. totally silent) , as otherwise this will cause a slight pop.
This is **not** a bug.

leezer3:
Somewhat rebuilt to work with the current master.

Following on from discussion over email, a potentially useful addition would be the ability to queue an announcement for playback via keypress.
This adds some of the framework for this (don't like the use of dynamic, but at the same time don't add massive fields to API stuff)

PR #274, opened 2018-09-19
Move Textures completely to API

leezer3:
### Benefits:
* Further reduce duplicated code between Object Viewer and the main program.
* Primary requirement for moving the object loader plugins into a single external DLL, rather than embedded into the main sim.


leezer3:
Supporting old Mono versions strikes again.....

Still, that now appears to compile on all the Travis workers.
Need to check the GetHashCode warnings and see if there's an easy fix, but Michelle had just disabled them, so we shouldn't need to worry too much.

leezer3:
Route Viewer never got the texture plugin interface ported to it, so it's currently left out in the cold....
We can probably use at least some of the API code here (Enums, TextureOrigin), but really need to get as far as implementing a viewer mode in the main program so it can be dropped.

leezer3:
Final build failure is unrelated.

leezer3:
Chucked out the two duplicate enums from Route Viewer.
More spam change than I'd like due to the property names being different (arrggh.....), but will make things much easier to maintain in the long run.

leezer3:
Couple more minor bits of duplicate / junk code

leezer3:
Still testing this, although everything looks good- Can set unloading of textures and bounce around some quite major routes (Plymouth, First Brno Track, NWM) with no problems I've found yet.

PR #275, opened 2018-09-21
 New: Add the multi-language support of TrainEditor

s520:
This PR add the multi-language support of TrainEditor.
In addition, move the language loading to a separate DLL file (OpenBveTranslate.dll).

I think I can integrate the OpenBveTranslate.dll to OpenBveApi.dll. (Example: namespace OpenBveApi.Interface)
And I think that I should move the setting loading to OpenBveApi.dll.


leezer3:
My initial thought would be that the commands might be better in the API, but the translation in the separate DLL still?

The API is designed to be consumed by plugins, which don't necessarily need translation functions.
It'd probably work in that way though.

Doing it that way also helps to keep things separated out, and not in the single lumps we've got at the minute.
For that matter, Object Viewer and Route Viewer could probably do with translations too, and that's only a few strings :)

When object loading gets moved to a plugin (This work in support of that) https://github.com/leezer3/OpenBVE/pull/274 ), some of the settings such as hacks will need to be available or at least set via the API, but I'm not sure we need the entire settings interface available.
Wouldn't hurt to have it mind!

s520:
Thank you for reviewing my code.
I integrate OpenBveTranslate.dll to OpenBveApi.dll. Please check it.
Additionally, I'm attempt to enable multi-language at ObjectViewer and RouteViewer.
I think that it will be possible to lower the threshold for new users in non English-speaking countries.

s520:
I made it possible to save the language setting of TrainEditor. Please check it.

s520:
I fix some issue.
- Issue where Options.LoadOptions () was called before FileSystem.FromCommandLineArgs (args) was called on RouteViewer
- Issue that OpenBveApi.FileSystem use GetExecutingAssembly()
Please review my code.

source/ObjectViewer/Options.cs
```diff
@@ -10,14 +10,25 @@ internal class Options
...
         {
+            string optionsFolder = OpenBveApi.Path.CombineDirectory(Program.FileSystem.SettingsFolder, "1.5.0");
+            if (!System.IO.Directory.Exists(optionsFolder))
+            {
+                System.IO.Directory.CreateDirectory(optionsFolder);
+            }
             CultureInfo Culture = CultureInfo.InvariantCulture;
-            string assemblyFolder = System.IO.Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
-            string configFile = OpenBveApi.Path.CombineFile(OpenBveApi.Path.CombineDirectory(OpenBveApi.Path.CombineDirectory(assemblyFolder, "UserData"), "Settings"), "options_ov.cfg");
+            string configFile = OpenBveApi.Path.CombineFile(Program.FileSystem.SettingsFolder, "1.5.0/options_ov.cfg");
```
leezer3: Marginally better to use the optionsFolder variable declared above in case this changes in the future.

source/ObjectViewer/Options.cs
```diff
@@ -128,43 +139,36 @@ internal static void LoadOptions()
...
-            {
-                MessageBox.Show("An error occured whilst saving the options to disk." + System.Environment.NewLine +
-                                "Please check you have write permission.");
+                Builder.AppendLine("interpolation = " + t);
             }
+            Builder.AppendLine("anisotropicfilteringlevel = " + Interface.CurrentOptions.AnisotropicFilteringLevel.ToString(Culture));
+            Builder.AppendLine("antialiasinglevel = " + Interface.CurrentOptions.AntialiasingLevel.ToString(Culture));
+            Builder.AppendLine("transparencyMode = " + ((int)Interface.CurrentOptions.TransparencyMode).ToString(Culture));
+            string configFile = OpenBveApi.Path.CombineFile(Program.FileSystem.SettingsFolder, "1.5.0/options_ov.cfg");
+            System.IO.File.WriteAllText(configFile, Builder.ToString(), new System.Text.UTF8Encoding(true));
```
leezer3: Crashes if no write access to the options folder (seen this one before.....)

leezer3:
Couple of minor tweaks that stand out at first.

Other than that, I'm not sure that **OpenBveApi.Interface.Interface ** is the best way to name this.
Using the second part of the namespace as the class name is asking for issues further down the line.
I think either change it so that the options and language are in separate classes,
 e.g:
OpenBveApi.Interface.Options & OpenBveApi.Interface.Translations , or possibly just loose the second part of the namespace?

Thoughts welcome!

s520:
Thank you for reviewing my code.
I fixed two minor tweaks. Please check it.
Next, since the Interface class of OpenBVE.exe conflicts with the name, I also recognize that the name OpenBveApi.Interface.Interface is not so good.
So, I would like to name it OpenBveApi.Interface.Translations.

PR #276, opened 2018-09-22
Move various World types to the API

leezer3:
This PR moves the World.Mesh etc. types to the API.

Follow-up to https://github.com/leezer3/OpenBVE/pull/274

Once merged, this is most of the way towards getting the object loaders into a common plugin.


leezer3:
And now we're heading hard towards the perpetual problem- We need another bit of work doing before we can move the animated object types to the API....

Namely, note that FunctionScripts are required, but these need access to the train based variables.
Potentially merge at 247db2c and work out the massive amounts of function script hacking in a separate PR?
Function scripts could seriously do with being in a separat DLL too.

leezer3:
Cross-ported all changes to Route Viewer. Still needs a serious test and bugcheck, but there we go.....

leezer3:
If this builds correctly, this may now be close to mergable status, but testing still needs serious work, as there are a lot of bits that could go wrong here.

We may have some minor speedups, but nothing really interesting. Probably set back by having to pass out a reference in each object, and the cross assembly function calls....

Once this is merged, we need to then work towards using a shared mesh library, whereby each mesh is stored in a display list, which is then called and translated by the GPU during the draw call, rather than the current much more inefficient method of moving the mesh within the world and drawing it directly.

leezer3:
Chucked out yet more duplicate code....

### BUGS:
* Route Viewer's transparency processing appears to be not working correctly. Z-sorting is not right, and using Sharp transparency causes some glitches.

leezer3:
Function scripts might be easier that I thought.

Let's see if the makefile will build these?

leezer3:
Okies, that builds. (Ignore the PR build cancellation, this is nowhere near ready to merge)

Next job is to consolidate and move the animated object classes into the API.

leezer3:
Squashed.

Animated objects are going to be a massive PITA:
We need to access specific functions in the host. Still thinking about that one.

Maybe merge as-is and think again???

leezer3:
Some more work done.

Still not absolutely happy here, but we can now execute a function script via the host interface. This eliminates the abstract function script experiment, and probably works somewhat better.

End-game would be to have the entire animated object inside the API, and executing anything it wants via callbacks?

Still not mergable, but working on it...

leezer3:
Semi-abstracted the animated object.
Here, moving some of the static functions into the object properties may work against us.

Need to do some more thinking....

leezer3:
Did I ever mention I hate makefiles.....

Should now be working again though.

source/Renderer/Renderer.LoadingScreen.cs
```diff
@@ -127,23 +126,23 @@ internal static void DrawLoadingScreen()
...
+				logoBottom	= (int)ScreenHeight / 2;
 			// take the height remaining below the logo and divide in 3 horiz. parts
-			int	blankHeight	= (Screen.Height - logoBottom) / 3;
+			int	blankHeight	= (int)(ScreenHeight - logoBottom) / 3;
 
 			// VERSION NUMBER
 			// place the version above the first division
 			int	versionTop	= logoBottom + blankHeight - fontHeight;
-			LibRender.Renderer.DrawString(Fonts.SmallFont, "Version " + typeof(Renderer).Assembly.GetName().Version,
+			DrawString(Fonts.SmallFont, "Version " + typeof(Renderer).Assembly.GetName().Version,
```
leezer3: Need to either pass in the version to render or set the assembly version for the renderer too

leezer3:
Emissive faces (Glow only???) in Route Viewer / Object viewer are glitched at the minute.

Classic example of why this needs merging into a single library though, too many things differing between the three of them.....

leezer3:
Getting somewhere now.

Can't see any major current bugs, and the biggest lump (RenderFace) is now in the new library.
Current diff shows a net loss of ~3000 duplicated LOC, but that's probably low, as XML-doc has been added to everything moved.

Next major push is to combine the object lists and hide/ show functionality. This will require marginally more work, but the gain will be worth it.

leezer3:
Lists and show / hide combined into Route Viewer.
Something is still a little funky, as enabling openGL display lists breaks stuff, but ignore for the minute.

Next need to cross port the backgrounds somehow?

leezer3:
Another fast-forward and squish.

Route Viewer and Object Viewer now run off the same renderer code as the main sim.
We appear to have a minor glitch for one frame when jumping between stations on RV (Something to do with jumping in the middle of the frame?), and the background code has been ported across to Route Viewer.

The background code is currently in the renderer lib for ease of porting / makefile, but this needs to change to it's own library (BackgroundManager.dll ?), and to have the dynamic background loader code contained within this.

leezer3:
Another PR which has got out of control :(

Starting to merge in parts of this one manually.

leezer3:
Fast forwarded in steps, and hopefully have the makefile working.

Really want to move the MeshBuilder to the API next, but don't do this until we have the thing building under CI again....

leezer3:
This makes a start on shifting the TrackManager into a separate shared DLL-
This partially works, but will be dependant on moving the TrainManager also for much of this work.

leezer3:
Another squash and rebase....

Considering fixing the makefile and merging into master at the minute, as this is way OTT in size.
On the flip side, there are still hacks which will want removing in the final part of this......

leezer3:
The shared signal manager isn't directly contributing to this, but it eliminates another bunch of shared code.

Next major aim would like to be abstracting the base TrackManager out into a shared library.

PR #279, opened 2018-09-26
New: Add command that adjust to the notch directly from command option value

s520:
This PR add command that adjust to the notch directly from command option value.
In the screenshot below the power notch is set to 5.

command_option

Assign the command to be added this time to the joystick and connect the contact of the joystick to the master controller.

Connect from each joystick button with wiring code by soldering, to create very easily the real cab's controller, such as power notch, brake notch, or any real buttons.


leezer3:
Good idea.
Small improvement- Please can you either hide or disable the new label / updown when it's not relevant to the current command?
Should be able to do that by checking in the same place where the description is set :)

PR #280, opened 2018-09-29
New: Re-write the DirectX parser

leezer3:
The current DirectX parser is a convoluted mess, requiring all kinds of string comparisons and other funky stuff.

This PR replaces the DirectX parser with entirely new code, derived from the MSTS parser and some other bits and bobs.

Flaws:
- It now uses the MeshBuilder concept, which may possibly be marginally less efficient than previous.
- ~~Frames not yet supported.~~
- ~~Some Binary X files are still broken with this new parser- [treeSet2.zip](https://github.com/leezer3/OpenBVE/files/2430664/treeSet2.zip)~~
- Testing is currently marginal. Propose this is added as an option once complete and tested better, with a fallback to the original parser if it barfs on a file.
- Error logging isn't perfect.


leezer3:
Squashed, fast-forwarded and a few more bugs fixed.

Issues:
* Template blocks aren't properly supported / ignored.
* Blender files broken (see the attached sample in #289 )
* ????

leezer3:
This now loads the blender file linked above.
No matrix operations yet....

leezer3:
Have implemented a basic Vector4 and Matrix4D into the API.

The parser now also contains basic transformations (Primarily POC), however these need serious re-doing.
Issues:
* Frame detection and cloning is in need of a rehack; The detection of the root frame currently relies solely upon the presence of the **Root** label, whereas we really need to have full-level detection (Two 'root frames in an object?)
* Normals aren't transformed at the minute. 
* Can we add a matrix transform operator to the Vector3 to cut down on the code? Should be easy!
* The blender file demonstrates transparency issues. Need to investigate the cause of this, as the Mesquoia file does not, and these have come from the same source.

leezer3:
Force push as I included the wrong binary dll version.....

leezer3:
https://github.com/assimp/assimp/blob/master/test/models/X/BCN_Epileptic.X
human

This displays correctly, and is a very complex 3DS Max text based model. (679kb)
On the downside, the block creation process is slow, and we need to see if there's a way we can speed this up.

leezer3:
#291 merged into this PR to make things a little tidier.

Current flaws:
* Assimp parser needs to be changed to use MSZip decompressor.
* Check handling of frame transforms by Assimp

leezer3:
Nuts, makefile broken. Need to add a makefile reference to Assimp to the main sim, will do that later along with some more minor tweaks from Resharper.

leezer3:
Some more playing around reveals Assimp is case-sensitive when parsing a text format file.

Need to check whether BVE4 and BVE5 are, but case sensitivity isn't a brilliant idea.

s520:
Assimp original code is as case-sensitive.

I will modify the code if necessary.

leezer3:
Have raised the Assimp issue with them to see if they actually want to handle this, along with a suggested patch (This appears to work, but whether this should be considerered a valid X file is probably up for debate)

https://github.com/assimp/assimp/issues/2238

leezer3:
Pushed fixes for a couple more variants in the NewXParser.
[OWN.zip](https://github.com/leezer3/OpenBVE/files/2616002/OWN.zip)

This one is technically broken (extra closing bracket halfway down the file), and appears to be a test export from something, but Assimp loads it correctly, and if there's one like this, there are probably more....

Also have handling for an X file variant which has a single material face index- This should be replicated to all faces. If not, the number of material face indicies must match the number of faces. The new parser and Assimp currently throw an exception in this case, need to test whether there are any in the wild, and if so what Michelle's parser does.

leezer3:
Another probably invalid X file which neither of the new parsers likes:
```
xof 0303txt 0032

Header {
 1;
 0;
 1;
}

Mesh {
 0;
 ;
 0;
 ;

 MeshMaterialList {
  0;
  0;
  ;
 }

 MeshNormals {
  0;
  ;
  0;
  ;
 }

 MeshTextureCoords {
  0;
  ;
 }
}
```

leezer3:
Another null object:
```
xof 0303txt 0032

Mesh {
 2;
 0.000000;0.000000;0.000000;,
 0.000000;0.000000;0.000000;;
 0;
}
```

Have patched this one, but we'll definitely need to be careful as these seem quite common in BVE4 content. 

We also now have the option to choose the X parser in the **Advanced Options** page of the main sim, and in a routefile, with the following code:
```
Options.XParser xParserNum
```

**xParserNum** may be set to the following:

**0** - Michelle's original X Parser.
**1** - The New X Parser.
**2** - Assimp

PR #282, opened 2018-10-11
Fix: Issue where the language setting shifted in some languages

s520:
This PR fix issue where the language setting shifted in some languages.

### DIFF
--- a/source/OpenBveApi/Interface/Translations/LanguageFile.cs
+++ b/source/OpenBveApi/Interface/Translations/LanguageFile.cs
@@ -57,7 +57,7 @@ namespace OpenBveApi.Interface {
                 }
                 Array.Resize<string>(ref LanguageFiles, n);
                 Array.Resize<string>(ref LanguageNames, n);
-                Array.Sort<string, string>(LanguageNames, LanguageFiles);
+                Array.Sort<string, string>(LanguageNames, LanguageFiles, StringComparer.InvariantCultureIgnoreCase);
                 comboboxLanguages.Items.Clear();
                 //Load all available languages
                 for (int i = 0; i < AvailableLangauges.Count; i++) {


leezer3:
Can you describe exactly what's going on for this one?
It shouldn't affect anything that I can think of.

I actually added the InvariantCulture check today, as I noticed that the comparison would be broken in a system using a RTL language (e.g. Arabic or Hebrew) as the system language, no matter what the UI language was set to.

s520:
Setting the language setting to Japanese will result in code = ko-KR instead of code = ja-JP at options.cfg.
So I think this issue is caused by [Change: Minor tweaks to API code](https://github.com/leezer3/OpenBVE/commit/a8c82439dc344730590bc2b5dc9758765d91e640).

leezer3:
I can't think how this change would cause that-

All it does is specify that the string should be compared R==>L case insensitively, using the period as the decimal separator & a couple of other mimor options.
Let me try and reproduce this on a Japanese Linux box before we do anything else;.

leezer3:
Hmm.
I can't immediately reproduce anything on Windows or a Ubuntu 15.04 box.

I'm still digging, but I can't find anything that looks like it could cause that....

s520:
Thank you for investigating.
I thought that this problem was caused by the disagreement of the order of "LanguageNames" and "AvailableLangauges".
So I unified the sorting rule into InvariantCultureIgnoreCase.

s520:
I tried that branch and confirmed that this problem has been fixed.
Thanks!

leezer3:
No problem.

I'm doing some further work in that branch restructuring things to generally make that section a little cleaner, but will try and get that merged in tomorrow or Monday.

leezer3:
Should now have be in the current mainline, with https://github.com/leezer3/OpenBVE/commit/33a9452d966935e76bc6c97a7b40bb8fba51444a

This also does a little trickery to load an embedded copy of the en-US language file from within the API if no language files are found.
(N.B. I briefly considered embedding all the language files, but this hides them from being easily user editable)

PR #285, opened 2018-10-30
Add some functions

s520:
This PR add the following two functions.
- New: Add the function of Input Device Plugin
- New: Add restriction of notch that driver can handle

First of all, I added the function of Input Device Plugin of BVE5.

With "SanYingInput.dll" using this function, [OHC-PC01](http://bvets.net/jp/drive/controllers.html) is made available for OpenBVE.

This plug-in ported ["三鶯重工 OHC-PC01 (for BVE) BVE device plug-in"](https://bitbucket.org/mikangogo/sanyinginput) provided under the MIT license.

In addition, "DefaultDisplayPlugin.dll" is a plugin that can change the default display of the game screen.

Next, I added parameters that limit the number of notches that the driver can handle.

This function is used as follows.

The vehicle using plural gears such as diesel car, chooses the accelerating force curve to increase for one of the number of the gears in plug in and controls it.

In addition, it also used for increasing acceleration of the train by selecting a noch that not usually used or switching to high acceleration mode.

Some trains change its acceleration by difference in voltage, and switching it by plugin.


leezer3:
Some thoughts:
* Version checking for train.dat files was implemented a while back (See at the top of the parser). Restricting the number of notches available to the driver directly is definitely a good idea, but it should check for the specific version first, as it's a breaking change.

## Input plugins
* The makefile doesn't currently build the input device plugins.
* The input device plugins should be set to build in the main solution; At the minute, they seem to need manually building.
* SanYingInput's window is a little too small, and cannot be resized by the user:
btest

**However:**
Have you considered implementing SanYingInput in the same manner as the RailDriver though, without the plugin?

All it's really doing is holding a fixed list of buttons / axis and their positions, which it then translates to the internal control positions as appropriate.
This overrides any controls set in the Options dialog, which is confusing at best.

If you've got one of these available, could you post the HID page and PID it uses?
If it's not using a generic controller board, restricting it should be easy enough. 

I'll try and drop the makers a mail and see if I get anywhere too.

leezer3:
OK, so the makers have responded.

They're using a commodity USB joystick board, with VID 0x0079 & PID 0x006.

Essentially, it's just a bitmask using 4 button inputs to create a notched power lever (Rather clever little trick!), along with additional functional buttons.
ohc-bve01 key definition

This seems to be in a bunch of the Chinese PS3 pad clones & similar stuff, which means that detecting this device specifically is likely to be an issue.
Now I know exactly what the thing is doing, it's easy enough to implement it internally (as opposed to via a plugin), but we'd need a way of setting a joystick to be one of these specifically. 

Do you actually have one available?

s520:
ginga81 wrote:

Hello,I am ginga81.

In Japanese comunity, we  has been discussed by below process.

Finally, we implemented this Pull request.

1.At the first, from a BVE5's plugin creator suggested us that the same system of BVE5's input device plugin management system should add to OpenBVE.

2.Additionally, a BVE5's plugin creator that he uses an BVE4's bug, he created the special plugin for BVE4.

At BVE4, the power notch can set to 8 for player and plugins officially.

But if using a bug, only for the plugin can set to notch line define up to 10.

He suggested to us need to new function for OpenBVE.

This is not a bug such as BVE4, separate the player's maximum notch and plugin's one.

3.When we are attempt to implement these suggestions, we thought that if the BVE5's input device plugin change only slightly, and easy to implement to OpenBVE, the input device plugin manager is more useful.

Then we thought that need for a sample plugin.

The base sample plugin is for BVE5, and it is necessaly to easy changes for OpenBVE's.

We searched plugins, and we found the SanYingInput plugin for BVE5 is a very good example. 

And this joystick device is very nice, useful device especially for Japanese.

This plugins sourcecode's license is an open source, and easy to change slightly and implement to OpenBVE.

4.We do not know what act the Raildriver, but SanYingInput is recognizing combining several joystick button to one action.

At the config setting can not set to one action by the several joystick button combination.

So we need to create the new plugin.

5.We passed through these processes, we created the new input device plugin system and SanYingInput plugin.

At these process, we did not discuss to include SanYingInput to OpenBVE.

S520 bought OHC-PC01, and already tested at Windows and Linux(Fedora).

leezer3:
Some thoughts:

- openBVE itself will support (in theory) an infinite number of power notches. The only thing I'm aware of which uses 10+ is my Western, and these are deliberately combined with the plugin to produce the gearbox simulation.
- Separating the player's max notch and the plugin's max notch is absolutely fine, but as per my review comments above, I think this needs to be only enabled when the train.dat version is set to support it.
- The controls system is rather different between BVE and openBVE, specifically in that BVE only really supports the use of a single joystick **or** the keyboard, whereas in openBVE we can use a combination of multiple keyboards, joysticks etc.
- Whilst I haven't got an issue as such with using a plugin and fixed control set to support the OHC-PC01, the fact that it's also a 'standard' joystick is a little more problematic, especially as the board in it is in a lot of other stuff, and not easy to identifiy specifically. At the minute, if we enable your plugin, both the keyboard controls and the OHC-PC-01 affect the train, which in an ideal world wouldn't happen.
- No issues with a sample plugin :)

A RailDriver is a special type of HID device, and doesn't show up as a standard joystick.

s520:
Thank you for reviewing my cord.

First, as you suggested, I changed it to do the version check of train.dat.

Secondary,  about plug in and the OpenBVE reacting each when we use joy-stick.  

We think that this is no problem because user can turn-off "Joysticks enabled" as necessary.

This comment was translated by F81 tec200. I really appreciate it.

leezer3:
That's reasonable enough.

It might be 'better' if an input plugin could disable joysticks when enabled if required? 
Don't know if you think that's a good idea or not?

leezer3:
That sort of thing was what I was thinking of.

Don't know if it's a good idea though, I'm just trying to reduce the inevitable confusion that will happen.......

s520:
If the plugin displays such a message, it will be loaded.

The settings as to whether to load plugins are saved, and plugins are loaded automatically the next time.

Unfortunately, I do not have a good idea how to display messages when OpenBVE itself is launched.

I think that it is better to have a warning message output to the log file.

s520:
If someone who are attempt to implant OHC-PC01 to real driver's cab controler, and parallel use the joystick button with soldering connect to real switch such as head light and wiper etc, I think that the joystick turn off forcibly is not better.

inputdevice9

I added a warning message about the input device plugin.

leezer3:
Good point, thanks :)

Do you want me to merge as-is at the minute, and tweak more later if required?

s520:
Thank you.

I really need your help.

PR #286, opened 2018-11-01
Change: Use API readonly vector defines where possible

leezer3:
This PR is a slightly more extensive followup to the commits over the last few days.
https://github.com/leezer3/OpenBVE/commit/e051f97dfebbf47bfc5a6f4fc1ce10f2f3f31f5f
https://github.com/leezer3/OpenBVE/commit/4f5fd5191186cd57a90d9a59c9fa6270129ac944
https://github.com/leezer3/OpenBVE/commit/243448a5347d5226905503321b1c0756bfe6a201
https://github.com/leezer3/OpenBVE/commit/4af4adc89cf4b39817967282485d7418959d8356

All of this is partially micro-optimisation, and partially readability. 
We use a lot of pre-defined common vectors in places such as transformations etc.
These were already defined in the API, so it makes no sense to create a new copy.

It may produce a (minor) speedup, depending on exactly how the compiler optimises the old code.
The potential risk of this is updating something by reference accidentally.
As the API vectors are defined as readonly, this shouldn't happen (or compile for that matter), but test this a little more before merging.

### DIFF
--- a/source/CarXMLConvertor/Convert.SoundCfg.cs
+++ b/source/CarXMLConvertor/Convert.SoundCfg.cs
@@ -13,7 +13,7 @@ namespace CarXmlConvertor
 	    private static MainForm mainForm;
 	    internal static Vector3 DriverPosition = new Vector3(0, 1, 0);
 	    //3D center of the car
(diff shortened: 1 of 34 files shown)


leezer3:
CI failure is unrelated.

Issue #289, opened 2018-11-10
Cannot display DirectX file at ObjectViewer.


ginga81:
![railshinkansenballast](https://user-images.githubusercontent.com/22734699/48297875-a6b80b80-e4f5-11e8-977c-41f06b4b80c2.jpg)
These sample of Directx object files exported from Blender-exporter and Metasequoia 3.0LE, these cannot display at Object Viewer.
Please correspondence these DirectX files.

Blender:2.79b with DirectX exporter(Built-in)
RailShinkansenBallast.blend.x

Metasequoia 3.0LE
http://www.metaseq.net/jp/download/2/
metaseqle30.exe
RailShinkansenBallast.x

ObjectViewer:OpenBVE-2018-11-09.zip

logs:
Warning		Unsupported template template Header encountered in x object file RailShinkansenBallast.x
Warning		Unsupported template template Matrix4x4 encountered in x object file RailShinkansenBallast.x
Warning		Unsupported template template IndexedColor encountered in x object file RailShinkansenBallast.x
Warning		Unsupported template template Boolean encountered in x object file RailShinkansenBallast.x
Warning		Unsupported template template Boolean2d encountered in x object file RailShinkansenBallast.x
Warning		Unsupported template template MaterialWrap encountered in x object file RailShinkansenBallast.x
Warning		Unsupported template template MeshFaceWraps encountered in x object file RailShinkansenBallast.x

Operating system
Ubuntu 16.04LTS
Metasequoia:Wine3.0 stable

Sample data of DirectX Object file:
[ballastx-issue.zip](https://github.com/leezer3/OpenBVE/files/2568151/ballastx-issue.zip)

ginga81:
I am very sorry.
From Metasequoia directx object is incredibly minimum.
Please zoom in maximum at Object Viewer.
DDS picture's texture color is not good.

s520:
In my environment, the color is not good like the screenshot below.
![343](https://user-images.githubusercontent.com/29241703/48299090-8135fc80-e50b-11e8-9b0c-f3e178c27a2a.png)

My environment
- Windows 8.1, Intel HD Graphics 4400
- fedora 28, Intel HD Graphics 500

leezer3:
The DDS issue appears to be because your texture uses RGBA compressed DXT5.

The plugin appears to expect DXT5 to be compressed as BGRA (It's heavily derived from Nvidia reference code).

[...]

PR #291, opened 2018-11-13
New: Add X file parser using Assimp

s520:
This PR add X file parser using Assimp.
#289


leezer3:
That's very interesting :)

A couple of initial observations on the code from a brief play above.

Other than that:
* Can you pass any error messages from Assimp?
* Related to the above, if the file is an X file, but Assimp fails to load it ( https://github.com/leezer3/OpenBVE/files/2430664/treeSet2.zip ; This is a valid compressed binary, but there's a bug somewhere in every C# deflate compressor that I've tried which truncates the file), we get no error message.
* The faces on the blender track model from the linked issue should be both sides. They only appear to be on the backface. 
* I suspect Assimp is failing to load one rail in the Blender exported model. (open3dmod viewer displays this issue too, so it's not your code)
* The Mesquoia object fails to load, complaining that the normals count is wrong. Again haven't investigated this properly, but I think it's in your MeshBuilder code somewhere.

---

Overall, I'm not certain this is the best method to go mind. 

My absolute preference would be to get my X parser into a 100% functional state, simply because it's original code which I designed, and is much less likely to have any licencing comebacks later down the line. (Note: I'm aware Assimp is relatively close to MIT, but pulling anything out of a third-party project essentially verbatim will always end up messy)

s520:
The display of BCN_Epileptic.X has failed.

However, it seems that the X file has been successfully parsed.

The attached file is converted into CSV format based on the analysis result.

This is correctly displayed.

[BCN_Epileptic.zip](https://github.com/leezer3/OpenBVE/files/2597171/BCN_Epileptic.zip)

PR #292, opened 2018-11-23
Update localization files

MarcRiera:
Following the changes to TrainEditor, which added localization support to it, the localization files had become out of sync, because the new strings were only added to English, Japanese and Portuguese. This pull request adds the new strings (and other previous strings which had not been added at the time) to the rest of languages.


leezer3:
Thanks!

Note however: If a string is not present in a translated language file, it'll just use the string from en-US, so this isn't vital, just a convienence for translators.

Really ought to try and setup something to allow this to happen automatically when a new string is added....

PR #294, opened 2018-11-26
Fix: Issue where parsing failed with some X files at AssimpXParser

s520:
This PR fix an issue where parsing failed with some X files at AssimpXParser.
https://github.com/leezer3/OpenBVE/pull/280#issuecomment-441667168

However, since the number of vertices is 0, an exception "nVertices must be greater than zero" will occur ...


leezer3:
Thanks.
Might need to rethink the exception in this case- Maybe just add a warning for a file with nothing in it, but it's really minor & doesn't affect anything visually so can probably be ignored.

PR #297, opened 2018-12-10
New: Add Obj file parser using Assimp

s520:
This PR add Obj file parser using Assimp.

With the original Obj parser the following Obj file is not displayed.
[ballastobj.zip](https://github.com/leezer3/OpenBVE/files/2662434/ballastobj.zip)

This file can be displayed on a parser I ported to C#.

In addition, ObjectViewer's parser settings can be saved.

Also added Options.ObjParser for route file.


leezer3:
Would it be better to combine the two Assimp DLLs into one?
This at least then keeps all the Assimp code in one place.

s520:
OK. I try it.

leezer3:
Thanks!

I've also fixed the primary problem with your object in the current wavefront parser- I was reversing the alpha value so it was 100% transparent :(

PR #298, opened 2018-12-11
New: Add microphone input

s520:
This PR adds the syntax called Track.MicSound that plays the sound from the microphone on the simulation.

The usage of this syntax is as follows.
`Track.MicSound(X, Y, BackwardTolerance; ForwardTolerance)`

Specify the range to be played by BackwardTolerance and ForwardTolerance.

The PLAY_MIC_SOUNDS command toggles playback on and off.

The PLAY_MIC_SOUNDS command is assigned to W key by default.

### DIFF
--- a/assets/Controls/Default keyboard assignment.controls
+++ b/assets/Controls/Default keyboard assignment.controls
@@ -34,6 +34,7 @@ MAIN_BREAKER, keyboard, B, 0
 ENGINE_START, keyboard, E, 0
 ENGINE_STOP, keyboard, E, 2
 DEVICE_CONSTSPEED, keyboard, BackSpace, 0
+PLAY_MIC_SOUNDS, keyboard, W, 0
 SECURITY_S, keyboard, Space, 0
 SECURITY_A1, keyboard, Insert, 0
 SECURITY_A2, keyboard, Delete, 0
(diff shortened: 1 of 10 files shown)


leezer3:
Interesting :)

Can you explain a little more as to what you're trying to achieve with this one?
I'm presuming you must be attaching this to a some sort of physical simulator or something?

If that's the case, I'd wonder if the 'better' place to attach the sound to would be to a train car?

s520:
We would like to make OpenBVE available as a conductor simulator.

We decided to reproduce the station announcement as the first step and added this syntax.

s520:
We aim at the simulator which can simulate a the duties of a conductor, a driver, and a station attendant of the Japanese railway system at once.
As the beginning, we had realized door-catching in each car.
Then, we put a speaker in 3D space to enable the announcement in the station, and enable simulation to broadcast the voice of own at a station, this is our pull request.

This feature let you put a plural number at constant distance in conformity to the station, and it is possible for the acoustic influence range as setting possibility realistically.
Also, it is possible for the reproduction of the outdoors speaker of the gatekeeper.
A sense of reality is given by reproducing that speaking with a microphone with a PC in 3D space.
The departure bell will enable sound reproduction in the future in 3D space, too; of the duties of the station attendant enable simulating it.
In the future, we are planning to display the display of a driver and the conductor viewpoint with the re-opening and shutting button of the door of the Japanese train and the other window as a conductor simulator separately.

ginga81 wrote and F81 tec200 translated.
I really appreciate it.

leezer3:
No problem.

I haven't really been able to test this much, but will merge into mainline.

PR #301, opened 2018-12-26
New: Add alternate sound option for continuous handle movement

leezer3:
This PR adds alternate handle sounds, played for when the handle is in continuous movement as per BVE5.

### Sound.cfg

**[Brake Handle]**
**ApplyFast** : Played on the second and subsequent continuous notch movement upto & including the notch before maximum.
**ReleaseFast** : Played on the second and subsequent continuous notch movement upto & including the notch before minimum.

**[Master Controller]**
**UpFast** : Played on the second and subsequent continuous notch movement upto & including the notch before maximum.
**DownFast** : Played on the second and subsequent continuous notch movement upto & including the notch before minimum.

### Sound.xml
**&lt;BrakeHandle&gt;**
**ApplyFast** : Played on the second and subsequent continuous notch movement upto & including the notch before maximum.
**ReleaseFast** : Played on the second and subsequent continuous notch movement upto & including the notch before minimum.

**&lt;MasterController&gt;**
**UpFast** : Played on the second and subsequent continuous notch movement upto & including the notch before maximum.
**DownFast** : Played on the second and subsequent continuous notch movement upto & including the notch before minimum.

### DIFF
--- a/source/OpenBVE/Parsers/SoundConfiguration/SoundCfg.Bve4.cs
+++ b/source/OpenBVE/Parsers/SoundConfiguration/SoundCfg.Bve4.cs
@@ -618,8 +618,14 @@ namespace OpenBve
 										case "apply":
 											train.Cars[train.DriverCar].Sounds.BrakeHandleApply = new TrainManager.CarSound(OpenBveApi.Path.CombineFile(trainFolder, b), panel, SoundCfgParser.tinyRadius);
 											break;
+										case "applyfast":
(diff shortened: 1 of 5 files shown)


leezer3:
Further notes:
* The current behaviour is to play the apply / release sound for each notch- This is maintained unless a new value is present.
* If the sound file for either of the new sounds is missing, this will gracefully fall back to playing the apply / release sound.
* The Apply / Release sound will always be played for the first notch movement if the handle is not at min / max. This is unavoidable unless we start delaying the handle application a couple of frames to see if the key will repeat (IMHO undesirable, and the current behaviour is probably better anyway)

PR #306, opened 2019-01-14
Change: Micro optimisations in the renderer

leezer3:
This PR makes some marginal optimisations in the renderer.

TLDR:
openGL internally operates on a 4-byte (32-bit) byte alignment.
We already deliberately cast our vertex co-ordinates to float when passing them to openGL, this does the same for our normals & texture co-ordinates, which appear to have been overlooked for this.

It also makes a couple of arrays passed to openGL as default parameters static readonly, as opposed to re-creating every frame.

### Previous:
```
2019-01-14 12:30:51 - OpenBve
Frames: 5463 - Time: 29640ms - Avg: 184.312 - Min: 164 - Max: 188
```
### New:
```
2019-01-14 12:28:15 - OpenBve
Frames: 5538 - Time: 29625ms - Avg: 186.937 - Min: 172 - Max: 191
```

FPS measurements are marginal mind, and we've definitely found in the past that the runtime caching things can have an effect. (Usually negated on the second load of the program)

This is sitting as PR for the minute until it's been a little more widely tested, just in case it causes lighting or texturing glitches on any edge cases.

### DIFF
--- a/source/ObjectViewer/ObjectManager.cs
+++ b/source/ObjectViewer/ObjectManager.cs
@@ -1397,9 +1397,7 @@ namespace OpenBve
             {
                 for (int j = 0; j < AnimatedWorldObjects[a].Object.States[i].Object.Mesh.Materials.Length; j++)
                 {
-                    AnimatedWorldObjects[a].Object.States[i].Object.Mesh.Materials[j].Color.R = (byte)Math.Round((double)Prototype.States[i].Object.Mesh.Materials[j].Color.R * Brightness);
-                    AnimatedWorldObjects[a].Object.States[i].Object.Mesh.Materials[j].Color.G = (byte)Math.Round((double)Prototype.States[i].Object.Mesh.Materials[j].Color.G * Brightness);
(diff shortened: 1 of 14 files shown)


leezer3:
A couple more minor little twiddles, stopping the allocation of 6 further doubles per frame, and changing to using operators for a color multiplication.

```
2019-01-14 12:58:48 - OpenBve
Frames: 5667 - Time: 29829ms - Avg: 189.983 - Min: 180 - Max: 194
```

It would really be interesting to know what the margin of error for this sort of thing was now....

leezer3:
```
2019-01-15 12:16:48 - OpenBve
Frames: 5705 - Time: 29766ms - Avg: 191.662 - Min: 177 - Max: 196
```
This one may well come into the margin of error threshold. Not sure if it will stay, everything in this branch at the minute represents experimental micro-opts...

More useful may be the fact that 95% of the time, we're using the inverse 255 representation of the color bits, and getting this via a multiplication.
We could therefore change to use a single color type (e.g. the current Color128 / Color96).
This though costs us 9 / 12 bytes of storage per struct. Experiment....

Issue #283, opened 2018-10-22
Extensions.cfg In ObjectViewer


OpenBveTrainSim:
Chris,
Suggesting that ObjectViewer.exe be made able to load Extensions.cfg train files
so that an entire train or consist may be viewed and corrected as needed
without the torture of loading and reloading the developing train and a test route
into the main program OpenBve.exe
Thanks
OBTS

PR #311, opened 2019-02-15
Add: some functions

s520:
This PR adds the following functions.

1. Function to set the state of the vehicle in ObjectViewer
2. Function to read extensions.cfg in ObjectViewer
3. Function to hide only the panel

1 is a function that makes it possible to set the state quantity of the vehicle in the Animated file.
At present, ObjectViewer can not preview files including state quantities related to vehicles, and it is necessary to start the main body one by one.
This inconvenience is solved.

2 is to solve #283.

3 is a function to hide the panel from the view from the driver's cab as BVE5 does. (CAMERA_NO_INTERIOR command)


assets/Languages/en-US.cfg
```diff
@@ -647,6 +647,7 @@ SECURITY_N = N/A
 SECURITY_O = N/A
 SECURITY_P = N/A
 CAMERA_INTERIOR = Switches to the train's interior view
+CAMERA_NO_INTERIOR = Switches to the train's interior view. However, the panel is not displayed.
```
leezer3: **CAMERA_INTERIOR_NOPANEL** would probably be a better name, certainly makes more sense in English :)

s520: Thanks!
I fixed it. Please check it.

leezer3:
Thanks :)
https://github.com/leezer3/OpenBVE/commit/b5c179eee99ec4e9b385813815f49b7d9b49ac19 has a very minor improvement in error handling, nothing special.

Couplers might also need implementing at some stage- It's possible to use a coupler to 'pull' a car through another, although that's IMHO very low priority. 

This trick was used by some European content to implement moving bogies.

I hadn't got around to adding a lot of this before, as it's unfortunately increasing the amount of code divergance between the viewers and the main program (e.g. there are three distinct copies of the train class  etc.), and I'd really quite like to get everything separated into smaller common libraries.
https://github.com/leezer3/OpenBVE/pull/276 is the current WIP playground for this, although it's not really right.
Additionally, the ongoing cleanup I've been doing is an attempt to get everything onto a similarly structured as possible codebase to make this easier :)

Issue #313, opened 2019-02-16
Issue with ObjectViewer on #312


s520:
### Description

When building from the makefile, the setting screen will not be displayed with F10 in ObjectViewer.
This phenomenon is common between Windows and Linux.
However, when building from Visual Studio or Mono Develop, this phenomenon does not occur.

### Reproduction

Build ObjectViewer from makefile.

### Related information 

* Operating system: Windows / Linux

leezer3:
No idea exactly why that helps, I'm presuming MCS must be partially initialising it somehow, but there we go.

s520:
I would like to confirm one, why use mcs?
Can we use xbuild instead?

s520:
This problem does not seem to depend on the version of mcs.
I am using the latest stable version of mono.
`Mono C# compiler version 5.18.0.240`

By the way, why use mcs? Why not xbuild?

leezer3:
https://github.com/leezer3/OpenBVE/commit/256e62ec23d59c8f76381c076b085d23afea3bda

Second attempt seems to work at this end on Windows, but it's broken the icon. (Adding it back again makes it crash as it can't find the embedded resources)

There's something a little funny about the form somewhere, can't put my finger on it at the minute.

mcs-
I didn't write the makefile myself, however:
It dates from ~2.5 years ago, and deliberately targets being able to build on Mono 3.2.8.
At the time it was written there were definitely issues with compatibility with xbuild vs. msbuild (I think they're discontinuing xbuild anyway in favor of msbuild?)

s520:
I used this method at first.
There was a demand from the tester that we wanted to display the setting screen.
With this method, the setting will not be reflected unless you close the screen one after another.

I think that you may raise the lowest version of mono already.
In many distributions update of mono is stopped ...

leezer3:
It's not quite clear what you mean there.

I **think** you're saying that something isn't refreshing after changing the settings?

If that's the case, then why not pull out the F5 refresh into a function, and call this immediately after closing the dialog?
Only issue with that would be that refresh of large objects can be slower than ideal.....

s520:
I am not confident that it will be transmitted in my poor English...

I mean the following.

- Animated objects are animated immediately after setting the state of the vehicle.
- However, the current method stops updating the screen of the parent window while the dialog is displayed.
- I want to update the screen of the parent window while I display the dialog.
- So I suggested a way to open a dialog in a separate thread.

leezer3:
[...]

PR #314, opened 2019-02-16
Fix: issue about #313

s520:
This PR fix issue about #313 .
The cause was that I incorporated the icon resource.
I deleted it and it worked normally. (mcs and Visual Studio)


leezer3:
That's interesting....

My 2nd commit breaks the icon, I couldn't get it to work otherwise.
Thinking about it, passing it into the new thread via the constructor might work, would have to try that.... 

As a secondary thought now I understand what you're trying to do better-
It might be an idea to lock on the Animated Object update & the update methods from the new form to prevent an object state changing mid frame.
I haven't investigated if this would have a performance hit though....

s520:
I have already locked on updating. Is there a problem?

- https://github.com/leezer3/OpenBVE/blob/master/source/ObjectViewer/System/GameWindow.cs#L38
- https://github.com/leezer3/OpenBVE/blob/master/source/ObjectViewer/formTrain.cs#L269

leezer3:
Ah, missed that :)
I was looking / thinking about within the object itself, but locking on the global update works.

PR #315, opened 2019-02-18
New: Add coupler stretch / compress sounds

leezer3:
This PR adds sounds played when the couplers stretch and compress.
These are ignored when the coupler distance is 0.

Sound.cfg:
```
[Coupler]
Stretch = file.wav
Compress = file.wav
```

Sound.xml
```XML
<Coupler>
    <Stretch>
        <FileName>file.wav</FileName>
        <Position>0,0,5</Position>
        <Radius>2.0</Radius>
    </Stretch>

    <Compress>
        <FileName>file.wav</FileName>
        <Position>0,0,5</Position>
        <Radius>2.0</Radius>
        </Compress>
</Coupler>
```

**Flaws:**
* The default coupler stretch / compress distance is 5cm- In a typical train, the slack is taken up almost instantly, and it doesn't sound perfect.
* We don't currently simulate unbraked cars, which means that the compress sound is only really heard as the train comes to a stand. 
* Using the default minimum radius, we can still hear more coupler sounds than I'd like. Perhaps drop the radius to 0.5?

### DIFF
--- a/source/CarXMLConvertor/Convert.SoundCfg.cs
+++ b/source/CarXMLConvertor/Convert.SoundCfg.cs
@@ -808,6 +808,42 @@ namespace CarXmlConvertor
                         }
                         i--;
                         break;
+                    case "[coupler]":
+	                    newLines.Add("<Coupler>");
+	                    i++; while (i < Lines.Length && !Lines[i].StartsWith("[", StringComparison.Ordinal))
+	                    {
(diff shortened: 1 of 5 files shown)


OpenBveTrainSim:
@leezer3 
Coupler sounds during acceleration\deceleration now seem a bit exagerated\amplified\unrealistic on some trains.  Thanks
OBTS

PR #317, opened 2019-02-19
New: add touch elements

s520:
This PR adds a touch element.

OpenBVE / BVE5 is chronically suffering from a key shortage.

So we decided to add a touch operation.

Please see the next video.
https://twitter.com/S520_BVE/status/1097600408912969728

In addition, the sample is [here](https://github.com/leezer3/OpenBVE/files/2880098/OpenBVE-Touch_test_files4.zip).

In order to operate this sample data, following data:
- F81 tec200's N700A Shinkansen train (https://www.youtube.com/watch?v=tBweD9T2b3Y)


s520:
I made it possible to apply also to 3D cab.

Please see the next video.
https://twitter.com/S520_BVE/status/1098009138343292928

In addition, the sample is [here](https://github.com/leezer3/OpenBVE/files/2884091/113-1000atccab-3D_2.zip).

In order to operate this sample data, following data:

- PM's Series-113 train (http://bve.jpn.org/jr113.html)

leezer3:
Very neat!

That's I think worthy of a major version number increase :)

I see you're using the openGL selection buffer, hadn't thought of that.

It'd probably be possible to create an automatic parser for existing Panel2.cfg files.
Not going to be perfect, but it might well get a lot more traction on the idea.
From a quick initial thought, all it'd really need would be detection of the common subjects, and perhaps something like a size check (to weed out small things like a digitial display showing the power notch)

s520:
Thanks!!

Is automatic parser used by the user to convert panel2.cfg to panel.xml?

leezer3:
Correct, that's what I was thinking of.

I'm not sure how well it'd work without trying it- There are some heavily layered panels (a
and presumably openGL selects the top polygon in a stack of transparent ones), but for example, we could assume something like this:

* Check if element is a DigitalNumber
* Check if subject is power / brake etc.
* If on-panel size is something over (this is just an example, not measured it) 128 x 128, then we can reasonably safely assume this is a power handle.

It's nowhere near perfect, and would probably need to be disabled by default, but it may well give wider adoption.

leezer3:
Another couple of small improvements:- 
* If over a touch item, I think it'd be a good idea for the cursor to change (Probably to the hand symbol)
* A debug mode where touch capable polygons are outlined would also be very helpful. Easiest way to do this is probably to run through the face vertices for a second time with GL_LINE_LOOP as the begin mode.

s520:
It may be possible to change the hand icon by using SelectMode within the MouseMove event.

s520:
When you add a cursor image to Data/Cursors, you can select it with the optional second page.
The default cursor image is created by F81 tec200.
Please wait for completion.

leezer3:
Very nice :)
I've just hacked together much the same as you with regards to the touch selection, which seems to work reasonably well.

Might have to be a little careful in animated panels with massive 3D power handles, but overall the actual render process takes virtually no time compared to updating the verticies for all the animated objects within the world.

s520:
I added a custom cursor.
In addition, I added "Panel2 extended mode", a mode that enables touch operation with existing Panel2.cfg.
Please check it.

leezer3:
That works really well!

Trying it on some existing panels, I think we probably need a specific cursor for increase and decrease to help with showing where the touch zones are visually :)

These are the best I could come up with based upon your existing:
nk_m
nk_p

s520:
OK. Let's try it.

s520:
I finished adding symbols to the cursor.
Please check it.

leezer3:
https://youtu.be/wEbpIOOAGjY

Really great!

s520:
Thanks!
Could you please merge this?

leezer3:
Done.

s520:
Thanks!!

leezer3:
I've added a basic converter into CarXMLConvertor for existing panel2.cfg files in this branch:
https://github.com/leezer3/OpenBVE/tree/PanelConvert

Something not quite right though- The camera restriction is out of whack somewhere. I think this is in the panel.xml parser, as from a brief skim the output looks right. Haven't debugged, will try and do that tomorrow :)

s520:
Let me also see.

s520:
Camera restrictions exist in this section as well as Panel2.cfg.
I tried converting existing Panel2.cfg, but it seems to be working normally.

leezer3:
The 2D cab for the 81xx gives a very visible breakage:
http://www.bvecornwall.co.uk/wordpress/gwr-81xx-class-large-prarie/

At work, so can't do a screenshot.

s520:
I'll give it a try.

s520:
Oh, I understand.
It seems to be enough to call the panel.xml parser in the following places.
https://github.com/leezer3/OpenBVE/blob/master/source/OpenBVE/Parsers/Train/XML/TrainXmlParser.CarNode.cs#L284

leezer3:
That makes sense-
A 2D panel needs a different restriction to the 3D one & that code assumes 2D if it's a cfg & 3D otherwise.

Will think about properly when home.

source/OpenBVE/Parsers/Panel/PanelXmlParser.cs
```diff
@@ -0,0 +1,1629 @@
...
+			if (CurrentXML.Root == null)
+			{
+				// We couldn't find any valid XML, so return false
+				throw new System.IO.InvalidDataException();
+			}
+
+			IEnumerable<XElement> DocumentElements = CurrentXML.Root.Elements("Panel");
+
+			// Check this file actually contains OpenBVE panel definition elements
+			if (DocumentElements == null)
```
leezer3: I think we probably want to check for count > 0 here?
Otherwise, if valid XML we get a collection with zero nodes.

s520: As you say, I would rather check the count> 0 and throw an exception.

PR #318, opened 2019-02-23
Add panel conversion to CarXmlConvertor

leezer3:
Still needs animated panels converting too.

We could probably also automatically convert the touch zones?

Followup work to https://github.com/leezer3/OpenBVE/pull/317

### DIFF
--- a/source/CarXMLConvertor/CarXMLConvertor.csproj
+++ b/source/CarXMLConvertor/CarXMLConvertor.csproj
@@ -51,6 +51,7 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="Convert.ExtensionsCfg.cs" />
+    <Compile Include="Convert.Panel2.cs" />
     <Compile Include="Convert.TrainDat.cs" />
     <Compile Include="mainForm.cs">
       <SubType>Form</SubType>
--- a/source/CarXMLConvertor/CarXMLConvertor.csproj
+++ b/source/CarXMLConvertor/CarXMLConvertor.csproj
@@ -52,6 +52,7 @@
   <ItemGroup>
     <Compile Include="Convert.ExtensionsCfg.cs" />
     <Compile Include="Convert.Panel2.cs" />
+    <Compile Include="Convert.PanelAnimated.cs" />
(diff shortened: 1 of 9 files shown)


s520:
In the case of a 3D cab, the area where the object moves should be the touch area.
The problem is how to acquire the range of movement of the object...

leezer3:
Yeah, that's a problem.
It might just be easier to document adding touch zones and leave it for the developer to play with.
On the flip side, an automatic conversion to something near-right would probably increase uptake.

---

We could probably make an acceptable best-guess on movement directions based upon the following:
* Object's center point- Assume that our touch zones are to the front / rear of this.
* In 'standard' railway practice, the direction of motion of any safety critical control (e.g. power or brake) is that when pushed away from the driver's position, it will decrease power, increase brake etc.
* Knowing the rotation / translation vector means we can probably move the zones to intercept.
With this, the target zones might be a little innacurate, but there.

---

The other possible solution would be to allow the mouse movement to control the increase / decrease, so that for example, a swipe up increased and a swipe down decreased. In this case, I suppose you'd only use the control for targetting, and it's actual movement direction would be irrelevant.

This would require rehacking of the selection code though, and probably wouldn't look as nice....

s520:
I think the same idea as Panel 2 extended mode is good.
But is there a way to know the size of the object?
I feel that it is not easy.

Also, I think it is difficult to use mouse movement for control.

leezer3:
Not easily I'd say.
You could use the vertex coords as above, but you start to get into the problem that you're making vast numbers of calculations for dubious benefits :)

I think we can (probably) make some reasonable basic assumptions about size if we look at the real thing.

* Generally speaking, power / brake controllers are designed to sit on a desk and be operated with one hand.
* The total travel from min - max will probably be 15 -20cm, or close to 180 degrees if rotary.

I haven't tested this, but based upon that I'd say a touch zone of 10cm either side of the center of the object is likely to be reasonable.

leezer3:
The last commit on this branch removes the panel.animated.xml concept, in favour of parsing the schema for the panel.xml file to check it's type.

Not certain it's the best idea, but I think it's marginally less confusing.

PR #320, opened 2019-02-23
New: Panel animated extended mode

s520:
This PR add "Panel animated extended mode".
This mode add automatically touch zone at panel.animated.
Please add the following to the touch section of option file.
`panelanimatedextended = true` 
If there is no problem, I will make it configurable by GUI.

### DIFF
--- a/source/OpenBVE/Graphics/Renderer/Touch.cs
+++ b/source/OpenBVE/Graphics/Renderer/Touch.cs
@@ -400,11 +400,13 @@ namespace OpenBve
 						{
 							switch (TouchElement.Command)
 							{
+								case Translations.Command.SinglePower:
 								case Translations.Command.PowerIncrease:
 								case Translations.Command.BrakeIncrease:
 								case Translations.Command.ReverserForward:
 									Status = Cursor.Status.Plus;
 									break;
+								case Translations.Command.SingleBrake:
 								case Translations.Command.PowerDecrease:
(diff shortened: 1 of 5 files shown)


leezer3:
Will take a look, bit probably not until tomorrow evening now :)

s520:
Oh, I forgot to check whether it is a single handle car.
I fixed this to work even with a single handle car.

leezer3:
Unfortunately, that's not quite as good :(

I suspect the zones I can't find are probably stuck somewhere at the center of rotation?
Perhaps we need the zone to traverse the entire height of the object?

Some test subjects:
UK 323 3D Cab 
http://railsimroutes.net/x-city_south/index.php

Reverser is reasonable.
Power handle isn't as good. The click zone is too far down.

D1015:
http://www.bvecornwall.co.uk/wordpress/br-class-52-western/
No touch zones findable at all. It's creating them in what seems to be approximately the right place, but nothing appearing.

FMCB 3000 series:
https://en.bvebarcelona.cat/downloads/trains/3000-series/
No touch zones findable again.

s520:
After all it is necessary to add the size of the object?

leezer3:
I'd guess so.

Don't know how much effort it's worth putting into it though- This could get complex very fast....

s520:
What about putting a vehicle with no license problem on the main program?
Of course it is for the touch compatible 3D cab.

PR #323, opened 2019-02-28
Fix: issue of loading panel.animated.xml

s520:
This PR fix issue of loading panel.animated.xml.
I fixed it to preferentially load the 3D cab.

### DIFF
--- a/source/OpenBVE/OldCode/TrainManager.cs
+++ b/source/OpenBVE/OldCode/TrainManager.cs
@@ -36,15 +36,16 @@ namespace OpenBve
 				Elements = new ObjectManager.AnimatedObject[] { },
 				Overlay = true
 			};
-			string File = OpenBveApi.Path.CombineFile(TrainPath, "panel.xml");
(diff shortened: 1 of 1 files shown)


leezer3:
That was actually deliberate.

With XML, there's nothing stopping us from having either a 2D or a 3D panel in the same file; All we need to do is check the schema & pass it to the appropriate parser.
With the native BVE2 / BVE4 formats it's a lot harder to do this without parsing the whole file (and for that matter, a panel2.cfg is only really an evolution of a panel.cfg)

With that in mind, it somewhat negates the need for panel.animated.xml as a specific deliminator for the 3D cab version.

Leaving the option for both names seemed appropriate though.

leezer3:
In other words, the XML panel is one 'format' but with both 3D amd 2D subsets if that makes any sort of sense?

s520:
I see.
Certainly in XML it makes a distinction between 2D and 3D.
I did not understand your intention.
I will close this PR.

PR #326, opened 2019-03-06
New: the Another Trains

s520:
This PR is make realisation of another moving trains.

First, please look at these movies for the 'Another' trains moving.
https://twitter.com/S520_BVE/status/1100878240057503745
https://twitter.com/S520_BVE/status/1101603017500053506
https://youtu.be/VYOz-gDbjOA
If you are using these functions, the another train can run on the another RailX freedomly.
This functions is very frexiblity as showing under.

1. The train that can stop any position freedomly.
2. The train can acceleration and deceleration freely, and set the acceleration rate and deceleration rate freely.
3. The train that can change to any rails (also when the train is running!)
4. The train that can change direction at the any position freely.
5. The train that can change to visible and invisible freely at the any time.
6. The visible and invisible status, that can change freely at any truck position of self train.
7. The completely simulation of Train.dat, that realize derailment.
8. automatic sound, based on sound.cfg.
9. The full outer view that realize from extensions.cfg.
10. the Animated object for Another train(most partial).
11. Track.Brightness that will be applied on Another train automatically.(for example at tunnel)
12. The train that can open/close doors freely at any distance.

This function can do these actions showing as bellow.

1. The train passing at main line when our train is waiting at the side of  line for pass from behind.
2. The opposite train comes, and waiting at the side line for pass the another train at the opposite main line, and passing from behind for opposite main line, finally, the waiting train is run to the behind.
3. On a single line, the train comes to the parallel platform from opposite railway. When the arriving the train at opposite platform, door is open and close, and run to backward again.
4. at single line, arriving and back, and moving to another rail.

We prepared a demo route that actually works.
[AnotherTrainsDemoRoute.zip](https://github.com/leezer3/OpenBVE/files/2938239/AnotherTrainsDemoRoute.zip)
The following files are necessary to operate this demo route.
- Mr. MidnightExpress-Ginga81's とーほぐ新幹線v0.09 (http://midnightexpginga81.web.fc2.com/touhoku_shinkansen/index_touhoku.html)
- Mr. F81_tec200's Series-E6 Shinkansen(https://www.youtube.com/watch?v=yRBr-Dsa1U4)
- Mr. MidnightExpress-Ginga81 and F81_tec200's Series-E6 Shinkansen outer view model(http://midnightexpginga81.web.fc2.com/touhoku_shinkansen/index_touhoku.html)

To test this demo route please follow the steps below.
1. Extract Touhogu-Shinkansen-Routedata-v0.09.zip and Touhogu-Shinkansen-Objectdata-v0.15.zip to OpenBVE's Railway folder.
2. Extract tec_E6_openbve.zip to OpenBVE's Train folder.
3. Extract Series-E6_OuterView_ver.0.01a.zip to the Tec_E6 folder of 2.
4. Confirm that you can operate any route file in the とーほぐ新幹線v0.09 folder(1. ) in 2. and 3. with Tec_E6 in 2. and 3.
5. Extract AnotherTrainsDemoRoute.zip to とーほぐ新幹線v0.09 folder.
6. Copy the Tec_E6 folder to the とーほぐ新幹線v0.09 folder.
7. single_line-waiting.csv etc driving on any train.

Next, I will explain the contents of demo routes.
- single_line-waiting.csv
This route crosses the train coming from the other side at the second station.

- arrive-moving-another-line_and_N-shape-switching_at_terminal.csv
In this route, the other train parformes like that:
1. One minute after the start of the game, arrive at the first station.
2. Turn to the end of turnout ahead of the terminal station.
3. Back to the turnout.
4. Turn to the track to be installed at home.
5. Trun back to the terminal station.

- all-passing-demo_at middle-station.csv
This route is overtaken after stopping at the second station, after which the opposite train arrived at siding, departed, departed, and traveled to the next station.

- other-aoba-brightness-test.csv
This route is for the test of Track.Brightness and starts moving backward when moving with CAMERA_POI_NEXT. Also, jumping to the "parallel run" station at the station jump will allow you to run in parallel.

Finally, I will explain the extension of Track.RailStart and Track.Rail related to other tracks.
In order to reproduce the cant in other trains this time we will introduce the concept of cant to other tracks.
However, this cant affects only the other trains and does not affect the rail object.
The extended syntax is as follows.
- Track.RailStart RailIndex; X; Y; RailType; CantInMillimeters
- Track.Rail RailIndex; X; Y; RailType; CantInMillimeters

CantInMillimeters tilts to the left if it is negative, and it tilts to the right if it is positive. It does not depend on Options.CantBehavior.

That all.  Please confirm.

This post was translated by Midnight Express Ginga 81 and F81 tec200. I really appreciate it.

### DIFF
--- a/source/OpenBVE/Game/AI/AI.OtherTrain.cs
+++ b/source/OpenBVE/Game/AI/AI.OtherTrain.cs
@@ -205,6 +205,9 @@ namespace OpenBve
 					}
 				}
 
+				Train.Specs.CurrentAverageSpeed = Train.Cars[0].Specs.CurrentSpeed;
+				Train.Specs.CurrentAverageAcceleration = Train.Cars[0].Specs.CurrentAcceleration;
+
 				// Dispose the train if it is past the leave time of the train
 				if (LeaveTime > AppearanceTime && SecondsSinceMidnight >= LeaveTime)
 				{
--- a/source/OpenBVE/Game/AI/AI.OtherTrain.cs
+++ b/source/OpenBVE/Game/AI/AI.OtherTrain.cs
@@ -1,0 +1,256 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace OpenBve
+{
+	internal static partial class Game
+	{
+		internal enum TravelDirection
+		{
+			Forward = 1,
+			Backward = -1
+		}
+
+		/// <summary>Travel plan of the train moving to a certain point</summary>
+		internal class TravelData
+		{
+			// Parameters from XML
+			internal double Decelerate;
+			internal double StopPosition;
+			internal double StopTime;
+			internal double Accelerate;
+			internal double TargetSpeed;
+			internal TravelDirection Direction;
+
+			// Parameters calculated by the SetupTravelData function
+			internal double DecelerationStartPosition;
+			internal double DecelerationStartTime;
+			internal double ArrivalTime;
+			internal double DepartureTime;
+			internal double AccelerationEndPosition;
+			internal double AccelerationEndTime;
+		}
+
+		internal class OtherTrainAI : GeneralAI
+		{
+			private readonly TrainManager.OtherTrain Train;
+
+			/// <summary>Travel plan of train</summary>
+			private readonly List<TravelData> Data;
+
+			private double TimeLastProcessed;
+			private double CurrentPosition;
+
+			internal OtherTrainAI(TrainManager.OtherTrain train, List<TravelData> data)
+			{
+				Train = train;
+				Data = data;
+				TimeLastProcessed = 0.0;
+			}
+
+			internal override void Trigger(double TimeElapsed)
+			{
+				// Trains need to stop more than 2 points.
+				if (Data == null || Data.Count < 2 || SecondsSinceMidnight == TimeLastProcessed)
+				{
+					return;
+				}
+
+				// Initialize
+				if (TimeLastProcessed == 0.0)
+				{
+					SetupTravelData();
+					CheckTravelData();
+					CurrentPosition = Data[0].StopPosition;
+				}
+
+				TimeLastProcessed = SecondsSinceMidnight;
+
+				// Calculate the position where the train is at the present time.
+				double DeltaT;
+				double Position = Data[0].StopPosition;
+
+				// The start point does not slow down. Acceleration only.
+				if (SecondsSinceMidnight >= Data[0].DepartureTime)
+				{
+					if (SecondsSinceMidnight < Data[0].AccelerationEndTime)
+					{
+						DeltaT = SecondsSinceMidnight - Data[0].DepartureTime;
+					}
+					else
+					{
+						DeltaT = Data[0].AccelerationEndTime - Data[0].DepartureTime;
+					}
+					Position += (int)Data[0].Direction * (0.5 * Data[0].Accelerate * Math.Pow(DeltaT, 2.0));
+				}
+
+				for (int i = 1; i < Data.Count; i++)
+				{
+					if (SecondsSinceMidnight >= Data[i - 1].AccelerationEndTime)
+					{
+						if (SecondsSinceMidnight < Data[i].DecelerationStartTime)
+						{
+							DeltaT = SecondsSinceMidnight - Data[i - 1].AccelerationEndTime;
+						}
+						else
+						{
+							DeltaT = Data[i].DecelerationStartTime - Data[i - 1].AccelerationEndTime;
+						}
+						Position += (int)Data[i - 1].Direction * (Data[i - 1].TargetSpeed * DeltaT);
+					}
+
+					if (SecondsSinceMidnight >= Data[i].DecelerationStartTime)
+					{
+						if (SecondsSinceMidnight < Data[i].ArrivalTime)
+						{
+							DeltaT = SecondsSinceMidnight - Data[i].DecelerationStartTime;
+						}
+						else
+						{
+							DeltaT = Data[i].ArrivalTime - Data[i].DecelerationStartTime;
+						}
+						Position += (int)Data[i - 1].Direction * (Data[i - 1].TargetSpeed * DeltaT - 0.5 * Data[i].Decelerate * Math.Pow(DeltaT, 2.0));
+					}
+
+					if (SecondsSinceMidnight >= Data[i].DepartureTime)
+					{
+						if (SecondsSinceMidnight < Data[i].AccelerationEndTime)
+						{
+							DeltaT = SecondsSinceMidnight - Data[i].DepartureTime;
+						}
+						else
+						{
+							DeltaT = Data[i].AccelerationEndTime - Data[i].DepartureTime;
+						}
+						Position += (int)Data[i].Direction * (0.5 * Data[i].Accelerate * Math.Pow(DeltaT, 2.0));
+					}
+				}
+
+				// Calculate the travel distance of the train.
+				double Delta = Position - CurrentPosition;
+				CurrentPosition = Position;
+
+				// Set the state quantity of the train.
+				if (Delta < 0)
+				{
+					Train.Handles.Reverser.Driver = TrainManager.ReverserPosition.Reverse;
+				}
+				else if (Delta > 0)
+				{
+					Train.Handles.Reverser.Driver = TrainManager.ReverserPosition.Forwards;
+				}
+				Train.Handles.Reverser.Actual = Train.Handles.Reverser.Driver;
+
+				foreach (var Car in Train.Cars)
+				{
+					Car.Move(Delta);
+					Car.Specs.CurrentSpeed = Delta / TimeElapsed;
+					Car.Specs.CurrentPerceivedSpeed = Car.Specs.CurrentSpeed;
+				}
+
+				// Dispose the train if it is past the leave time of the train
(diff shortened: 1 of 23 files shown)


leezer3:
Genius!

I may not be able to properly go through this and test for a week or so (I'm on holiday from Sunday), but will get to as soon as I can :)

Will do a little looking and leave a few comments as and when though.

leezer3:
A few very basic thoughts / questions (Incredibly brief skim, haven't tested this yet):

* **OtherTrain** isn't quite right as a name. Can't think of anything good at the minute though....
* Can we make the OtherTrain a subset of the abstract base train, and then by doing so eliminate the second OtherTrain array in TrainManager? Something like when running through the train array, if is ordinary train do X, else do Y?
* Marginally related- It's been on the todo list for a while to set it up so classic AI trains can be different to the player train. This would probably be a good place to do that. In principle in it's simplest implementation, all that really wants is a route command pointing to an XML which in turn points to the desired train folders.

s520:
- Are there any other good names?
I believe that this function can be used for cars as well as trains.
I am happy to think of the names together.

- In order to integrate TrainManager.OtherTrains into TrainManager.Trains, it is necessary to divide the operation depending on whether it is the OtherTrain class or not in some functions.
Since I thought that it should be divided rather than mixing, I created TrainManager.OtherTrains.

- I can not understand what you are saying about the third one. I'm sorry.

leezer3:
The third one:
If you use the PreTrain or RunInterval command, this creates a preceeding (or following) AI train.
The exterior, sounds and performance will always be identical to the player train at the minute :)

leezer3:
The second- Whilst I see your logic with this, it seems to me to be more 'logical'  to keep all the trains in one place.

Related to this, at some stage we might want to make these interact with the signalling system?

Anyone else with thoughts is most welcome to have a say!

s520:
I think that it is necessary for another train to be able to set a train different from the player's train.
Also, since the characteristics of other trains this time can be operated freely, I think that other trains can ignore the signal.
If you want to associate with a signal you should use a conventional AI train.

ginga81:
Basically, this technology is based on TrackFollower.
Additionally, this is a 'programmable move animated object'.
This is using by RailX, extensions.cfg, Train.dat, and sound.cfg, but these are not for 'train'.
But, to simulate motor sound that effected by speed, start, stop, animated and more, we adopt train simurate routine.
As a result, have to add the another rail's cant data.

When we want to add the animated object using for such as train, we are very difficult to use.
For example, the TrackFollower is must to use the calculation function, but it is not to use easily.

So, to add the train object and use more easily, we adopt the extensions.cfg.
The another train program is read the extentions.cfg, and add the each car automatically.
As the same, to use easily to add the driving sound, we adopt the sound.cfg and train.dat.

So he said that when we want to insert to another 'train', must use AI train technology.

To obey the signaling system such as ATS, need to use the positioning area check , time check and realizationing time, and more XML parameter.

leezer3:
Sure, at the moment they don't really behave like conventional trains, but I suppose the point I'm making is that it may well be a good design philosophy to deliberately assume that one day they may well become these.


I've lost count of the number of things that have had to be changed or worked around because the current code structure makes assumptions based upon what classic BVE content could or could not do.

---

In a similar development of this idea, it seems to me that we're actually quite close to making other rails driveable by the player.
I don't know if you've come across this one (European app), but BVE Route Randomiser used to be able to do this:
http://web.ncf.ca/fb214/ - Final link at the bottom of the page
As you've just demonstrated, the player rail isn't much more than a path in the world, and the point of reference by which the world is generated around.

**Something like this I think:**
Your code generates a track for the other rails. I haven't investigated exactly how you've done the curve generation (the hard part!). It also allows track followers to specify which track they are running on.
Add a routefile command (Something like `.switch Track1;Track2` ), and

s520:
I used the rail object's RailTransformation to create tracks to other rails.
https://github.com/leezer3/OpenBVE/pull/326/files#diff-5e96f155b1fe3cf321e83791e82d9ed1R659
So, strictly it is not a curve but a straight line between two points.

I was not assuming to run the player train to other tracks at all.
What you say may be possible using TrackManager.GeneralEvent. (`.switch Track1; Track2`)
If you allow the player to change points, I think that problems will arise at stations and signals.
I think that behavior should not occur unintended by the creator of the route.

leezer3:
A few minor comments from a basic test above.

I haven't tried to deliberately break this yet, but will do so at some stage.

---

Your comments are noted about the player on other tracks ideas.
I think that's the subject of a seperate PR when we've worked the kinks out of this one :)

s520:
After completing your test, will you merge this PR?

leezer3:
Absolutely.

The thing is that you're changing some fundamental stuff in the train manager & AI, and I want to be as sure as I can be that it's stable & the issues have been looked at before making it public.
This goes back to the design discussions- Once something is public & it works a certain way, it's very difficult to change in the future, even for the better :)

s520:
Thank you very much.

First of all there was a name problem.
I still do not have an answer about this problem...

Next is the problem of integrating TrainManager.OtherTrains into TrainManager.Trains.
Do you think that this should be integrated?
If you think so, I will work on integration.

leezer3:
Unfortunately, I don't have a good answer to either at the minute :)

There are pros and cons to both approaches, and my opinion isn't necessarily right, it needs to be a consensus! (This was one of the main ways things went wrong originally- It was either Michelle's way or not at all, and I'd like to avoid that)

s520:
I see.
As ginga81 says, this is "programmable move animated object".
I think that it can be used for cars, airplanes, etc. without limiting to trains.
So the name "other train" may not be appropriate.
However, "programmable move animated object" is too long.

ginga81:
We are not a native England, please suggest good short name of "programmable moving animated object".
We are native Japanese, so please forgive us.

JonnyTech:
> OtherTrain isn't quite right as a name. Can't think of anything good at the minute though....

[NPC](https://www.urbandictionary.com/define.php?term=NPC)'s?

OpenBveTrainSim:
Consider
"AIO" meaning "AutomatedIntelligentObject" or "AnimatedIntelligentObject"
or
"AIU" meaning "AutomatedIntelligentUnit" or "AutomatedIntelligentUser"

OBTS

leezer3:
TrackFollower?

It's an extension of the existing tech after all :)

OpenBveTrainSim:
TFO = TrackFollowingObject or TrackFollowingOperator
TFA = TRackFollowingAnimation

s520:
TFWO = TrackFollowingWorldObject ?
The reason to add 'World' is because the (player) train can also say 'TrackFollowingObject'.

ginga81:
Write something of candidate at here, please decide the best word by Mr.leeser3 or S520.
Or use good 'EMOJI' button for vote?

OpenBveTrainSim:
UFO = UnmannedFollowingObject or UnnaturalFellowOperator or UnwittedFriendlyOrganizism
With that I say good night, Ladies and Germs!
OBTS

s520:
Documentation for OtherTrain is below.
https://5c87e4a7ff03bd0008baaa20--openbve-japan.netlify.com/en/routes/xml/othertrain.html

Please check it.

leezer3:
Thanks, looks reasonable-
We can probably just drop in the HTML output into the existing site until the Hugo stuff is ready.

* Better wording for right under the example: `As you can see, the file consists of one <Definition> section, one <Car> section and one <Stops> section. The <Stops> section consists of two or more <Stop> sections.`
* In the stop attributes, there are two things I think you intended to be bold?
* For Time, something like this (We need to specify the units it uses) `Time sets the duration of the stop of the train in seconds. This value should not be an in-game time.`
* Same principle as the above applies to LeaveTime.

Other other small point:
For the direction, can we please accept F and R as well as 1 & -1? Just makes it that little more readable in human terms.


---

On the name subject, how about a compromise of TrackFollowingObject?
Adding World to it as well I think is getting a little too long :)

s520:
For now, I corrected your first and second point.
https://5c882d1d5655dd000a650e2c--openbve-japan.netlify.com/en/routes/xml/othertrain.html

The format of the time is hh.mmss, how do I write? I'm not good at English...

For the direction, I will fix to accept F and R as well as 1 & -1.

Certainly the name is too long, so I agree with TrackFollowingObject.

s520:
By the way, when do you switch documents to Hugo?
We have converted all the documents.

leezer3:
I thougjt you were still working on that one- Last I looked there was still essentially no translations.

Will try and deal with that this weekend maybe.....

s520:
Which document has not been converted? I forgot...

Certainly the translation is not yet complete, but we should call for translation assistance in other languages, such as in forums.

leezer3:
@s520 
Couple of things:
* Need a PR to the base Hugo docs repository for the documentation update so it syncs through.
* Any chance of the requested changes before merge?

Cheers :)

s520:
First, I need to change the name in both the program and the document.
Next, for the direction, I will fix to accept F and R as well as 1 & -1.
When these two are done, you are ready to merge.
I will complete the work tomorrow.

s520:
I am sorry to keep you waiting.
The merge is ready.
Please check it.

leezer3:
Thanks!

I've made some minor wording changes to the documentation, and implemented some of the tweaks I mentioned above and merged.

s520:
Thanks!.
Now I am developing a BVE5 parser.
It is still in testing, but I will attach it.
[OpenBVE-BVE5_Parser.zip](https://github.com/leezer3/OpenBVE/files/2996255/OpenBVE-BVE5_Parser.zip)
(I also use your branch code.)

PR #333, opened 2019-04-01
Remove outdated info from credits.html

MarcRiera:
I am not sure if this will be included in the Debian package, but as this was imported from the old openBVE website, there is outdated information about myself that I would like to remove/update. Thanks!


leezer3:
Redone the thing completely in plaintext not HTML, TBQH it was completely outdated.

Still working on this, need to sort out a changelog next I think.

PR #343, opened 2019-05-05
Move StaticObject to the API

leezer3:
In order to create objects via plugins, we need to move the StaticObject structure to the API.

This should build correctly, just putting it in via a PR to ensure that CI catches any issues....

When done, the next PR in this series will add an ObjectInterface to the API & move the object loaders to plugins.
Animated objects will be a little more complex (as they must load anonymous objects themselves), will think about that afterwards.


leezer3:
Note that this alone removes 1,700 duplicated LOC, plugins for objects should do a whole bunch more.

PR #344, opened 2019-05-05
Move MeshBuilder to API

leezer3:
This moves the next component necessary for Object Plugins to the API- The MeshBuilder helper class.

Again, should build with no real trouble, just want to run past CI before merging.


leezer3:
Note:
There will likely be some constructor changes in places once plugins are in the API.

It's better to keep things building and do this gradually though, as the last large change to shared everything kinda got overboard and still wasn't ready to merge.....
Follow-up to https://github.com/leezer3/OpenBVE/pull/343

PR #345, opened 2019-05-05
New: Move object loaders to shared plugins

leezer3:
This is currently WIP, but ongoing.

### Done:
* CSV / B3D
* X
* OBJ

### TODO:
* MSTS
* Loksim

My current thoughts are to combine the multiple obj / X loaders into one single base plugin, and then use a constructor call to set the desired outcome. 

Multiple plugins might be better, but enabling / disabling them might not be quite as easy :)

@s520 - Any thoughts on the best way to deal with this one?

---
So, our currently missing types are the animated / quasi animated ones.
These may be a little more complex, as the animated type calls back into other stuff.

I'm in two minds as to whether to merge this now, and work on the animated ones in a separate PR, or to try and do both in one....


leezer3:
Simple when you think about it- Just use an API method with object as the parameter type.
We then check if this is the type we want in each plugin and bingo!

Getting rid of 9,000 lines of duplicate code so far is massive.

leezer3:
Other thoughts- We could probably merge the DirectX helper plugin into the object loader plugin. 

Don't want to do this for the MSTS plugin when I get there though- All MSTS data files share the same structured format, and this may well allow us to load cabviews etc. in future :)

leezer3:
I hate makefiles....
I think we're just missing the reference to the API for our new DLL, will have a look tomorrow probably now.

leezer3:
Okies, will merge the Formats.DirectX with the new plugin tomorrow and merge I think......

PR #347, opened 2019-05-09
Change: Create matrix methods in openBVE API & use these

leezer3:
This PR changes the matrix operations to use native API copies.

Don't know if this will be merged, any other opinions welcome.

### Benefits:
* This makes it much easier to move away from openTK (and potentially openGL) if ever required- We can now do the world / camera space conversions internally.
* Less messing around with constructors, we just pass the required native API structs.
* Performance change should be nil.

### Drawbacks:
* This is very much a partial implementation of the functions in question, only pulling in precisely what we're using at the minute.
* Much less testing coverage. (This is relatively simple matrix math, so unless I've stuffed things up we shouldn't need testing.....)
* Increases the API size for a potentially questionable benefit: It's likely that we could pull these in from another library should we choose to go in that direction.
* A couple of hacks to account for existing API behaviour.
* Nil performance change (?)

### DIFF
--- a/source/ObjectViewer/ObjectViewer.csproj
+++ b/source/ObjectViewer/ObjectViewer.csproj
@@ -39,6 +39,7 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <DebugSymbols>false</DebugSymbols>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
     <CheckForOverflowUnderflow>False</CheckForOverflowUnderflow>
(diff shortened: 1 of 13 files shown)


leezer3:
Another use (sooner or later....) if we complete the Matrix transformations will be in the animations for X and S files.

These both use transformation matricies to produce keyframe animations, and when #345 is merged, I'd like to avoid the resultant plugins having to add a reference to openTK for these. (This increases the reference count, and introduces the possibility for trouble if someone uses the new and likely incompatible upcoming openTK version)

leezer3:
Merging the shared renderer means this is a tad broken.
On the flip side, this is better for inclusion when the renderer lib is finished, as this would mean that we only require unsafe code in this lib, and could keep the main sim entirely in managed code.

leezer3:
Now that things are getting much closer to having the entire shared render library working, this needs revisiting.

Probably best to pull out the Matrix4 code and merge this.

Can then worry about whether we want to actually use it at a later date, although a selectively completed set of functions is never going to be fun- Sooner or later, we'll need another one, at which point that'll need implementing, or it could well just be easier to stick with known reference code.....

leezer3:
Merged as part of #399

PR #353, opened 2019-05-28
Shared Renderer, Part 4

leezer3:
This moves the following to the shared renderer:
* Face rendering function
* Some reset / init functions


leezer3:
N.B. This only moves one of two RenderFace functions (That which renders a face with a specified material)
This is because the other makes calls to the ObjectManager.

Leave for the moment.....

leezer3:
The PRs on this so far loose us ~1,600 lines of shared code.

PR #355, opened 2019-05-30
Shared Renderer, Part 5

leezer3:
This moves the following to the shared renderer library:
* Lighting
* Texture manager


leezer3:
This should also make all texture load calls run through the API interfaces.

N.B. One method remains in the 'old' texture manager at the minute, but that's dead simple to remove (Probably just pass the parameters to it, as it's not being used too often)

CHECK:
Some of the route parser texture loads can be simplified.

leezer3:
Squash and merge this one again once CI completes.

PR #357, opened 2019-06-28
Add: Motor sound editor

s520:
This PR adds MotorSoundEditor.
Along with that, I made various corrections.
- Change language file to XLIFF format. (Because the traditional CFG format does not assume translation of multiple programs.)
- Fixed the problem that MP3 can not play in non-Windows environment. (By default, NAudio uses Media Foundation, a Windows API.)
- Supports RIFF / WAVE 24-bit integer, 32-bit integer and 32-bit float playback.
- Supports playback of Ogg Vorbis format.
- Move SoundBuffer class and SoundSource class to SoundManager.dll.

Although there is a motor sound editing function in TrainEditor, its editing function is poor. In addition, since there is no playback function, the author has to start and check the main unit each time. This is very inefficient.
MotorSoundEditor solves these problems.


s520:
- Basic operation
basic_operation

- Sound source index setting
sound_source_index

- Scaling
scaling

- Playback
playback

leezer3:
In general, very nice!

I might not have time to test this properly for a few days, but still :)
Some minor thoughts to start with:
* Would it be better to combine the Vorbis and MP3 plugins into a single NLayer plugin which supports both?
* https://github.com/leezer3/OpenBVE/pull/357/files#diff-1a0f2ced78eaf05644f0c62ba4336619R24 The final clause here is only relevant to 'official' FreeBSD stuff, and is probably unnecessary :)
* XLIFF is neat, but much harder to edit for non-technical people. I think Transifex will do them, but decent desktop editors are much thinner on the ground; What are you using in this regard?
* As an alternative to XLIFF, we could just add a new section to the language files for each viewer. Not ideal I'll admit.
* There's rather a lot of 'noise' in this from line spacing changes. Not a lot we can do here I don't think, but makes the diff a lot harder to read.
* The inverse model update function seems to have vanished. I don't think this has been used since well before Michelle left, so we could probably drop the thing entirely.
* https://github.com/s520/OpenBVE/blob/MotorSoundEditor/source/SoundManager/Sounds.Convert.cs#L12 This is probably better as a function of the Sound itself in the API.

s520:
First of all, the format of the code follows [".editorconfig"](https://github.com/leezer3/OpenBVE/blob/master/.editorconfig). Basically I think it follows the format of your recent code. Also, the current code has a mix of tabs, spaces, and so on, and it looks bad. Although diffs are difficult to read, I hope that you will help to unify the formatting of the code.

Next I think that it is better to separate Vorbis plug-in and MP3 plug-in as it is. It also makes it easier to isolate problems.

Also, XLIFF is supported by [Transifex](https://docs.transifex.com/formats/xliff) and [Poedit](https://poedit.net/).

Well, is inverse model not used? The default setting of OpenBVE is inverse model, not linear model. (https://github.com/leezer3/OpenBVE/blob/master/source/OpenBVE/System/Options.cs#L212)

Finally, I will fix the FreeBSD license statement later.

leezer3:
I meant linear, sorry :)
Either way, whilst it's an option, there's no actual way to set it other than by editing the options file, and it's un-documented at best. 

Formatting is a tricky subject-
I've considered a wholesale reformat plenty of times, but that breaks any existing branches and PRs.  The codebase is as I'm sure you're aware a total mess in plenty of other ways too!
The way I've basically ended up doing it is to apply the 'new' formatting to any new / moved / modified code, but to leave existing files alone where possible- This maintains the best readability of the diff.

s520:
I am currently using the VS 2019 reformat function.
Can this function only reformat the new code part?
https://docs.microsoft.com/en-US/visualstudio/ide/reference/options-text-editor-csharp-formatting?view=vs-2019

leezer3:
Whilst I'm only on VS2017, I *think* this only works on complete files.

It's not the end of the world, just makes things harder to read (This diff is ~100,000 lines, of which ~70,000 are the languages, and I'd guess 2/3 of the rest are just whitespace / newline changes)

s520:
Are you coding with the reformat feature disabled?
From now on, I would like to match your way as much as possible.

leezer3:
It's a bit of all-sorts TBQH.

I've turned off Visual Studio's code formatting, and only use that provided by Resharper.
I also have a (bad?) habit of only selectively committing- With GitKraken, it lets me stage on a line-by-line basis, and so anything that's just whitespace tends to not get committed.

This is great for keeping the diff sizes down, but it leaves us in exactly this boat, where the code formatting is an inconsistent mess, and a PR from someone else can give a massive diff where the actual code change is minimal.

Wouldn't worry about it to be honest, anything important can always be rebased manually if necessary.

s520:
Thank you so much for explaining.
I totally forgot, but I used Format document on Save to automatically reformat the entire file when I saved it.
In the future, we will disable this extension and keep it partially reformatted.

s520:
I fixed the coding format.
Please check it.

leezer3:
Thanks, I've merged.

Done a small bit of moving around so that SoundSource , SoundBuffer and the SoundModel enums are base types of the SoundManager, as this makes it so much neater to refer to them :)

s520:
Thank you.
I am now improving the editor to also edit sound.cfg / .xml.
Also, my colleagues are working on the editor's documentation.

PR #360, opened 2019-07-04
Change: Use the AbstractCar to play all sounds

leezer3:
This PR is intended to remove the need to supply both a reference to the train and a car index when playing a sound.

Instead, we should just pass the abstract car type.

### Benefits:
* Simpler code
* If we ever get as far as being able to detach a car from a train, this will make that much easier, as the sound parent object is now the car itself.

### Hazards:
* It's somewhat harder to stop all the sounds for a train; We can no-longer just pass the abstract train to the sound manager. This has been worked around by moving this method to the abstract train, and using linq in the derived version to check whether the sound parent is contained in the car array.
* With a change this major, there's a decent chance something has broken. I've done some basic testing so far (it builds, basic sounds seem OK)

@s520 -
Thoughts?


source/RouteViewer/Audio/Sounds.Update.cs
```diff
@@ -488,36 +497,136 @@ protected override void UpdateInverseModel(double timeElapsed)
...
 						AL.SourcePlay(source.OpenAlSourceName);
 						source.State = SoundSourceState.Playing;
 					}
 				}
 			}
+
+			RecAndPlay(listenerPosition, false, clampFactor);
+		}
+
+		private void RecAndPlay(Vector3 listenerPosition, bool IsLinear, double clampFactor) {
```
leezer3: Wants deleting again, got into Route Viewer accidentally....

source/OpenBveApi/Train/AbstractTrain.cs
```diff
@@ -39,5 +39,11 @@ public virtual void Derail(int CarIndex, double ElapsedTime)
 		public virtual void Dispose()
 		{
 		}
+
+		/// <summary>Stops all sounds from this train</summary>
+		public virtual void StopAllSounds()
+		{
+
+		}
 	}
```
s520:
```csharp
public AbstractCar[] Cars;
```
It may be better to move Cars from the Train class of each program.
Then you can eliminate the override of the StopAllSounds function.

source/RouteViewer/Audio/Sounds.Update.cs
```diff
@@ -23,7 +24,14 @@ protected override void UpdateLinearModel(double timeElapsed)
 			 * */
 			Vector3 listenerPosition = Camera.AbsolutePosition;
 			Orientation3 listenerOrientation = new Orientation3(Camera.AbsoluteSide, Camera.AbsoluteUp, Camera.AbsoluteDirection);
-			Vector3 listenerVelocity = Camera.AlignmentSpeed.Position;
+			Vector3 listenerVelocity;
```
s520: This change should be undone.
I think RouteManager TrainManager.PlayerTrain is null.

source/SoundManager/Sounds.cs
```diff
@@ -364,27 +364,26 @@ public SoundSource PlaySound(SoundBuffer buffer, double pitch, double volume, Op
...
 		/// <param name="volume">The volume change factor.</param>
-		/// <param name="position">The position. If a train and car are specified, the position is relative to the car, otherwise absolute.</param>
-		/// <param name="train">The train the sound is attached to, or a null reference.</param>
-		/// <param name="car">The car in the train the sound is attached to.</param>
+		/// <param name="position">The position. If a train car is specified, the position is relative to the car, otherwise absolute.</param>
+		/// <param name="car">The parent object the sound is attached to, or a null reference.</param>
 		/// <param name="looped">Whether to play the sound in a loop.</param>
 		/// <returns>The sound source.</returns>
-		public SoundSource PlaySound(SoundBuffer buffer, double pitch, double volume, OpenBveApi.Math.Vector3 position, object train, int car, bool looped)
+		public SoundSource PlaySound(SoundBuffer buffer, double pitch, double volume, OpenBveApi.Math.Vector3 position, AbstractCar car, bool looped)
```
s520:
```csharp
public SoundSource PlaySound (SoundBuffer buffer, double pitch, double volume, OpenBveApi.Math.Vector3 position, Abstract Car car, bool looped)
```
I think that this overload is unnecessary.

```csharp
public SoundSource PlaySound(SoundBuffer buffer, double pitch, double volume, OpenBveApi.Math.Vector3 position, object parent, bool looped)
```
You can make it common with this function.

leezer3:
The trouble with moving cars out from the train class is that when we're accessing them, we want the derived type not the generic one.
Using an array of abstract cars in the abstract train at the minute would need a lot of extra casts...

It would be possible to move the entire car class into the API & do it that way, but that somewhat defeats the point of abstracting it.

Will do some more on this tomorrow.

source/RouteViewer/Audio/Sounds.Update.cs
leezer3: Will double check, but if I recall it's not actually called.

Leaving it there is ?possibly? marginally better, as it means the code is common between Route Viewer and the main sim.

s520:
That's right for you.

If so, we will need a cast at the time of access.

Please dismiss this...

source/RouteViewer/Audio/Sounds.Update.cs
s520: Correctly TrainManager.PlayerTrain.Cars is null.
Try launching RouteViewer.
System.NullReferenceException occurs.

leezer3: Almost right- It did crash, but not with this one....

I've one two things to fix this, which is I think a little better-
* Added a check for whether we have a loaded route before running the update functions for sounds etc.
* Changed the current camera view to be track based, not interior.

source/SoundManager/Sounds.cs
leezer3: Absolutely, thanks :)

source/RouteViewer/Audio/Sounds.Update.cs
s520: Thanks! I've confirmed.

leezer3:
Merged into master, although the PR seems to have got a little confused with some local changes....

PR #361, opened 2019-07-05
Player Train: Access via method

leezer3:
This is another PR aimed at simplification and reducing interdependance.

At present, the PlayerTrain is checked and accessed via the TrainManager.PlayerTrain variable.
Often, we only need to know if a train *is* the player train in a function, and have no need for the rest of the TrainManager, normally because we're working with an abstract train in a separate assembly.


leezer3:
A property looked the better solution, but very similar!

PR #362, opened 2019-07-06
Update: SoundEditor (Close #359 )

### LINKED ISSUE #359: Questions about the motor sound section of sound.xml
I think that the current sound.xml is as follows, according to [this](https://github.com/leezer3/OpenBVE/blob/master/source/OpenBVE/Parsers/SoundConfiguration/SoundCfg.Xml.cs#L544).

```xml
  <CarSounds>
    <Run>
      <Sound>
        <Index>0</Index>
        <FileName>Sound\run0.wav</FileName>
      </Sound>
    </Run>
    <Motor>
      <0>
        <FileName>Sound\motor0.wav</FileName>
      </0>
    </Motor>
  </CarSounds>
</openBVE>
```

There is no problem with the Run section, but because the Motor section contains numbers in the element name, it is invalid XML.

However, [this](https://github.com/leezer3/OpenBVE/blob/master/source/CarXMLConvertor/Convert.SoundCfg.cs#L96) is normal XML because it is written in the same way as the Run section.

May I modify the parser to match the CarXMLConvertor writing method?

s520:
This PR adds the editing function of sound.cfg/.xml to MotorSoundEditor.
It was renamed the SoundEditor accordingly.


leezer3:
It's looking very promising, here are a few first thoughts:
(Feel free to disagree!)

1. I think we need the train folder displayed /editable somewhere. Either at the top or the bottom probably. This would then allow the Input value to be purely relative, and not have the current save logic.
2. With the above in mind, I wonder about disabling editing entirely until either a new file has been created or an existing opened?
3. As we can load the sounds, it'd be good to validate them as being a playable sound before adding them to an index. (At the minute, it'll quite happily add anything)
4. English typo: Iput value
5. Add a value *might* be better as a popup? This would mean though that the 
6. Cosmetic: Probably better to anchor the groupboxes on the left to the top rather than the bottom.
7. Crashes with ATS sound indexes above 100 (256 is the max supported)

leezer3:
Another thought-
Would it be better to merge the train editor with this too?

That would give one app to edit an entire train.

s520:
Thank you for checking.

First, I'm planning to change the regular storage format to an intermediate file format (.xml). At this time, the file path is recorded as an absolute path. Also, export Train.dat and Sound.cfg or Train.xml and Sound.xml from the intermediate file format.

Given the above ...

1. The file path is recorded as an absolute path, so the display may be unnecessary. When it is displayed, it will be displayed in the title bar.
2. Since file paths are not saved as relative paths, there will always be no problem with being editable. It asks the save location before exporting, so the program can create a relative path from that path.
3. That is a good idea. For example, it would be nice to display a check mark or cross next to the file path text box.
4. I will fix it.
5. In the part of sound setting, right?
6. I intentionally locked it to the bottom because I thought it was easy to use when maximizing the window.
7. Oops, I will fix it.

I think it is a good idea to put together in one app.
Currently, editing such as extentions.cfg can not but rely on a text editor. You mean creating a new TrainEditor that allows you to edit all configuration files related to trains.

s520:
Also, in the future, we need to create an editor in order to recommend XML format configuration files, right?
It may be good to make it this opportunity ...

leezer3:
Don't know :)

My instinct is to lock the sidebar to the top, as this means that when changing the window size, they always remain in the same relative place. It's debatable though.

With the popup, that was my thought, yes.
I don't know how well it'd work with editing existing values though. The current editing works nicely with these.
I suppose we could just have both (If you defined the relevant groupbox solely in code as opposed to via the designer, it could even use a common one, although that might get messy), but it's trivial, and may end up confusing matters.

leezer3:
> 
> 
> Also, in the future, we need to create an editor in order to recommend XML format configuration files, right?
> It may be good to make it this opportunity ...

That was my thought with merging the train editor into one.
Essentially speaking the most of the tabcontrol & logic in the current train editor could be copied wholesale into another tab in the new app. All that really needs doing is to delete the sound tab from it.

We then have the following process (or something like):
Open an existing train, or create new. Until this is done, all tabs etc. disabled.
With this, store the root path (e.g. to train.dat / train.xml)
All future paths are now relative to this base folder.
For example, if adding a new sound, first validate it and then copy into the new folder.
Phyisical properties etc. of the train can be edited in the first tab

s520:
There is no motor section or acceleration section in the current train.xml, right?
Let's formulate the specifications.

leezer3:
> It seems that train.xml defines the brake section by Car, but should it be put together in one section in Train? Is it safe to set different values for each car?

It should be safe to set different values per-car for the brake systems.

At present, each car is essentially an independent item, and there is no actual simulation of the brake pipe pressure affecting different cars:
https://github.com/leezer3/OpenBVE/blob/master/source/OpenBVE/Simulation/TrainManager/Train/BrakeSystem.cs#L24


> There is no motor section or acceleration section in the current train.xml, right?
> Let's formulate the specifications.

The issue above is somewhat why I haven't got as far as attempting to work out a specification for acceleration etc. yet;
At present, the train.dat refers to the train as a whole, and stuff like acceleration is averaged across all motor cars. I'd need to revisit and check exactly how this is handled, but I think that each motor car contains a copy of the entire curve and does stuff with it.....

Ideally, all properties would be able to be set on a per-car basis. I *think* this will largely work at the moment, but we may well find that the physics system makes assumptions somewhere that break things.
Most things can actually be copied wholesale from the train.dat into an identical per-car property.

My current thought-process is something along these lines:
* Create new derived car types for **TrailerCar** and **MotorCar** , each only containing the relevant properties / methods.

s520:
I understand. I will be able to edit each item by car in the editor.
However, I think that the safety equipment etc. will be set to the top car rather than set for each car.

leezer3:
That seems reasonable for the minute.

An possible interesting alternative would be to set the safety equipment as that installed on the driver car.
At the minute though, I think that's adding far more complication than it's worth.....

s520:
I am working on the development of TrainEditor2 (provisional name).
This editor supports importing and exporting Train.dat, Extensions.cfg, Panel2.cfg, Panel.xml, Sound.cfg, Sound.xml.

TrainEditor2

s520:
I will close this PR once.

PR #363, opened 2019-07-09
Refactor TrackFollower to take an absolute or a relative position

leezer3:
Marginal bit of refactoring work in the TrackFollower class.

This makes it so that it will take either an absolute track position (Moves the follower to said position, was the previous behaviour), or a relative position. 

### Benefits:
* A lot less passing around of relative positions.
* Moves the static methods used to update the track follower in Route Viewer into methods pertaining to the follower as per the main program.

### Drawbacks:
* Only relative in the main program.
* A bunch of stuff probably should use this (e.g. some of the camera based stuff), but doesn't for legacy reasons.

### TODO:
* Test and check that nothing has gone awry.
* Make RouteViewer use the multiple track logic from https://github.com/leezer3/OpenBVE/pull/326

### DIFF
--- a/source/OpenBVE/Game/AI/AI.SimpleHuman.cs
+++ b/source/OpenBVE/Game/AI/AI.SimpleHuman.cs
@@ -376,15 +376,15 @@ namespace OpenBve
 						// next station stop
 						int te = Train.Cars[0].FrontAxle.Follower.LastTrackElement;
 						int currentTrack = Train.Cars[0].FrontAxle.Follower.TrackIndex;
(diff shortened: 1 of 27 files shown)


leezer3:
Rebased this onto the latest master with associated fixing.

Will do a little more testing, but the Route Viewer changes could do with being merged ASAP, as this gets us somewhat closer to being able to shift the Tracks array to the RouteManager.
Precursor for https://github.com/leezer3/OpenBVE/pull/371 as we need TrackFollowers in there in order to get the full animated structure into the API.

PR #364, opened 2019-07-19
Change: Bump required Mono version

leezer3:
Stretch is now Debian OldStable, and this gives us a much newer Mono version to play with!

This *should* be seamless hopefully.
Whilst we don't actually require anything newer than 3.2.8 at the minute, this should let us use a few newer features if we need.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ matrix:
 #Nightly builds for upload. Use lowest possible Mono version for compatibility purposes
     - os: linux
       dist: trusty
-      mono: 3.2.8
+      mono: 4.6.2
       script: make publish
     - os: osx
       mono: latest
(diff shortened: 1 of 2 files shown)


s520:
I think it's good.
In 3.2.8, if you [manage resource files in Visual Studio](https://docs.microsoft.com/en-us/dotnet/framework/resources/creating-resource-files-for-desktop-apps#VSResFiles), build errors will occur...
Thus 3.2.8 seems to have some problems.

leezer3:
Thanks :+1: 
I completely agree that 3.2.8 is rather elderly and has issues, but the rationale for keeping it at that was to track Debian OldStable (Debian Oldstable represents the absolute minimum version of *anything* likely to be supported simply because of their philosophy on software versioning)

s520:
Yes. I already understand.
When do you actually bump up mono?
I think it is better to do this before getting the Debian package certified.
(Even if it says, it's still the future...)

leezer3:
The short answer would be when the Debian OldStable Mono version changes, although that depends on me remembering to look!

I don't anticipate the Mono version changing again until at the earliest next year, which is when Buster is pencilled in to become OldStable.

PR #366, opened 2019-07-29
Sound Refactor

leezer3:
This moves a few sounds and related properties into the enclosing types.

The ultimate aim will be to eliminate the Train.Specs, Car.Specs & Car.Sounds structures, so that all are found within the appropriate enclosing devices etc.

**Okies, so as usual, this has got a little larger than the original first post spec. Various additions detailed below:**
* New sound **emergencyRelease**
* New panel / animated variable **pilotLamp**
* New panel / animated variable **passAlarm**
* New panel / animated variable **stationAdjustAlarm**


source/OpenBVE/Game/Events/EventTypes/Station.cs
```diff
@@ -145,9 +133,8 @@ public override void Trigger(int Direction, EventTriggerType TriggerType, TrainM
 							{
 								Train.StationState = TrainStopState.Pending;
 							}
-							
-							int d = Train.DriverCar;
-							Program.Sounds.StopSound(Train.Cars[d].Sounds.Halt.Source);
+
+							Program.Sounds.StopSound(Train.SafetySystems.PassAlarm.Sound);
```
leezer3: Needs a refactor, so that we call a halt method on the pass alarm instead.

Will then allow future abstraction of the pass alarm for more interesting implementations.

source/OpenBVE/Simulation/TrainManager/Train/PassAlarm.cs
```diff
@@ -1,8 +1,50 @@
-﻿namespace OpenBve
+﻿using SoundManager;
+
+namespace OpenBve
 {
 	/// <summary>The TrainManager is the root class containing functions to load and manage trains within the simulation world.</summary>
 	public static partial class TrainManager
 	{
+		internal struct PassAlarm
```
leezer3: Need to comment this file properly.

source/OpenBVE/Simulation/TrainManager/Train/SafetySystems.cs
```diff
@@ -0,0 +1,12 @@
+﻿using OpenBveApi.Runtime;
+
+namespace OpenBve
+{
+	internal struct SafetySystems
+	{
+		/// <summary>The PassAlarm</summary>
+		internal TrainManager.PassAlarm PassAlarm;
+
+		internal DoorInterlockStates DoorInterlockState;
```
leezer3: May want to end up as a separate safety system?
Possible out of scope for this PR

leezer3:
Rebased.

Also added a further commit, which allows us to whitebox the air-compressor entirely.

leezer3:
New sound added to both Sound.xml and sound.cfg:
**emergencyRelease** (in the [Brake] section)

This allows a dedicated sound to be played when the EB is released, instead of the previous behaviour which was to play the brake release sound.
If no sound is specified, the brake release sound will be played for backwards compatibility reasons.

leezer3:
Another minor addition:
We already have sound effects for the PilotLamp, so it was logical to hook into that with a dedicated panel / animated variable.
**pilotLamp** - Returns 1.0 if the pilot lamp is lit (e.g. all doors closed and ready to depart) , 0.0 otherwise.

leezer3:
Similar vein- We need a panel / animated variable for when the pass alarm triggers. Will add tomorrow or something.....

leezer3:
Another marginal thought- Should the pilotLamp variable be renamed before this gets merged?

It's somewhat unfortunate that pilotlamp is the name of an on-off animation in the panel formats.....
Could add an overload to the panel code to use just **lamp** , but that seems somewhat pointless....

leezer3:
**passAlarm** - Returns 1.0 if the pass alarm is active, 0.0 otherwise

leezer3:
New panel / animated variable:
**stationAdjustAlarm** - Returns 1.0 if the station adjust alarm is active, 0.0 otherwise

leezer3:
Next step is I think to move the alarms into a separate namespace (OpenBve.SafetySystems probably)

Really need to start on the separate TrainManager dll too....

leezer3:
Namespace moved.

Need to do a little testing, and then I think merge these into mainline again.

leezer3:
Travis failure unrelated, but this is a tad broken, need to fiddle with the order things are constructed in a bit.

PR #367, opened 2019-08-02
 Change: use NuGet packages

s520:
This PR changes the dependent library to get from NuGet package.
In addition, make the following changes.

- Change the version of .NET Framework from 4.0 to 4.6.1
- Stop using mcs in makefiles and use xbuild or msbuild instead.

First, I changed the target framework to 4.6.1 because most of the recent useful libraries use .NET Framework 4.6.1 or .NET Standard 2.0.
Second, if you use the NuGet package, it is difficult to run on mcs. In the past, you said that xbuild and msbuild are incompatible, but xbuild wraps mcs and msbuild wraps csc, so using mcs should cause compatibility problems. Fortunately, the minimum version of Mono is 4.x, so compatibility issues can be ignored.


s520:
Just in case, even if you change the version of OpenBveApi's .NET Framework, there is no need to rebuild the train plug-in.

leezer3:
All of those are reasonable except openTK.
Need to test DotNetZip a little, but don't see a real problem there.

It's missing a bunch of patches from the last year.
These are primarily Mac and Windows gamepad related input fixes, which we want to keep.

I've opened an issue on openTK here as to whether they're likely to bump the Nuget package:
https://github.com/opentk/opentk/issues/940

s520:
Well, OpenTK only uses the binary you built.
troubled...

But, I was relieved to hear that there was no problem with others.

leezer3:
The openTK situation is somewhat of a mess at the minute.


Exactly what's happening as far as patches for the legacy 3.0 version is somewhat up in the air- At one point, there was the intention to provide no patches at all, but this has relaxed a little since then, and they've been accepting bugfix PRs, but no features.
Unfortunately, at the minute these patches haven't made their way to the the Nuget package.

The maintainers changed about 2 years ago, and the decision was made to completely re-write the entire library and windowing system. 

This is still ongoing, and to be honest, I really wouldn't want to try and shift a project of this size over any time soon.

s520:
OK. Depending on the response from OpenTK, we will return it to our own binary.

This change is also part of the preparation for getting out of Windows Form.
I suggest Xamarin.Forms as a good candidate for migration.
It supports Windows and Unix with WPF and GTK#.

s520:
## Issue
https://github.com/dotnet/standard/issues/514
https://devblogs.microsoft.com/dotnet/net-framework-4-7-1-runtime-and-compiler-features/

> The versions listed here represent the rules that NuGet uses to determine whether a given .NET Standard library is applicable. While NuGet considers .NET Framework 4.6.1 as supporting .NET Standard 1.5 through 2.0, there are several issues with consuming .NET Standard libraries that were built for those versions from .NET Framework 4.6.1 projects. For .NET Framework projects that need to use such libraries, we recommend that you upgrade the project to target .NET Framework 4.7.2 or higher.

https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-implementation-support

There seems to be a problem with the .NET Framework 4.6.2-4.7.1.
This is due to 4.6.2 not fully supporting .NET Standard 2.0.

We have the following options:
1. Leave 4.0.
2. Change to 4.5.
3. Change to 4.7.2.

Let's look at each one.
### First choice:
#### Advantage
Nothing has ever changed.
#### Disadvantage
The latest useful library can not be used.

### Second choice:
#### Advantage
Some recent libraries are available.
#### Disadvantage
Libraries other than some can not be used.

### Third choice:
#### Advantage
Most libraries are fully usable.
#### Disadvantage
The default Debian "stretch" can not use the corresponding Mono.


**I recommend the third choice.**
Currently, our minimum version of Mono depends on the previous Debian release.
Unfortunately, at present, OpenBVE has been removed from the Debian repository. 
From now on, it will be the earliest to return to Debian from "bullseye". 
So what we should target is "buster". 
The Mono version of "buster" is 5.18, so it supports 4.7.2. (https://packages.debian.org/ja/bullseye/mono-complete)
Fortunately, the Mono project has released the latest stable version for "stretch" in the repository. 
Even now, Mono needs to be introduced separately, so the complexity remains the same.

leezer3:
Don't know, sorry.....
Unfortunately, we don't have any clear stats on the actual operating systems and Mono / .Net versions in use, and this means that it's somewhat hard to make a decision based upon anything more than educated guesswork.

My personal instinct probably is to keep tracking oldstable and bump the .Net version to 4.6.1, and revisit if there's something specific that 4.7 is required for.

s520:
I faced the same problem as the following site.
https://elleneast.com/?p=4302 (Japanese site. I'm sorry.)

This site shows the problem that many .dll files are output to the build output path.
This issue is fixed in 4.7.2.

leezer3:
Interesting.

That only appears to apply if we build with 4.6.1 *and* reference a DLL built against Netstandard.

If we keep the local dependancies, or only reference ones which aren't built with Netstandard targeted, this issue goes away.
I think one of the packages you're pulling in must be the Netstandard version. Let me test at this end and see what I can do.

openTK have also bumped for us:
https://www.nuget.org/packages/OpenTK/3.1.0

s520:
You're right.
I use [ReactiveProperty ](https://github.com/runceel/ReactiveProperty)to create TrainEditor2.
This library is targeted at .NET Standard 2.0.

leezer3:
Unfortunately, I don't think we have a good option here. 
Looking at the Mono version shipped by other major distributions:
* Ubuntu LTS (Xenial) has Mono 4.2.1
* OpenSUSE (15.0) has Mono 5.10.1
* Arch / Manjaro has Mono 5.20.1
* Fedora has Mono 4.8

My strong feeling is that we should keep the Mono version as low as possible. 
Mono 5.18, which AFAIK is the minimum for .Net 4.7.2 only hit 8 months ago, and it definitely hasn't had time to reach all the distributions, other than cutting edge stuff.

Possible workaround thought-
Could we compile ReactiveProperty ourselves without the .NetStandard target?

s520:
If you say that far, I will stay at 4.6.1.
Instead, we will ignore this problem.
There is no problem in operation.

leezer3:
I think for the minute we should probably get this PR finished and merged- If I understand the situation correctly, we won't be affected by this until TrainEditor2 is complete & ready to be merged.

Operation wise, I agree it (probably) won't be a problem, but if we start redistributing all of these extra DLLs, that's going to add to the download size which isn't a good thing.
I'd also be very cautious about what happens when a built copy is tried on Mono.
Either way, things will need very careful testing and consideration.

s520:
Thank you.
I will correct this PR a little.
Specifically, I am upgrading OpenTK.

s520:
Fixed.
Please merge.

Issue #369, opened 2019-08-05
System.Xml.Linq missing


Traumflug:
### Description

With any route and train tried so far, the application crashes at 61% of the splash screen.

### Reproduction

- Have Ubuntu 19.04.
- Downloaded Debian package 1.6.0.0 from the download page, opened it with the software installer (which takes care of installing dependencies) and installed it.
- Also installed packages bve-train-br-class-323 and bve-train-br-class-323-3dcab to get a train.
- Started the thing on the command line. Main window appears.
- Installed route package "Mühltalbahn" using the built in package manager.
- Selected the one available train and one of the routes.
- Clicked Start.

## Route

http://www.bvecornwall.co.uk/wordpress/the-muhlkreisbahn/

## Train

https://packages.ubuntu.com/disco/bve-train-br-class-323
https://packages.ubuntu.com/disco/bve-train-br-class-323-3dcab

## Logs
```
05.08.2019 12:35:03
OpenBVE 1.6.0.0 Crash Log
Program is running on the Linux backend
Current screen resolution is: Windowed 960px X 600px 
Current routefile is: /home/mah/.config/OpenBve/LegacyContent/Railway/Route/MKRB/Eilzug_OEBB5047.rw
Current train is: /usr/share/games/bve/Train/BR_Class_323
The current routefile caused the following exception: 
System.IO.FileNotFoundException: Could not load file or assembly 'System.Xml.Linq, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' or one of its dependencies.
File name: 'System.Xml.Linq, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'
  at OpenBve.Loading.LoadEverythingThreaded () [0x00ccf] in <d5078fc0703c4355b6dd6355b9e6eefe>:0 
  at OpenBve.Loading.LoadThreaded () [0x00000] in <d5078fc0703c4355b6dd6355b9e6eefe>:0   at System.Environment.get_StackTrace () [0x00000] in <d0e12f672b88444ab4b6d9b2ecf20142>:0 
  at OpenBve.Loading.LoadThreaded () [0x00000] in <d5078fc0703c4355b6dd6355b9e6eefe>:0 
  at System.Threading.ThreadHelper.ThreadStart_Context (System.Object state) [0x00000] in <d0e12f672b88444ab4b6d9b2ecf20142>:0 
[...]

PR #370, opened 2019-08-05
Fix: #369

s520:
libmono-system-xml-linq4.0-cil is missing from the Debian package dependency.

### DIFF
--- a/DebianControl.sh
+++ b/DebianControl.sh
@@ -36,7 +36,7 @@ Maintainer: leezer3 <leezer3@gmail.com>
 Architecture: all
 Version: $Version
 Provides: bve-engine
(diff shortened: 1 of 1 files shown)


leezer3:
Great, thanks :)

Probably ought to push 1.7.0 with this included, although I could just alter the current package.

Was hoping to leave until all the refactoring into separate libraries was done, although that's still a long way off.....

Thoughts?

s520:
I also hope that 1.7.0 will be released after TrainingEditor2 is completed, but this problem is serious.
Is it better to disseminate this problem on the homepage?

It is subtle...

leezer3:
Fixed version (Debian package only) now pushed, along with a front page note.

That will hopefully do for the minute :)

PR #371, opened 2019-08-09
Move the Animated Object to the API

leezer3:
This is some ongoing work to move the Animated Object related functionality into the API, so that these can be moved to a common set of plugins.

This depends on the TrackFollower moving to the API (for TrackFollowerFunction), which in turn depends on the Events moving to the API.....

Whilst this is mergable at the minute, & eliminates *some* duplicated code, probably wants to hold off until the Events related stuff is merged.


leezer3:
Quite a bit of heavy lifting done in here.

* The Camera has been converted to a singleton class- This allows us to pass it back to the API via a dynamic. (Whilst doing this isn't clever, it's the only way to currently update overlay objects...)
* Animated object updates are now in the API. 

We're most of the way towards being able to move the entire object lump to the API, but need TrackFollowers there first, in order to get TrackFollowingObjects into play.

I don't think there are any bugs at the minute, but this is now doing a lot of clever tricks with references, which has the danger to go badly wrong.....

leezer3:
Track follower is now in the API:
https://github.com/leezer3/OpenBVE/commit/3b809b607303f2e3caa0731e70637a8295aa2c3f

Axles need to shift across next, but this may be a PITA with the CarSounds.
Possibly use a base / derived class?

leezer3:
TrackFollowingObjects are now in the API.

Next issue is the WorldSound, which uses both SoundBuffer and SoundSource:
https://github.com/leezer3/OpenBVE/blob/master/source/OpenBVE/Game/ObjectManager/AnimatedObjects/WorldSound.cs

Need to think about this one tomorrow....

leezer3:
It works :)

All the animated object related stuff is now in the API, and I'm semi-confident that I've eliminated the bugs......

@s520 -
Did you want to take a look at this before I even consider merging?
It uses a lot of fun and games with references, and I can't say I'm overwhelmed with the host interface changes either.

The plus side is that it'll let us combine the final object loaders into a single set of plugins.
It'll also make it much easier to untangle the mess when we finally get as far as rebuilding the internals of the animated object.

s520:
Please give me a little time.
Let's merge it into my working base and check...

s520:
I think there is no problem :)

PR #374, opened 2019-08-23
Upgrade MotorSoundEditor to TrainEditor2

### LINKED ISSUE #359: Questions about the motor sound section of sound.xml

s520:
This PR upgrades MotorSoundEditor to TrainEditor2.
I would like to share my current situation.

## Task list (Update: Sept 2, 2019)
- [x] TrainEditor functions
- [x] MotorSoundEditor functions
- [x] MotorSound simulation functions
- [x] PanelEditor functions
- [x] SoundEditor functions
- [x] Data input / output functions
- [x] UI adjustment

Close #359


s520:
@leezer3 -
What do you think?

leezer3:
No chance to test yet, but will try and do so ASAP :)

leezer3:
Looking very reasonable.

I needed this commit to actually launch though:
https://github.com/leezer3/OpenBVE/commit/a8058d80763581e4dea7b9d9476d114c2509aa0a
(my fault...)

s520:
Thanks!

What do you think about the problem of System.***.dll being copied in large numbers?
There seems to be no problem in operation, but ...

leezer3:
Don't have a particularly good answer :(
We could run a post-build event to dump them into a hidden folder & edit the .config file to look there, but that still means we have the extra download size.

Need to test a Linux built version on Windows too & vice-versa.

Almost makes me wonder how much work would be involved in trying to build a forked copy of ReactiveProperty against .Net 4.....

s520:
Umm...
Porting ReactiveProperty will not be impossible. (It's not a very good choice.)
Let's try it for a moment.

At least binaries built on Windows worked fine on Linux.

s520:
ReactiveProperty itself can be ported, but the related System.Reactive port should also be ported. Unfortunately, this is likely to require the porting of all relevant libraries...

leezer3:
System.Reactive was available pre 4.6 ( https://stackoverflow.com/questions/40004669/system-reactive-rx-net-4-0-on-nuget ) , but I agree it's not an especially good idea.

s520:
PanelEditor has been added.
Please check it.

The NuGet package is used to simplify maintenance.
Rebuilding the library source erases the above benefits.
So I disagree with rebuilding the library.

source/TrainEditor2/Extensions/Utilities.cs
```diff
@@ -8,6 +8,14 @@
 
 namespace TrainEditor2.Extensions
 {
+	internal enum NumberRange
```
leezer3: I'd be tempted to move the number parsing utilities into the API.

In the existing parsers, parsing and validating takes up quite a bit of code, and could definitely be simplified into global helper functions.
Not saying do that now, but as you've written these functions, it might be an idea to think about it so that they could be dropped in as changes are made?

s520: Thank you.
The latest code has been modified to the suggested code. (Not committed yet.)

leezer3:
Don't worry, I'm not especially serious about that point, rather ensuring we've explored all options :)


Current build of TrainEditor2:
* Most of the options in the file menu don't seem to be working.
* When loading using the default en-US language (from a fresh built copy), the left menu isn't translated until a new language is selected.
* Panel editor appears to work. A preview would be nice, but may be slightly harder work. Not sure....
* It can be a little confusing as to why a tab is disabled (for example the panel editor- Only works when Train --> General is selected. (I'd probably link this to the driver car. As a secondary thing, the XML format supports different panels / 3D cabs on a per-car basis. Possibly wants to be a car property?)

source/TrainEditor2/Extensions/Utilities.cs
s520: I had a misunderstanding.
If you move to the API, I will follow it.

s520:
- Do not mind the menu part.
Currently working.
- Let's check the translation.
- The panel preview is expected to be technically difficult.
I would like to make it a function to be implemented in the future.
- I think that it is possible to move the panel for each vehicle, but is train.xml planning to have the panel for each vehicle?

leezer3:
Correct, train.xml allows for a panel for each vehicle if you want this.
https://github.com/leezer3/OpenBVE/blob/master/source/OpenBVE/Parsers/Train/XML/TrainXmlParser.CarNode.cs#L262

Changing the camera car in exterior view (Num7 / Num1) will also change the interior car if a panel has been provided. 

This isn't perfect, as it doesn't take account of whether a car should be visible from the interior of another. (I really should sort this, but 2D panels don't really work with it.....)

s520:
I understand.
Let's fix it.
Perhaps I should change the class that holds the Panel class from the App class to the Car class.

s520:
In mono 4.6.2, the build fails and an exception occurs during operation as shown in the video below.

output

This issue is caused because ReactiveProperty requires an assembly that is not included in mono 4.6.2.
To solve this, the minimum version of mono must be 5.18 or higher.

Fortunately, mono 5.18 is compatible with .NET Framework 4.7.2, which solves the massive DLL problem at the same time.
https://www.mono-project.com/docs/about-mono/releases/5.18.0/

In addition, Debian "buster" includes mono 5.18.
https://packages.debian.org/buster/mono-complete

leezer3:
Getting there :)

Another thought-
At the minute, TrainEditor2 is the only thing that requires a later Mono version.
With this in mind, I suppose we could exclude it from the Mono 4.6.2 builds. (Or even distribute it separately, but I really don't like that thought)

Advantages of this approach-
* Keeps the backwards compatibility. 

Disadvantages-
* You'd need two downloads somehow. One for stretch / older Linux versions, and one for newer. Easiest way would probably be to create a new folder on the server.
* Confusion as to which should be downloaded from a user's perspective.
* Needs a separate build script maintaining.
* Users of the 'old' mono version don't get the new train editor. On the flip side though, I think the number of users of the editing tools is likely to be highly minimal....

To be honest, I'm still concerned that requiring Mono 5.18 is a bad move- It's just not in enough of the default repositories of current distributions for me to be happy.
The Mono provided repositiories are up-to-date and work around this, but the more hoops someone has to jump through to get this working, the less likely they are to bother....

s520:
How about upgrading to 4.7.2 for TrainEditor2 **only**?

That way, you can leave choices for your users.
However, Mono 5.18 is required for the build.

s520:
Ready to merge.
Please check.

leezer3:
https://travis-ci.community/t/unable-to-locate-package-referenceassemblies-pcl/2686
Mono 5.18 seems to be broken on Travis.....

Need to actually try testing this on Windows before doing much else and trying to think of any clever solutions, might need a few days.

s520:
There is another problem.
As shown below, NuGet cannot detect xbuild in 5.x.
https://github.com/NuGet/Home/issues/7202

As a result, the build fails.
This has been corrected by the following PR, but is not included in the stable version of NuGet.
https://github.com/NuGet/NuGet.Client/pull/2891

I suggest that you **temporarily** build with 6.0.0 as a solution.

I have already confirmed that there is no problem with 5.x.

leezer3:
https://github.com/mono/mono/issues/10179

This is one of the major reasons I really want to keep the Mono version as low as possible (And is the one I know about!).

A workaround for these appears to be here (essentially changing to methods for which the signature didn't change):
https://github.com/xp-runners/reference/issues/75#issuecomment-419212058

Need to apply that, then build on mono-latest and test on 4.6.2 :)

s520:
As mentioned earlier, the following issues are unavoidable in mono 4.x.
If you are testing, do it in 5.x.
https://github.com/leezer3/OpenBVE/pull/374#issuecomment-526842834

Also, for the Debian package, I think you should separate TrainEditor2 and request the OpenBVE package and mono 5.x.

leezer3:
I get that; This is a related but separate issue :)

What I'd at least like to get as a minimum is that if the *main* game is built with 5.x (and auto-uploaded), it'll still run on 4.6.2 as we don't use any features from 5.x at the moment.

When 4.6.2 is dropped from oldstable. (Presumably approximately next July), the hack can be revisited, although it's not doing any real harm.

At present, unless something has changed I'm not aware of, this won't happen.

Whilst TrainEditor2 in that case would still crash, it keeps the minimum version required for most components at the lowest possible level.

It's not an elegant solution at all, but still...

s520:
I agree. I think it can't be helped.
Will you do the package separation or will I do it?

leezer3:
Will try and sort merging this tomorrow I think- Got to go to work after fighting with Travis all morning to get the 5.20 Mono working ( https://travis-ci.community/t/mono-install-script-subversion-parsing-failing/4989 )

Improvements can always be dealt with at a later date once people have had the chance to play with it.

leezer3:
Merged.

Added one small additional message if you attempt to open a train.dat / train.xml file prompting to use the import dialog.

My next major target is to pull the TrainManager into a shared library, similar to the RouteManager.
This will then allow all the parsers to be changed to plugins similar to the object / texture / sound ones.
At this point hopefully, it'll eliminate a bunch more of the duplicated code, and can hopefully be cross-ported into TrainEditor2, so that it will be able to import *any* train.

PR #378, opened 2019-09-03
Change: Alter string split methods

leezer3:
Alters the string split & related methods to hopefully use something that is compatible when built using the latest Mono compiler on 4.6.2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ matrix:
 #Nightly builds for upload. Use lowest possible Mono version for compatibility purposes
     - os: linux
       dist: trusty
-      mono: 4.6.2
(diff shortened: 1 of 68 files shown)


leezer3:
Arrgh....
For some (undocumented!) reason, Travis requires the complete Mono release number when building against a release in the 5.* series, as opposed to falling back to the closest variant.

5.20.0 fails.
5.20.1 works ok (oddly, the Mono site gives 5.20.0 as the current release. This was found when digging in the Github releases)

If this now builds without needing a kick on Travis, will merge for the moment, subject to reversion once the build has been tested a little more extensively on Linux.

leezer3:
Bingo, build no longer fails :)

PR #380, opened 2019-09-05
Fix: Linux builds not uploading after Mono bump again...

s520:
I'm sorry. There seems to be an error.

### DIFF
--- a/UploadScript.sh
+++ b/UploadScript.sh
@@ -12,7 +12,7 @@ function wrap() {
   if [ "$TRAVIS_OS_NAME" = "linux" ] ;
     then
 	echo "Linux worker"
-	if [ "$mono_version" != "5.20.1" ] ;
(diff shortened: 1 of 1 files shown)


leezer3:
Will do a better fix once I'm home, ought to be ignoring the minor ver & build, else this'll break if they patch it.

s520:
Oops...
Thanks.
This PR will close.

leezer3:
Kicked so that it parses the Mono version into an array and tests the first two entries.
Should now be uploading again.

Re-visiting the build and upload scripts is something that's somewhere at the bottom of my pile of stuff I ought to do.... (Keep thinking about a CAKE build system and a bit more logging)
We could probably dump the SLN building workers too now the makefile builds the SLN.

s520:
Certainly uploaded.
Thanks!
I can now post an introduction to TrainEditor2 on the forum...

Issue #381, opened 2019-09-05
[Request] Couple section with model


ManuelMejias:
Articulated train part it's a problem to model with two pieces, a coupler model could be a solution to this.

Some simulators such as Train Simulator saga use a unique model piece to joint trains avoiding that "empty" result on a curve transition.

leezer3:
That sounds like a very logical addition.

I'll try and look into it at some stage soon-ish.

### Brainstorming thoughts:

- Bogie connects to two cars.
- We need the position, direction, up & side vectors for the bogie in order to move the animated object appropriately.
- For a car, this is calculated as follows:
```C#
// calculate positions and directions for section element update
Vector3 d = new Vector3(FrontAxle.Follower.WorldPosition - RearAxle.Follower.WorldPosition);
Vector3 u, s;
double t = d.NormSquared();
if (t != 0.0)
{
	t = 1.0 / Math.Sqrt(t);
	d *= t;
	u = new Vector3(Up);
	s.X = d.Z * u.Y - d.Y * u.Z;
	s.Y = d.X * u.Z - d.Z * u.X;
	s.Z = d.Y * u.X - d.X * u.Y;
}
else
{
	u = Vector3.Down;
	s = Vector3.Right;
}
Vector3 p = new Vector3(0.5 * (FrontAxle.Follower.WorldPosition + RearAxle.Follower.WorldPosition));
```
Assume therefore:
- Replace the front / rear axles with the appropriate car axles.
- Up vector can probably be averaged across both cars as opposed to from the car root?
- Show / hide as per bogies & car objects?

Issues:
- This would anchor on the car axle pivot point, not the end of the car ==> Is it worth calculating an offset point? Can get a directional vector from the two axle positions and use this. but effort.....
- Check how the coupler stretch / compress works. IIRC this is just a regulation of the distance between the car ends?
- Doesn't take account of coupler stretch / compress properly.
- No scale object function (see above, probably generally useful!)

leezer3:
Implemented- You can now add a **Object** to a coupler declaration.

Will try and get around to thinking about growing / shrinking objects some stage soon.

PR #386, opened 2019-09-07
New: Add coupler objects

leezer3:
This implements the idea from the following issue: https://github.com/leezer3/OpenBVE/issues/381

Usage is simple-
In an extensions.cfg file, add the following:
```
[Coupler0]
Distances = 0.5,0.5
Object = coupler.b3d
```
In a train.xml file:
```XML
<Car>Car0.xml</Car>
<Coupler>
<Distances>0.5,0.5</Distances>
<Object>Coupler.b3d</Object>
</Coupler>
```
From a first bash about, it appears to work very nicely.

### Flaws:
- No stretch / compression, just a moving object. (Object grow / shrink functions to be a new PR)

@s520 
Not added this to the new train editor at the minute, you're probably best to do this.
Some testing from someone else would be good too!


leezer3:
Another (related):
This clones a nasty lump of update code.

C# doesn't do multiple inheritance, but we might be able to put this into a shared interface, need to look a bit how it might work. At the moment though, this isn't *too* important.....

leezer3:
Minor changes, primarily designed to get the coupler to be a property of the car as opposed to a member of the train.

Need to test that this hasn't accidentally broken anything using couplers to pull bogies or anything of that nature....

Issue #388, opened 2019-09-10
No motor sounds when using sound.xml


PC28K:
### Description
no motor sound using sound.xml, normal for sound.cfg, no error in log.
run and other sounds are normal.
original train data was imported into trainediter2 to export sound.cfg to xml.
https://pcws.kr/file/2019.09.10-19.32.mp4

## Route
route is uchibo line for bve4 (bvets.net)

## Train
train is https://www.youtube.com/watch?v=tBweD9T2b3Y

## Logs
openBVE Log: 2019-09-10 오후 7:47:08

19:47:09  Initialising game window of size 960 x 600
19:47:09  Game window initialised successfully.
19:47:09  Loading route file: G:\ob\userdata\railway\Railway\Route\Uchibo\1157M.csv
19:47:09  G:\ob\userdata\railway\Railway : Railway folder found.
19:47:09  Route file format is: CSV
19:47:10  FileName G:\ob\userdata\railway\Railway\Object\PCheck.bmp not found in Track.Marker at line 307, column 1 in file G:\ob\userdata\railway\Railway\Route\Uchibo\1157M.csv
19:47:10  BeaconStructureIndex 5 references an object not loaded in Track.Beacon at line 345, column 1 in file G:\ob\userdata\railway\Railway\Route\Uchibo\1157M.csv
19:47:10  BeaconStructureIndex 5 references an object not loaded in Track.Beacon at line 428, column 1 in file G:\ob\userdata\railway\Railway\Route\Uchibo\1157M.csv
19:47:10  BeaconStructureIndex 5 references an object not loaded in Track.Beacon at line 1493, column 1 in file G:\ob\userdata\railway\Railway\Route\Uchibo\1157M.csv
19:47:10  RailIndex1 6 could be out of range in Track.Crack at line 1529, column 1 in file G:\ob\userdata\railway\Railway\Route\Uchibo\1157M.csv
19:47:10  BeaconStructureIndex 5 references an object not loaded in Track.Beacon at line 1560, column 1 in file G:\ob\userdata\railway\Railway\Route\Uchibo\1157M.csv
19:47:10  BeaconStructureIndex 5 references an object not loaded in Track.Beacon at line 1622, column 1 in file G:\ob\userdata\railway\Railway\Route\Uchibo\1157M.csv
[...]

PR #389, opened 2019-09-10
Fix: a crash in the motor sound simulation

s520:
This PR fix a crash in the motor sound simulation.

However, the numerical value of the scale of the Motor sound settings graph is no longer displayed.

This is because I decided to use OpenGL.

Can I use the main program or viewer text display mechanism?

I am worried about this.

What do you think?

Related to #388 .


leezer3:
I don't see any reason why the text rendering function couldn't be lifted relatively easily, all it probably wants is the shared renderer importing and the appropriate functions calling.
openGL is openGL at the end of the day.

Having said that, I'm not totally sold on the GLControl idea, as this adds another bunch of overheads, but sooner or later it comes down to what's easiest to maintain.

s520:
Thank you.
Let's incorporate it immediately.
In fact, I didn't realize that I used LibRender until a while ago.

s520:
Now you are ready to merge.
Please check.

leezer3:
I'll merge and do some testing later this week :)

PR #397, opened 2019-09-26
Improve FPS when using TFO

s520:
This PR improve FPS when using TFO.

## Technique
- Use VBO, IBO, and VAO.
- Since the model conversion of the vertex is a bottleneck, execute it with a shader.

## Known issues
- ~Lighting is different from the conventional one.~
- ~Z-Sort of 2D cab does not work and is not transparented correctly.~
- ~The overlay is transparent.~
- ~Differences in cab lighting.~
- Transparencies are still a little off though.

### Verification environment
CPU: Intel Core i5-4200U
GPU: Intel HD Graphics 4400
RAM: 4 GB
Route: Includes [E6](http://midnightexpginga81.web.fc2.com/touhoku_shinkansen/download-touhoku.html)'s TFO
Train: [E6](http://midnightexpginga81.web.fc2.com/touhoku_shinkansen/download-touhoku.html)

### Result
||v1.7.0.1|This PR version|
|:---:|:---:|:---:|
|FPS|15|30|

## Comment
This improvement is absolutely necessary to make effective use of TFO.
But I haven't got any answers to any known issues.
Can I borrow your wisdom?

In writing this code, I referred to the @LabRatAndy 's code. Thank you very much.


leezer3:
Just tried a random BVE2 train.

I presume you're referring to the 'bouncing' of cab elements, which seem to be shifting slightly back and forth.

That feels to me like there's a precision mismatch somewhere, which is producing imprecise rounding effects (I'd guess at loading sway causing slight camera movements, which alters the rounded result)-
Internally, we've always used double-precision vectors, and cast these to float at the final GL.Vertex etc. but this code uses single precision vectors throughout.

The first thing I'd try is converting all the OpenTK vector / matricies to use the double-precision eqiuivilants. 
As a secondary thing, if we match the struct layouts, it would be possible to use our vector types in unsafe code calls, e.g. as per this experiment https://github.com/leezer3/OpenBVE/pull/347


Will do some other thinking and see if I come up with something else.

s520:
What you say is that GLSL uses single precision, but until then it is managed with double precision.
I would like to try it if I am aware.

cwfitzgerald:
@s520, this is remarkably impressive!

With regards to double precision, I would highly recommend avoiding dp on the gpu at all costs, the cost for dp math on most graphics cards are 16-32x slower than sp.

s520:
I tried using double precision matrix and vector, but still seems to have problems with Z sorting ...

s520:
Currently, the 2D cab is displayed as shown below.

Wrong_Z_Sort

It should be as shown below.

Correct_Z_Sort

leezer3:
Done a little more playing.

It's definitely a double-precision vs. single precision issue, although I can't track where exactly (I suspect in the AnimatedObject file, but the OpenTK Quaterniond is missing the option to rotate via a Vector3d)

https://github.com/leezer3/OpenBVE/blob/master/source/OpenBVE/Parsers/Panel/Panel2CfgParser.cs#L17
Easiest workaround is to bump the StackDistance in the panel parsers up slightly.
0.000005 works and eliminates the glitches.

leezer3:
Lighting-
We don't seem to be applying blending correctly somewhere in the render pipeline.
new

old

Take a look at the edges around the catenary insulator and stanchions. 
When this is fixed, the overall brightness just needs a small boost. Is the shader specifiying the GL.Light or something stupid?

s520:
Well, I changed everything to double precision...
https://github.com/s520/OpenBVE/blob/ImproveFPS/source/OpenBveApi/Objects/ObjectTypes/AnimatedObject.cs#L804
Quaternion.RotationBetweenVectors in the above code returns OpenTK.Quaterniond.

Are there any side effects of changing StackDistance?
If there is no problem, I would like to adopt it.

It may be a shader problem with lighting.
https://github.com/s520/OpenBVE/blob/ImproveFPS/assets/Shaders/default.vert

I'd like to be able to write GLSL code equivalent to conventional writing.

leezer3:
There *probably* isn't any downside to changing the StackDistance, but we're in uncharted territory :P

If we assume that each unit of stackdistance represents 1m, then we're moving the panel layers within the world by fractions of a millimeter, which I don't see as an issue.

s520:
The Z-Sort issue has been fixed.
It was my inadvertent mistake.

The rest is lighting issues.

LabRatAndy:
@s520 I glad that I have provided some inspriation to move the project towards more modern rendering techiques. And have managed to get it to work. 
I have been trying for a number of weeks to get my code to run as I always get an access violation exception when it executed the DrawElements method in the EBO. I was begining to think that openTK was faulty, I'd still like to know what I did wrong though. 
I would also agree with @cwfitzgerald  that double precsion values whilst available should be avoided, at least that's what I've read. 
@leezer3  all details about materials(colours, textures, shinieness, etc) and lights are passed to the shader as uniforms and the shader then calculates the colour based on these. At least that's how I understand it to be.
I followed the learn openGL tutorial https://learnopengl.com/ to teach myself about VBO EBO and VAO etc .

leezer3:
Double-precision is a funny thing TBQH.

It's only really become mainstream / nicely usable over the last 5 years or so. I suspect that in our case it's likely to make very little difference, as we're bottlenecked in so many places it's unreal, & by the time we fix these something else will be the issue.

Thought- Can we provide a (at least temporary) switchable render path?

s520:
@LabRatAndy 
First of all, it may be basic, but make sure that the EBO handle is not 0 and you are not trying to run in a different thread than the main thread.
I inadvertently created an EBO in a worker thread and raised an AccessViolationException :)

I am also opposed to handling double precision in GLSL.
GLSL can use double precision since OpenGL 4.0.
This will drop support for hardware below 4.0...

@leezer3 
I still have no answer on the lighting problem.
Today, let's be able to switch to rendering in the conventional immediate mode.

s520:
Added options and commands to switch renderers.
Please check it.

s520:
For Lighting, almost the same drawing results were obtained with ambient light alone.
The problem seems to be the result of drawing with diffuse light.

assets/Shaders/default.vert
```diff
@@ -0,0 +1,84 @@
...
+
+	// Lighting
+	// Ambient
+	vec4 ambient  = uLight.ambient * uMaterial.ambient;
+
+	// Diffuse
+	vec3 norm = normalize(iNormal);
+	vec3 lightDir = normalize(uLight.position - worldPos);
+	float diff = max(dot(norm, lightDir), 0.0);
+	vec4 diffuse = uLight.diffuse * (diff * uMaterial.diffuse);
```
leezer3: ```vec4 diffuse = (uLight.diffuse * (diff * uMaterial.diffuse)) * diff;```

I think that's where we're going wrong- That gives the expected results, although I need to test a bit more.

leezer3:
Try the change I've made above- This gets us back to the correct lighting results.

Not convinced this is quite right (the math is making my head hurt!), but I think this line is where we're going wrong- Both diffuse lighting vectors need to be multiplied by the diff, not just one.

assets/Shaders/default.vert
leezer3: This also seems to give approximately correct lighting, and thinking about is is probably the correct option:
```vec4 diffuse = (uLight.diffuse * diff) * (diff * uMaterial.diffuse);```

leezer3:
Off, disregard that idea :(
New renderer not enabled.....

We're also missing parts of the ATS overlays from both new and old renderers.

s520:
I confirmed the overlay issue.

Also, the light attenuation rate may be related.

s520:
I feel like the direction of the light is reversed.
Multiplying minus as shown below is a little dark, but the result is closer.
`vec3 lightDir = -normalize(uLight.position - viewPos.xyz);`

leezer3:
Interesting, with that change applied, I get somewhat of a pulsing effect as the camera moves.
(Find a bit of plain straight track and watch the ballast lighting)

s520:
Well, it may be the difference between point light source and parallel light source.
My code seems to represent a point light source.

s520:
Bingo!
Except for the driver's cab, the lighting is perfectly matched!

leezer3:
Much better!

Transparencies are still a little off though.
Try something using the default BVE2 catenary and look around the insulators / struts for a good example.

s520:
I didn't seem to have a transparency problem...

leezer3:
new

It's glitching through somehow.
Really noticable on stuff with massive transparencies, such as this:
new1

s520:
How about this?
An additive face blending issue has been fixed.
GL.AlphaFunc seems to have an effect on GLSL as well as immediate mode.

leezer3:
I think we're about as close as we're going to get with this one- I certainly can't see anything wrong at the minute!

My actual FPS seems a little lower with the new renderer, even with your TFO demos. (On the Plymouth route, my FPS have dropped from ~50 to ~40. On the TFO demos, I'm seeing a similar magnitude drop)

This may well depend on the graphics card though, as I'm using a GTX1060, which has considerably more raw power than the HD4400, and FPS has never been an issue, it's always been the slight stutter on block boundaries that's been a slight issue here.

Do you want me to merge this and do any further tinkering after the community have had a chance to break it?

s520:
It seems that FPS is largely due to the GPU architecture.
Some members tried it, but the results seem to be different depending on the GPU.
I think this renderer should be switched at your own discretion.
Many improvements other than the renderer are included, so I'd like you to merge them once.
After that, I think it's better to make further improvements to the new renderer.

leezer3:
Done, many thanks!

leezer3:
I might rename the Renderer switch option though.

Something like a dropdown:
openGL 1.2
openGL 3.0

This is slightly more informative than the current checkbox, dunno....

PR #399, opened 2019-09-29
WIP: Shader improvements

leezer3:
This PR is intended to make various improvements to the shaders introduced with #397 
The end goal is to remove the reference to OpenTK which the API now pulls in.

### Implemented:
* Marks the LibRenderVertex type as sequential. (As this is being passed to openGL as a pointer, it *must* be sequential. Shouldn't matter, but it should be explicitly stated)
* Renames a couple of shader methods.
* Moves most shader based stuff to LibRender. (The API should have no knowledge of rendering whatsoever- This is the job of the render library)
* Creates Vector3f and Vector2f types in the API. As we're chucking everything about as pointers, we might as well use the internal data structure which we control.

### TODO:
* Change the matricies in the API to use our internal copies. Requires merging #347 into this branch.

### Broken:
* Animated objects using the LED function don't work at the minute, as they have no direct knowledge of the VBO. Need to think about this....

cc @s520


s520:
I think it ’s a good idea.
Please consider the following points.

- Can I change the namespace of LibRenderer2/openGL files to LibRenderer2.openGL? If possible, I would like to divide the namespace into folders.
- Animated using LED is broken even in #397 ? Since the vertices of LEDs change, I intended to deal with them by recreating VAO. However, I couldn't verify enough because I didn't have Animated with LED. Can you provide Animated using LED?

Thank you.

leezer3:
If I recall, the only place LED function is actually really used is in DigitalGauges in a Panel2.

I don't think its broken in the current mainline, I just commented it out temporarily.

Namespace can definitely be changed, this is still very much a draft.

s520:
OK. I will check later.
To be honest, it's a little strange that it doesn't work.

OpenBveTrainSim:
@leezer3 & s520:
Not sure what "LED" abbreviates: Light Emitting Device\Objects?
Though if "Enable the new renderer" is checked, the exhaust clag is not proper on this train:
http://www.mediafire.com/?ertdlgmf0818gv0
and random objects disappear from routes.
Thanx OBTS

s520:
Put simply...
DigitalGauges in Panel2.cfg are internally processed as Animated files.
The function name at that time is LED.
I don't think you can use LEDs from pure animated files.

leezer3:
The exhaust clag issue will be fixed in today's build.

If you've got a route with missing objects. we need a link to take a proper look & details on which objects are missing (These are likely to be animated objects) to actually investigate what's going on properly.

OpenBveTrainSim:
Chris:
Confirming exhaust clag fixed in 2019Oct01 beta, using either old\new renderer.
Still missing objects in this route
http://sites.google.com/site/bveinvirginia/
Specificly, a dumper truck at the first road crossing in prior beta versions is now missing,
along with various building walls missing, using either renderer in the 2019Oct01 beta.
Also, new renderer has noticable keyboard response delay compared to the old rederer,
and the cab view FPS is ~1/3 of old rederer; exterior FPS is only slightly lower.
Thanx OBTS

leezer3:
Try today's build.

This should deal with the missing objects. (A somewhat complicated mathematical oversight....)

Speed improvements may take a while.

leezer3:
Everything is now working off the API copies, other than TrainEditor2.

Some further thoughts:
Matrix4, Vector4 and Quaternion are now probably subject to the OpenTK licence. (OpenTK has been used as a reference source, with chunks of code lifted with slight modifications); Although the base math isn't copyrightable, the licence is OK and we should credit accordingly.....

This branch should be slightly faster at the minute (much less boxing between types), but can likely be sped up further with some more unsafe pointers to openGL.

This also increases the API surface area a little, but on the flip side means that there's no reference to openTK required for consumers. Definitely important for anyone wishing to build a new parser plugin I think.

OpenBveTrainSim:
Chris:
Confirming previously missing route objects now appear\fixed with 2019Oct02 Beta version.
Keyboard response for NEW renderer now improved though still slower than OLD renderer.
FPS comparing NEW / OLD renderers still ~1/3 for cab view, ~9/10 for exterior view.
Thanx OBTS
PS: LED must be "secret hacker" vocabulary?

s520:
@OpenBveTrainSim 
LED has the same general meaning as an abbreviation for Light Emitting Diode.

leezer3:
Interesting- The vanishing objects bug isn't quite right yet.

Just found another variant, which only appears to apply to animated files using a mixture of animated and static parts.

```
;-------------------------
;NWM BVE Route Object File
;-------------------------
;
;Object by TSC team
;Not for commercial use.
;This object and its textures may be freely reused. 
;Please acknowledge the original authors and leave this header in place.



;module 104 quarry bank 75m from signal

[Object]
States = lc\lc_l_ob.csv

[Object]
Position = 3.8,0,0
States = lc\lc_r_ob.csv

;flashing red lights
[Object]
States = lc\lc_red1.csv, lc\lc_red2.csv 
StateFunction = floor[mod[TIME * 2, 2]]
TranslateYFunction = ((section>0) | (trackdistance >= -50 & trackdistance <= 75)) * (10)

;left barriers
[Object]
Position = -0.2,1,0
States = lc\lc_b1.csv
RotateXFunction = ((section>0) | (trackdistance >= -50 & trackdistance <= 75))   *(1.32)
RotateXDamping = 1, 1.2

[Object]
Position = -0.2,1,25
States = lc\lc_b2.csv
RotateXFunction = ((section>0) | (trackdistance >= -50 & trackdistance <= 75))  *(-1.32)
RotateXDamping = 1, 1.2

;right barriers
[Object]
Position = 14.4,1,0
States = lc\lc_b1.csv
RotateXFunction = ((section>0) | (trackdistance >= -50 & trackdistance <= 75)) * (1.32) 
RotateXDamping = 1, 1.2

[Object]
Position = 14.4,1,25
States = lc\lc_b2.csv
RotateXFunction = ((section>0) | (trackdistance >= -50 & trackdistance <= 75)) * (-1.32)
RotateXDamping = 1, 1.2
```

With this (placed as a freeobj, anything will do for the referenced object types), only the parts using a StateFunction / RotateFunction work on either this branch or mainline.
Something in the optimisation routines?

cc @s520

Issue #411, opened 2019-12-08
Panel.xml is not works well at 1.7.1.0


ginga81:
This sample touch elements demo data is works well at 1.7.0.3.
[testpanel-demo.zip](https://github.com/leezer3/OpenBVE/files/3935760/testpanel-demo.zip)
But at 1.7.1.0, this data is seems as panel.animated.

Additionally, the touch area can recognition seems perfectly, it cannot work to the changing the screens.
(It seems not work, but the screens are in close contact with each other and displaying quickly about 60fps. So, it is working maybe good.)

When we are using the Panel.xml, the all of 3D object's UV are not pasted, all white.
Please look at this screenshot.
![Panel xml](https://user-images.githubusercontent.com/22734699/70382874-d65e6f00-19a6-11ea-8302-bf78e7b76e86.jpg)

When this sample data is loading and if the Load in advance option is off, and Unload unused textures or Enabale the new renderer option is on, OpenBVE is crash with coredump.

Can you check it out ?

leezer3:
I can reproduce the crash, but not the UV issue.
I presume you should have a grass texture on the ground and textures on the objects in the background?

May take a few days to get to the bottom of it, but will try and do so. First impressions is that a bufffer isn't bound correctly somewhere.
cc @s520 any thoughts?

s520:
I cannot reproduce the UV problem, but I can reproduce the panel display problem.
I think the display issue on this panel is due to [the commit](https://github.com/leezer3/OpenBVE/commit/52b8b08346fe2025b37680474d6a0280f582b8c7) made to the #405 proposal.

The following nightly builds were normal.
https://vps.bvecornwall.co.uk/OpenBVE/Builds/OpenBVE-2019-11-24.zip

leezer3:
Interesting, thanks!

I can't see anything immediately obvious in the camera restriction commit, and it doesn't touch any openGL / VBO calls at all.

I'll try and pull apart all the commits from the 28th over the next few days and see where we go.

s520:
I think the camera restriction has been lifted accidentally in the case of Panel.xml.
I think the following parts are suspicious.
https://github.com/leezer3/OpenBVE/commit/52b8b08346fe2025b37680474d6a0280f582b8c7#diff-6dea8cfed5f14fc09b2fefad85e9802cL75

s520:
Also, it seems that the screen transition due to touch is not successful in the November 24 nightly build.
This should be investigated separately.

[...]

PR #416, opened 2019-12-31
Fix: the issue that the screen cannot be changed by touch

s520:
This PR fix the issue that the screen cannot be changed by touch. (#411)

If you give true to the initialization function of the CarSection class, objects that should be hidden will also be displayed by the initialization function of the AnimatedObject class.


leezer3:
Great, thanks :)

PR #420, opened 2020-01-10
Change: Move to an abstract cargo type

leezer3:
### Significant Changes:
* Moves cargo to be an abstract type on a per-car as opposed to per-train basis.
* Implements cars which carry no cargo.

### Rationale
BVE assumes that all cars carry passengers, and hence thier mass is updated according to the passenger ratio when stopping at a station.
This isn't valid for locomotives, and isn't really helpful if we're trying to move beyond our current bounds (for example, later we might want station specific cargo)

Small baby steps at the minute- Fuel usage needs to be implemented for motor cars (if appropriate), but that will feed into another modification of the mass figure, and should not be done as a cargo type; otherwise we'd need far more subtypes.

### DIFF
--- a/source/CarXMLConvertor/Convert.ExtensionsCfg.cs
+++ b/source/CarXMLConvertor/Convert.ExtensionsCfg.cs
@@ -454,6 +454,8 @@ namespace CarXmlConvertor
 				newLines.Add("<MotorCar>False</MotorCar>");
 				newLines.Add("<Mass>" + ConvertTrainDat.TrailerCarMass + "</Mass>");
 			}
+			// BVE has all cars as passenger carrying by default
+			newLines.Add("<Cargo>Passengers</Cargo>");
 			if (CarInfos[i].AxlesDefined)
 			{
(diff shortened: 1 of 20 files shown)


leezer3:
The basic passengers have now been converted to use an abstract cargo type.

Going to see if I can implement a cargo type which doesn't change loading next, and allow this to be changed somehow. (XML field for cargo type?)
Need to see how this interacts with the sim as well, may throw a fit if one of the cars is no longer the weight it expects....

leezer3:
Reworked relative to the current master, and added an XML cargo tag.

The main use of this (currently) would be for locomotives which obviously do not carry cargo; At present, the mass for these is updated according to the passenger ratio at the most recent station. 

Probably mergable at present, but will do so after the next point release.

leezer3:
Rebased onto latest master, and a few more tweaks.

Have also implemented a basic robust frieght type. This keeps the car mass calculations the same, but this does not damage, and hence has no effect on scoring.

Needs documenting before merging ( https://github.com/leezer3/OpenBVE/issues/746 !!!)

Issue #426, opened 2020-01-25
Excessively high rail numbers causes a System.OutOfMemory Exception


RetentiveGOAT:
### Description

I can't load certain routes with the new renderer as it throws a System.OutOfMemory exception but I have 8 GB of RAM, 8 GB of GDDR5 RAM for my video card and an Intel Core i5-7400. I found that I cannot load the NYCT 3 line, as well as the 5, and other NYCT routes. It seems to access a lot of memory but doesn't load the route.

### Reproduction

If the issue occurs in multiple routes/ trains, please provide one or two samples.
In order to reproduce the issue and debug it, it's helpful to have the following:

## Route

Download Link: https://mega.nz/#!0YRhEapQ

Key: ybQRS2o3AFBKa6Dz47UBMVDZgxxDXW8VvkQNIvL6Pzk

## Train

n/a. Applies to every train.

## Logs

Logs may be accessed through the 'Report Problem' button on the bottom left of the screen.
Please provide a log, and a crash-log if the game actually crashes.

### Related information 

* Operating system: Windows 10
* Method of control: Keyboard

leezer3:
This one seems to be being caused by some exessively high rail numberings (Over 1000), which is overflowing the available memory in 32-bit .Net.
It's not the new renderer per-se, it's just tipped over the edge.

Should be able to do something, but may only be a hacky fix as a first step.

Will try and sort, although this one is somewhat complicated

leezer3:
Fixed in the latest daily (31st January)

Will get a full build out on the site sometime in the next week or two :)

PR #427, opened 2020-01-27
Change: Shift rails to a dictionary accessor.

leezer3:
This changes rails to use a dictionary type accessor, as opposed to a single array.
This drastically reduces memory usage in the route parser and sim where an author has used mega high rail numbers.

### Issues:
* ~~Curve calculation for trains running on secondary tracks seems broken. (Probably using a for loop somewhere we shouldn't now...)~~ Fixed.
* ~~Not implemented in Route Viewer / Object Viewer, these are broken in this PR at the minute.~~ Route Viewer doesn't implement the parser changes, but implements the backend changes. Should be enough, and hopefully will move the route loading code to a plugin soon.

cc @s520 - Multiple rail running is primarily your code.

### DIFF
--- a/source/ObjectViewer/Hosts.cs
+++ b/source/ObjectViewer/Hosts.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Drawing;
 using OpenBveApi.Hosts;
 using OpenBveApi.Interface;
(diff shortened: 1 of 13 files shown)


leezer3:
So, I *think* this now works as designed.
Loading any route with non-continous rail numbers will be faster in general, and will use less memory.

https://github.com/leezer3/OpenBVE/issues/426 now loads nicely, and uses a sensible amount of memory (300mb vs over 3gb before crashing).

KiryKirályliget works OK, as do the demo OtherTrain stuff from @ginga81 & @s520

Any chance of this being tested before I merge by someone else?

PR #431, opened 2020-02-14
Add: license files [ci skip]

s520:
This PR adds missing license files.


leezer3:
Great, thanks :)

I'll try and remember to look at the licence dialog in the main sim too.
This either needs updating or linking to each.

PR #434, opened 2020-02-19
Rail Sounds- Add event and fix issues

leezer3:
https://bveworldwide.forumotion.com/t1852-rail-sound-delayed#19276

This prompted me to add a rendered event for the RailSounds change to investigate the issue.

Findings:
* The events appear to be being correctly placed.
* Note however, there appears to have been a bug in that the RailSounds change event was duplicated in every block.
* Code appears sane and I can't find an immediate issue, but still looking.

Don't merge until this has been tested properly, especially in case the event block correction causes something I haven't noticed....

Other thoughts-
* Can likely dump the TrackPositionDelta from the RailSoundsEvent constructor- I can't think of a sensible reason to use this.
* Is it worth adding an independant rail sounds change trigger? Probably not, but could be useful for very little work.


leezer3:
Now I find it, must have been a copy and paste error somewhere as 1.4.3 is ok.....

PR #444, opened 2020-03-19
Route Loader Plugin [WIP]

leezer3:
This PR contains a WIP route loader plugin.

At present, the route loader (from the main sim) has been cloned into a new plugin, and various changes made.

### Current Status:
* Plugin builds.
* Plugin loads.
* Routes load.
* It works (?!?!?)


leezer3:
cc @s520 -
A lot of stuff moved around in here, some may need to move again; Much of the positioning is experimental. 

I don't much like the current way of loading the TrackFollowingObjects either, but it'll need to be like this until the TrainManager has been moved into a shared library and we can implement the train loading plugins (?)

s520:
I'm examining undocumented features from source code.
(I have already found several features that are not described in the Wiki.)
Once this is done, I will try this PR.

leezer3:
Still not working- Fails when attempting to split into expressions (well before my other changes)

Need to backtrack what I've broken, will do that this weekend.

leezer3:
That was terminally stupid, messed up the format checking so it thought everything was RW files :(

Now a little closer to working, but the route isn't now updated quite right on the host application, so when we create the TrackFollowers to get the world positions when calculating curves it all messes up.

Possibly move the CurrentRoute to a property of the host application?
Might alternatively be possible to create a subset TrackFollower which follows it's own array?

Play around......

leezer3:
OK, another terminally stupid bug squashed (variables should have been updated, so moved to just a static for the minute), and this thing appears to work.

Next going to move it so it does the preview too, followed by stripping out the parser from the main sim entirely.
Assuming that works, will then do the same with Route Viewerr.

When that's done, will look at some further moving and re-writing to make this thing a little more manageable.

leezer3:
So, this now (well at least seems to!) work properly :)
The parsers have been removed from both the main sim and Route Viewer, and all appears well at the minute.

With a change this large, there's definitely a distinct possibility that things will break nastily so I'm not merging any time soon. Going to try and do some more, and attempt to break this over the next few days.
Still need to consider whether some bits want to move around too. 
Final thing I'd like to do is to remove the currently required renderer reference passed to the plugin.

leezer3:
Done a little more work to this, and fixed a few more issues, as well as removing the renderer reference from the route parser plugin.
This has really added too many API methods for my liking, but still.....

**Should** be working OK in the main program.
@s520 - Did you get a chance to take a look at this?

s520:
Sorry for my late reply.
I will watch this on the weekend.

s520:
I could not read successfully on multiple routes.
The screen is dark.

BAD

v1.7.1.3
OK

Route: http://www.nozomi.vc/09_tokakanse/

leezer3:
Hmm, the lighting is broken.

Everything is there (switch to wireframe), just seems to be lit with pure black.  I'm sure it wasn't doing this earlier......

Need to so some more fiddling obviously.

leezer3:
That was stupid, I'd forgotten to copy across the lighting defaults....

Does that work for you now?

s520:
I confirmed that the route was displayed successfully. (main sim/viewer)

leezer3:
Need to rebase (again....)

Intending to merge this into the main branch after the next release, which should be imminent.
Can then work out any kinks once the general board has had a go at it.

leezer3:
Pulled forwards relative to master.
Prequsite for https://bveworldwide.forumotion.com/t1877-animating-pantograph-height-changes#19700

### TODO:
* New compatability signals aren't passed yet.
* ?? Further refactoring to object creation in the final loop ??
* ?? Move some of the support classes to RouteManager2- BVE5 parser will want similar ones ??
* TEST!

leezer3:
I'm now deliberately testing this, with a view to merging into mainline next week and letting the general community loose on it.

Haven't done quite as much of the refactoring as I'd like, but getting this into the main build will make things a ton easier in the long run.

### Later Work:
* Route Viewer doesn't support time of day at the minute. This allows dynamic / XML based backgrounds to work, but the time in Route Viewer is just currently synced to the system clock, as opposed to anything from the route. Probably needs a specific set time key or something, don't know....

OpenBVE.sln
```diff
@@ -34,6 +34,7 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OpenBve", "source\OpenBVE\O
 		{678031C7-6D87-4F5F-9C4A-12F86B76342C} = {678031C7-6D87-4F5F-9C4A-12F86B76342C}
 		{B520B1D7-3889-4C88-9E0F-CB96802D5CD1} = {B520B1D7-3889-4C88-9E0F-CB96802D5CD1}
 		{F714A9FA-1C4B-456B-BACF-D6763223A81C} = {F714A9FA-1C4B-456B-BACF-D6763223A81C}
+		{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC} = {FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}
```
leezer3, line 37: Need to make sure the thing rebuilds as a dependancy of Route Viewer / Main program

PR #445, opened 2020-03-24
Change: use API functions to determine encoding

s520:
This PR contains the following changes in addition to the title.
- Move: the function that determines the extension of the object file to the API
- Fix: the range of $Char and $CharAscii argument
- Fix: typo in CsvRwRouteParser.PreprocessOptions.cs

### DIFF
--- a/source/ObjectViewer/Hosts.cs
+++ b/source/ObjectViewer/Hosts.cs
@@ -211,48 +211,22 @@ namespace OpenBve {
 
 		public override bool LoadObject(string path, System.Text.Encoding Encoding, out UnifiedObject Object)
 		{
-			TextEncoding.Encoding newEncoding = TextEncoding.GetEncodingFromFile(path);
-			if (newEncoding != TextEncoding.Encoding.Unknown)
+			if (string.IsNullOrEmpty(path))
 			{
-				switch (newEncoding)
-				{
-					case TextEncoding.Encoding.Utf7:
-						Encoding = System.Text.Encoding.UTF7;
-						break;
-					case TextEncoding.Encoding.Utf8:
-						Encoding = System.Text.Encoding.UTF8;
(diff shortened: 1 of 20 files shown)


leezer3:
Thanks :)

I think the whole extension checking is unncessary-
That is the intention for the *CanLoadObject* function in the object interface. 

Off the top of my head, it's only CSV / B3D files which have no actual format detection routine (magic bytes or similar), and these perform a simple extension check within that function.

Only real way I can think of to improve this detection is by a textual search for the two variant **MeshBuilder ** statements over the first 20 lines or so.

---

I believe the only place BVE2 / BVE4 legitimately allows for files with no extension is in the Signal declarations.
This extends it to the entire *LoadStaticObject* function, which I think is wrong, unless you've got an example / documentation otherwise?

leezer3:
e.g. DirectX plugin:
https://github.com/leezer3/OpenBVE/blob/master/source/Plugins/Object.DirectX/Plugin.cs#L28

Runs a check on the magic bytes for the format.

s520:
The previous code allowed for files with forgotten extensions, so I added that.
https://github.com/leezer3/OpenBVE/blob/cb9a9d99a0d982561d1fe40e8a953ee874230781/source/OpenBVE/Game/ObjectManager/ObjectManager.LoadObject.cs#L31-L72

What were the intentions of adding this feature previously?

I will delete it if it was not intended for use other than for signals.

leezer3:
If you look at the call sites, it was never actually being used- Everything other than the signals specifically checks for the existance of the file before passing it to the LoadObject function.
The signals have their own function to do this.

s520:
OK.
I will remove the corresponding code from LoadObject and LoadStaticObject.

s520:
I removed it.
Please check it.

PR #449, opened 2020-04-04
Fix: issues with changing speed and rail index while driving at TFO

s520:
## An issue with changing speed
If you specify the deceleration or acceleration as shown below, TFO will stop for a moment.
```xml
<Stop>
  <Decelerate>1.62</Decelerate>
  <StopPosition>400</StopPosition>
  <StopTime>00.0000</StopTime>
  <Accelerate>0</Accelerate>
  <TargetSpeed>60</TargetSpeed>
  <Direction>1</Direction>
  <Rail>2</Rail>
</Stop>
```
I made the following changes to fix this problem.
- Add `<Point>`
This element can specify the passing speed of the point.
However, it has no stop-related elements. (`<Doors>`, `<StopTime>`, `<Direction>`)
```xml
<Point>
  <Decelerate>1.62</Decelerate>
  <Position>400</Position>
  <PassingSpeed>60</PassingSpeed>
  <Accelerate>0</Accelerate>
  <TargetSpeed>60</TargetSpeed>
  <Rail>2</Rail>
</Point>
```
- Add `<Points>` to `<Stops>` aliases.
- Add `<Position>` to `<StopPosition>` aliases.

Also, the first and last elements of `<Points>` (or `<Stops>`) must be `<Stop>`.

## An issue with changing rail index
Here, the top is defined as the far side of the route, and the last is defined as the front of the route.
When the beginning of the TFO passes the point where the rail index changes, the rail index changes for all axles.
Therefore, if the rail after the change does not overlap the rail before the change by the length of the train, the axle position will be incorrect.
This PR change will change the axle rail index when the axle passes the point where the rail index changes.

### DIFF
--- a/source/OpenBVE/Game/AI/AI.TrackFollowingObject.cs
+++ b/source/OpenBVE/Game/AI/AI.TrackFollowingObject.cs
@@ -1,5 +1,6 @@
 ﻿using System;
-using System.Collections.Generic;
+using System.Linq;
+using OpenBveApi.Routes;
 using OpenBveApi.Trains;
 
 namespace OpenBve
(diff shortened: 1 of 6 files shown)


leezer3:
@s520 
https://bveworldwide.forumotion.com/t1883-objects-in-an-animated-container-disappear-in-routeviewer-and-openbve-after-the-insertion-point-of-the-animated-file#top

I think this has broken something with the point changes.

PR #464, opened 2020-04-22
Update: Building.md [ci skip]

s520:
This PR update Building.md.
- If NuGet version is lower than 2.16, package restore fails.
- Build fails when using Mono package from distribution.
  - On Fedora I have confirmed that the same version also fails. We need to see what happens in Debian.
  - Ubuntu on Travis seems to get Mono from the Mono project repository.

Related: https://bveworldwide.forumotion.com/t1890-missing-files-in-compilation-of-source-in-linux-not-debian

### DIFF
--- a/Building.md
+++ b/Building.md
@@ -14,12 +14,12 @@ This file is a WIP set of instructions for building openBVE from source.
 #### When to use Mono
 
 - Mono 5.20.1 or later, x86 or x64
-- NuGet 2.16 or later
+- NuGet client 2.16 or later
 
 ### Linux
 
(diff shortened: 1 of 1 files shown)


s520:
I checked the following contents.

|                 | Distributor's Mono version | Build status |
| --------------- | -------------------------- | ------------ |
| Debian buster   | 5.18.0.240                 | Failed       |
| Debian bullseye | 6.8.0.105                  | Success      |
| Fedora 31       | 5.20.1                     | Failed       |

For Fedora 31, the same version of the Mono project's repository can be successfully built, so it's possible that a patch from the distributor is affecting it.

PR #465, opened 2020-04-24
Fix: Building.md [ci skip]

s520:
The changes are as follows:
- Added description about GNU Make.
- Fixed the command to install NuGet client.

I searched the openSUSE build environment to solve the problem posted on the forum.
Would you like to add a note about openSUSE?

### DIFF
--- a/Building.md
+++ b/Building.md
@@ -21,6 +21,7 @@ This file is a WIP set of instructions for building openBVE from source.
 - Mono 5.20.1 or later, x86 or x64
 - NuGet client 2.16 or later
 - OpenAL
+- GNU Make
 - debhelper (Debian and compatibles only)
 
 #### Debian and compatibles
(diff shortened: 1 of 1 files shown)


leezer3:
No idea, never tried openSUSE for anything serious.

If we're not careful though, we'll have a set of instructions for everything under the sun.
My impression from the post in question was of someone rather non-technical. 

Marginally surprised that make isn't a default install on Debian.
Might be better to link to the build-essential metapackage in this case? We don't use the other components of it, but it's a little more well-known, and probably far more useful.

s520:
I agree with your thoughts.
There are descriptions of the major distributions, Debian and RHEL, which should be sufficient.

I changed to use metapackage instead of make.
If there's nothing else to write about, please merge this PR.

s520:
Also, we need to link [the build page](https://openbve-project.net/source/) on the official homepage to this file.

leezer3:
OK, took a little messing around, but the build documentation now mirrors this file.

s520:
I checked it out.
Thanks!

PR #475, opened 2020-05-03
Change: Add try / catch to some more GL inits in case of failure

leezer3:
https://github.com/leezer3/OpenBVE/pull/474

Wrap the VBO creation in a catch block. 
Testing at draw time is going to be an expensive no-go, but this at least eliminates another point of failure.

Question though is whether we bother....
Assuming the shader has been created properly, we *ought* therefore to have a valid GL 3.0 context and thus the ability to create VBOs etc.

cc @s520

### DIFF
--- a/source/LibRender2/BaseRenderer.cs
+++ b/source/LibRender2/BaseRenderer.cs
@@ -173,10 +173,12 @@ namespace LibRender2
 
 		internal int lastVAO;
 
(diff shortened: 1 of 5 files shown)


s520:
As you say, if the shader compiles properly, we can create VAO etc.
If we check the shader status, this try-catch is probably unnecessary.
If you choose to adopt this, we should write it to the log.

PR #479, opened 2020-05-09
Move door interlock restrictions to TrainManager and fix related AI behaviour

MarcRiera:
This pull request solves the following issues:

1. The checks related to door interlocking were scattered in the code for user input, AI and stations. They are now part of TrainManager.
2. The AI would only open the doors if both sides were unlocked. If a plugin restricted the doors to only one side, the AI would not even try to open the doors and wait indefinitely.
3. At stations where doors should be opened on both sides, boarding does not start until both sides are open. The AI now keeps trying to open all doors in this situation even if one side is already open, in case partial door interlocking was applied initially, to avoid the AI from waiting indefinitely doing nothing.

By the way, the AI currently does not notify plugins about the recently added door virtual keys. Given that the AI is supposed to represent a real human pressing buttons and that plugins may rely on them for functionality regardless of the state of the AI, I think it makes sense to add this too. What do you think?

### DIFF
--- a/source/OpenBVE/Game/AI/AI.SimpleHuman.cs
+++ b/source/OpenBVE/Game/AI/AI.SimpleHuman.cs
@@ -185,7 +185,14 @@ namespace OpenBve
 						{
 							Train.ApplyEmergencyBrake();
 						}
-						
+						if (Train.Station >= 0 && stopIndex >= 0 && Train.StationDistanceToStopPoint < Program.CurrentRoute.Stations[Train.Station].Stops[stopIndex].BackwardTolerance && (Program.CurrentRoute.Stations[Train.Station].StopsHere(Train) & (Program.CurrentRoute.Stations[Train.Station].OpenLeftDoors | Program.CurrentRoute.Stations[Train.Station].OpenRightDoors) & Math.Abs(Train.CurrentSpeed) < 0.25 & Train.StationState == TrainStopState.Pending))
+						{
+							// doors not fully open at station - open doors
+							if (Train.Specs.DoorOpenMode != TrainManager.DoorMode.Automatic)
+							{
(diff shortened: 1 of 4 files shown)


leezer3:
That's an interesting thought.

No idea how practical that would be, but it makes reasonable sense.
On the flip side, a *lot* of what's currently handled by plugins should IMHO be part of the main sim; There's far too much duplication of effort involved- For example, rain and wipers are on my perennial todo list.

MarcRiera:
I found out about this while developing a train plugin which implements a door selection mechanism that applies door interlock. The doors need to unselect automatically after the driver presses the button to close the doors, so the easiest way is using Virtualkeys.LeftDoor/VirtualKeys.RightDoor as a trigger. Door states are not an option because they update after the doors are fully closed.

Implementing key presses for the AI may be a bit tricky, specially when closing doors. The AI always tries to close both sides, so we would need to check before which side is open and only trigger that one (if we want the AI to look like an experienced driver and not someone just pressing random buttons around). Maybe we can merge this for now and I will look at the buttons separately and try to find a reliable solution.

I agree that there are lots of things currently done with plugins (wipers, cab lights, pantographs, head lights, etc.) that could be implemented in the main sim. There are far too many plugins implementing the same over and over.

leezer3:
Sounds reasonable.

Have you considered implementing the door logic in your plugin's AI method, as opposed to the main sim AI?

MarcRiera:
> 
> 
> Sounds reasonable.
> 
> Have you considered implementing the door logic in your plugin's AI method, as opposed to the main sim AI?

Yes, I already use the AI method to expand the built-in AI to the plugin's features, but this door logic is a safety device that is active too with a real driver, not just AI. It works just fine when the user drives the train, it is only when using the AI that it does not work due to OpenBVE not passing the state of the button. By the way, I guess this also applies to the variables related to the door buttons for animated objects, not only plugins.

PR #486, opened 2020-05-23
New: Support animated gifs as textures

leezer3:
This adds support for using an animated gif as a texture file.

Some nasty, hacky code in there, but appears to work.

Needs testing to ensure it doesn't explode before merging....


leezer3:
http://vps.bvecornwall.co.uk/OpenBVE/Experimental/AnimatedGif2.zip

Might be slightly less broken this time.....

Major issue is that this pulls the entire texture off disk every time to get the new frame.
Can probably cache this in the texturemanager, but get the thing actually working reliably this time.

ginga81:
I tested AnimatedGif2.zip, but still white objects and errors.
most of objects are errors.
some of one's are...

main program's default png
21:44:35  No plugin found that is capable of loading texture Data/Compatibility/stop.png
21:44:35  Plugin Texture.BmpGifJpegPngTiff.dll raised the following exception at LoadTexture:Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: startIndex

object folder's sample gif
21:44:35  No plugin found that is capable of loading texture .../Railway/Object/TFO_sample/sample.gif
21:44:35  Plugin Texture.BmpGifJpegPngTiff.dll raised the following exception at LoadTexture:A null reference or invalid value was found [GDI+ status: InvalidParameter]

Train folder's outer-view's object
21:44:35  No plugin found that is capable of loading texture Train/Series-103/jnr103/tsuma.png
21:44:35  Plugin Texture.BmpGifJpegPngTiff.dll raised the following exception at LoadTexture:A null reference or invalid value was found [GDI+ status: InvalidParameter]

source/LibRender2/Textures/TextureManager.cs
```diff
@@ -198,29 +203,69 @@ public Texture RegisterTexture(Bitmap bitmap, Bitmap alpha)
...
+				}
+				else
+				{
+					texture = animatedTextures[handle.Origin];
+				}
+				double elapsedTime = CPreciseTimer.GetElapsedTime(handle.LastAccess, currentTicks);
+				int elapsedFrames = (int)(elapsedTime / handle.FrameInterval);
+				if (elapsedFrames > 0)
+				{
+					handle.CurrentFrame += elapsedFrames;
```
leezer3, line 233: Runaway number.
Needs to be modified.

source/LibRender2/Textures/TextureManager.cs
```diff
@@ -198,29 +203,69 @@ public Texture RegisterTexture(Bitmap bitmap, Bitmap alpha)
...
-			Texture texture;
-
-			if (handle.Origin.GetTexture(out texture))
+			if (texture == null && handle.Origin.GetTexture(out texture) || texture != null)
 			{
+				if (texture.MultipleFrames)
+				{
+					handle.MultipleFrames = true;
+					handle.FrameInterval = texture.FrameInterval;
+					texture.CurrentFrame = handle.CurrentFrame % texture.TotalFrames;
```
leezer3, line 267: Optimise this block in conjunction with above?

source/LibRender2/Textures/TextureManager.cs
```diff
@@ -198,29 +203,69 @@ public Texture RegisterTexture(Bitmap bitmap, Bitmap alpha)
...
+				double elapsedTime = CPreciseTimer.GetElapsedTime(handle.LastAccess, currentTicks);
+				int elapsedFrames = (int)(elapsedTime / handle.FrameInterval);
+				if (elapsedFrames > 0)
+				{
+					handle.CurrentFrame += elapsedFrames;
+					handle.LastAccess = currentTicks;
+					for (int i = 0; i < 4; i++)
+					{
+						GL.DeleteTexture(handle.OpenGlTextures[i].Name);
+						handle.OpenGlTextures[i].Valid = false;
```
leezer3, line 238: Optimise- Can probably use a jagged array to generate the GL textures for all frames on demand and cache

ginga81:
I thought after that when we use this animated gif with transparent, we can beautiful effect to the steam locomotive smoke!
And I thought that at the future, APNG more than useful from animated GIF.

https://github.com/xupefei/APNG.NET

leezer3:
Now that's an interesting thought.
My inclination though would be to not support transparent color in these, but instead to suggest the use of an alpha channel. 
(Transparent color has only ever been a slightly hacky workaround for the lack of an alpha channel)

Animated GIFs with an alpha channel work OK with the current code.
If animated PNGs are implemented, I don't see why this wouldn't be the same.

### Samples:
(This contains one object with an alpha gif and one with the solid gif)
[Objects.zip](https://github.com/leezer3/OpenBVE/files/4705580/Objects.zip)

ginga81:
I tested https://github.com/leezer3/OpenBVE/commit/b7b55002e20264163b354f8686ab61750617db67
Sample- solid color.b3d is animates about 1 minute par frame.
sample2- alpha channel.b3d is very fast animation, but noisy shown as below.
noisy

leezer3:
Interesting.

Both are OK on Windows, I'll try the VM later.
Haven't got any good ideas as to why just at the minute, maybe the timer isn't quite right (I'd expect both to have the same speed issues in that case though)

leezer3:
@ginga81 Please try building / testing the latest commit from this branch.
I've replaced the Mono GIF decoder entirely, and I think this solves most of the issues you were seeing.

Animation speed still seems to be a little off on my Debian virtual machine, but I think this is most likely because clock-ticks in virtual machines aren't too consistant.
Might need to revisit that in the timer class, but I don't think that running OpenBVE under a virtual machine is a common thing.

ginga81:
I built the branch of animatedgif after 'Change: Implement our own GIF decoder, as the Mono one is no good'.
And I ran of the version 1.7.561.129-user 64bit.
But, the animated gif is not shown, and only white.
What will do I do?
white animated gif

leezer3:
Hmm...
I'll have to try that myself from built directly on Linux.

It worked OK when I copied it to the VM, so something must be broken when compiled differently (?)

ginga81:
Certainly I can't believe my build is really the latest gifanimated.
Because, I am not still have skill perfectly about programming and compile. 
I would appreciate it if you could tell me how to make sure that I built the latest one.

leezer3:
Don't worry, it's my fault, and is actually somewhat more broken than I thought.

I've partially fixed it, but solid colors are now glitchy.

TLDR:
Initial issue that you saw was with having multiple animated GIFs in the world, which I must have overlooked to start with.

ginga81:
I tested download and rebuild to 1.7.564.368-user 64bit.
But, still whited.
Sorry...

Kenny-Hui:
It works here, but I have 2 questions:
1. On my machine it takes ~4s to load an ~2.5s 200x200 gif. During the loading it will freezes OpenBVE til it's loaded.
I am worried that the player will think OpenBVE crashed if we put a longer gif.
Also since the simulation does not actually pause when OpenBVE is freezed, there might be a chance that the player will miss the station stop, or even ATO will overrun. Is there anything we can do about this, or we're stucked?


2. The transparency is odd, it appears to automatically remove the background despite not adding any TransparentColor command. This applies for all type of background color, whether it's blue or black. It almost looks like OpenBVE will look for the color that covers the most out of the image, then set it to transparent.

image
(TextureCoordinates is flipped, but shouldn't matter)

Here's my current code:

`CreateMeshBuilder,`
`AddVertex,10.05, -0.0, 1.04, 0,1,0`
`AddVertex,10.05, 0.5, 1.50, 0,1,0`
`AddVertex,9.65, 0.5, 0.90, 0,1,0`
`AddVertex,9.65, -0.0, 0.44, 0,1,0`
`AddFace,3, 2, 1, 0`
`LoadTexture,t.gif,`
`SetTextureCoordinates,0, 0, 1,`
`SetTextureCoordinates,1, 0, 0,`
`SetTextureCoordinates,2, 1, 0,`
`SetTextureCoordinates,3, 1, 1,`

leezer3:
The transparency is a glitch in the GIF decoder.
I need to look into that one.

TLDR:
A GIF animation in some modes only stores the changed pixels from the previous frame.
The decoder implementation at the minute I don't think us respecting this quite right.
(The .Net decoder was better, but the Mono inbuilt version seems buggy as hell, hence trying the alternative)

leezer3:
The GIF decoder should be somewhat better now, and will *probably* sort your problematic GIF.

It's not quite right on some though, and needs a little more work.

**Loading times:**
It doesn't much surprise me that your GIF is slow to load, these will be at the minute. There's a bunch of optimisation that can already be done once the thing works right (multiple passes through bitmaps whilst loading), but ultimately, your GIF is probably loading 20 odd textures into OpenGL memory at one per frame, and this will never be particularly fast.
In-game time *shouldn't* be advancing whilst textures are loading in, will have to take a closer look at this.

leezer3:
@Kenny-Hui try the latest commit from the branch.
This should massively speed up the animated GIF decoder, although it's a little slow when loading one of these into view for the first time.

Possibly, we ought to preload all animated GIFs into memory at the start of the game and exclude them from texture unloading.
A little concerned about overuse of these mind.....

Kenny-Hui:
Much better! (Around 0.1s lag spikes, does not affect gameplay either)
Though I've noticed some transparency issue
image
The left is 1.8.0.1, the right is this gif branch. I've made sure the background color is 0,0,255, and I have added SetDecalTransparentColor.

This only happens on specific image file, so I linked it below:
[SpeedModeSwitch.zip](https://github.com/leezer3/OpenBVE/files/6515643/SpeedModeSwitch.zip)

ginga81:
I tested this Objects.zip, but not animated.
Noises are removed completely clean.
https://github.com/leezer3/OpenBVE/pull/486#issuecomment-636321342

leezer3:
Please try the latest commit from the branch.

This will hopefully get everything animating correctly at the correct speed, or at least it definitely does on my Debian VM.
I don't quite understand why this works (it shouldn't be any different), but there we go.

Not going to merge this before the next release (might be at the end of the week if nothing else broken turns up in the meantime), but will do so afterwards.

ginga81:
I tested this branch with compile.
but, AnimatedGIF is white.
[image: Screenshot from 2021-07-06 04-08-01.png]

2021年7月6日(火) 3:57 Christopher Lees ***@***.***>:

ginga81:
I created the animatedgif and this daily-build.
But, still white and no shown.

leezer3:
It's not in 1.8.2.0, which is why it's just showing the first frame of the GIF.

I've tested the daily build just now on a Debian VM, and this definitely works correctly for me with the sample objects I uploaded above.

At the minute, white suggests it's not loading the texture correctly to me. 
If it's one you've created yourself, its possible the decoder doesn't like it (this is more likely as you're on Linux), but not sure....

Are there any errors in the log file?
Also, do Object Viewer / Route Viewer work?

I'll try a Ubuntu VM in a minute, but that ought to be the same as Debian....

leezer3:
Try the new build in a few minutes. Accidentally seem to have been converting the filename to lowercase in the GIF decoder.

ginga81:
I tested the newest nightly build. 
It id very nice animation moving!
This is the wonderful!

PR #488, opened 2020-05-29
Change: Refactor car sound playback

leezer3:
This refactors car sound playback to use a method call within the sound.

In general, it's much simpler code than the previous, and should probably be merged prior to https://github.com/leezer3/OpenBVE/tree/TrainManager (this would allow the removal of a bunch of stored references)

### Potential issues:
* Each CarSound now stores it's car reference internally. Are these all correct?
* No immediate method to change which car a CarSound plays from. (Introduced as an API playback function a while back, this might need redesign?)
* The emergency brake sound appears to have a bug / undefined piece of behaviour- By the looks of things, it was intended to play once (simulataneously) positioned for each car in the train, but either by us or Michelle, actually only played in the driver car. Check when this was introduced!
* null checks moved to within the sound play call. Have any been missed?

### DIFF
--- a/source/OpenBVE/Audio/Sounds.Update.cs
+++ b/source/OpenBVE/Audio/Sounds.Update.cs
@@ -166,8 +166,9 @@ namespace OpenBve
 						 * For non-looping sounds, check if the sound is still playing.
 						 * */
 						gain = (gain - GainThreshold) / (1.0 - GainThreshold);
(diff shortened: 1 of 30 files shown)


leezer3:
Mostly implemented in the commit above.
For the minute, still allowing the parent car reference to be passed. This is beacuse car sounds aren't loaded properly for all cars in BVE2 / BVE4 sound configs.

PR #493, opened 2020-06-15
Fix: a crash on Intel iGPU

s520:
This PR includes the following changes.
- Fix: a crash on Intel iGPU since v1.7.1.4
  - This is probably because the IBO is not deleted.
- Fix: the issue that WAVE file cannot be played since v1.7.0.0 when the data chunk size is incorrect
  - This problem occurs in Klaxon0.wav of mackoy's 113 series.
- Change: Avoid reparsing the same object file
  - This change speeds up the loading of route files and reduces the memory usage.

---

(Reference information) Table of load times and maximum memory usage
|          | Loading time | Maxium used memory                                   |
| -------- | ------------ | ---------------------------------------------------- |
| v1.7.1.6 | 1min 35s     | \>= 2GB (Crash in less than a quarter of the route.) |
| This PR  | 46s          | 1072MB                                               |

Measurement environment
- Hardware: Intel Core i5-10210U, 8GB memory
- Route: ginga81's Tohoku shinkansen (とーほぐ新幹線仙台以北v0.06.01/東北新幹線はやぶさ1_fromSendai_v0.07 (summer).csv)
- Train: F81 tec200's E5 series


source/OpenBveApi/Objects/ObjectTypes/AnimatedWorldObject.cs
```diff
@@ -1,20 +1,25 @@
...
 	{
-		/// <summary>Holds a reference to the host application</summary>
-		private readonly Hosts.HostInterface currentHost;
 		/// <summary>The signalling section the object refers to (Only relevant for objects placed using Track.Sig</summary>
 		public int SectionIndex;
 
 		/// <summary>Creates a new Animated World Object</summary>
 		/// <param name="Host">The host application</param>
-		public AnimatedWorldObject(Hosts.HostInterface Host)
+		public AnimatedWorldObject(HostInterface Host) : base(Host)
```
leezer3, line 15: The Hosts.HostInterface was actually deliberate and is likely to be required. (Mono compilation sometimes doesn't seem to like it otherwise)

See what CI does.

leezer3:
Secondary thought:
If we're caching static objects, you probably don't even need to clone them most of the time.

Only places I can think of that require cloning:
* Cracks / forms (objects may be transformed)
* Animated object using the TextureShiftFunction

s520:
It's confusing because StaticObject has a function that applies the transformation to itself rather than a clone. (Apply***)
I think it's better to remove the function that transforms the instance itself and replace it with a function that transforms the clone instead.
That way, we can pass the cache as-is without cloning it.

source/OpenBveApi/Objects/ObjectTypes/AnimatedWorldObject.cs
s520: I did not know that.
As far as I can see from the CI results, the build seems to be successful, but I will send this part back.

leezer3:
https://github.com/leezer3/OpenBVE/blob/master/source/OpenBveApi/Objects/ObjectTypes/StaticObject.cs#L457

This function?
Doesn't seem to be being used. Looks to be left over from the original GL1.2 renderer, where every static object was transformed by the world data.

Should be able to dump it.

s520:
It is also so, but it is a function our part of the following that I wanted to say.
https://github.com/leezer3/OpenBVE/blob/e0ffa49a1be3e00a06bd3a9a81fa734b2b507d1d/source/OpenBveApi/Objects/ObjectTypes/StaticObject.cs#L235-L453

In particular, ApplyScale is also used to reverse Extensions.cfg.

leezer3:
Ah, I see what you mean.

The main other isuse with these is that they're used in the CSV / B3D object parser to handle a ScaleAll, RotateAll command.

s520:
There are also instances where the Animated file parser may transform the instance directly.
https://github.com/leezer3/OpenBVE/blob/e0ffa49a1be3e00a06bd3a9a81fa734b2b507d1d/source/Plugins/Object.Animated/Plugin.Parser.cs#L1413-L1429

The StaticObject read may be transformed later like this way.
So it's complicated and I decided to clone the cached object.
Even though I'm doing a lot of wasted cloning, the reduction in load times and memory usage is clearly effective.

source/OpenBveApi/Objects/ObjectTypes/AnimatedWorldObject.cs
leezer3: If it builds. don't worry about it I'd say (Although it's repeated in a bunch of places)

I think it was the original makefile / MCS from memory.

s520: OK.
I fixed it.

s520: Oops, I found a missing fix.

s520: I think it has been fixed.

leezer3:
Squashed and merged.

I'll dump the unused function in a minute, and we can think about reshaping the transforms later.

We could possibly use a getter with no setter in order to throw and hence find when 'wrong' transforms are done, but that always has the risk of missing something.

PR #494, opened 2020-06-20
Fix:compile error with mono-complete package

ginga81:
At
https://www.mono-project.com/download/stable/#download-lin-ubuntu
This installed filename at Ubuntu is
/usr/lib/mono/msbuild/15.0/bin/Microsoft.CSharp.targets
So, compile error causes.


leezer3:
Reasonable enough.

No idea why Travis has never picked this one up, but it just goes to show that casing in csproj files is a mess.

PR #496, opened 2020-06-24
Update: NuGet packages

s520:
This PR update NuGet packages.
However, it does not update NLayer and NLayer.NAudioSupport.
Because they have dropped support for .NET Framework 3.5.
Using the latest version below .NET Framework 4.7.2 will generate a large number of dll files.

Also, this PR fix the error about the route plugin in ObjectViewer and TrainEditor2.
They don't filter the route plugins, so we get an error on startup.


leezer3:
Thanks!

I'll take a proper look at this tomorrow.

Issue #507, opened 2020-08-11
New ObjectViewer will not free memory after reload/object deletion


LubuntuKHYT:
### Description

ObjectViewer after v1.7.0.3 would not free video memory after object deletion/reload
of object. If the VRAM is about to be full and you keep reloading the object(maybe 3-5 more times depending on the object), there will be white texture. Clearing the object also won't free the video memory(unless you close the window). This does not happen before the new render update.
### Reproduction

Load any object(Train's extension.cfg is more effective, take MTR SP1900 EMU for example) in ObjectViewer above 1.7.0.3. On reload, you will see the memory just bump up instead of staying on the same VRAM level like before(You can check on W10 taskmgr/GPU-Z). When it's about to max out, just reload like 3-6 more times. Then you will see white texture, along with error messages:

No plugin found that is capable of loading texture E:\Path\MTR West Rail Line SP1900 EMU\exterior\111.png
Plugin Texture.BmpGifJpegPngTiff.dll raised the following exception at LoadTexture:Out of memory.

Using the Old (GL 1.2) and the New (GL 3.0) make no difference on this issue.

## Train

[Google Drive Link](https://drive.google.com/file/d/0B5mnFfKjPZGoNklqN1VTcE8xRHM/edit)

## Logs

Not in-game problem, attached ObjectViewer report instead. Problem starting from line 10.
[log.txt](https://github.com/leezer3/OpenBVE/files/5055052/log.txt)

### Related information 

* OS: Windows 7 Build 7601 SP1 (Happens too on my previous Windows 10 1909/2004 install)
* Control Method: Keyboard

![GPU-Z_2020-08-11_14-07-51](https://user-images.githubusercontent.com/28094366/89863152-1c29a200-dbdc-11ea-968d-82cc8a0e5d77.png)

leezer3:
Been meaning to look into this.

We seem to have a memory leak in that something is holding onto a reference to the VBOs used for GL3.

Investigating at the minute.

leezer3:
https://vps.bvecornwall.co.uk/OpenBVE/Builds/OpenBVE-2020-08-19.zip

Should now be fixed from the current build onwards.

Issue #305, opened 2019-01-13
openBVE Licence & Associated


leezer3:
Changing the licence to something more widely recognised has again been raised in #304 
This is a specific issue to discuss the mechanics and roadblocks involved.

Now, fundamentally I'm by no means opposed to this.
The original driver for openBVE's current public domain licence was Michelle, and indeed she was actively opposed to any licence at all:
```
   My attitude differs from so called "free software" licenses. I try to
   explain:
 .
   With proprietary software, the developers usually try to restrict what
   you are allowed to do with the software to the uttermost extreme that
   is legally possible. This falls under the category of copyright.
 .
   Then there is something that is sometimes called green copyright. So
   called free software falls in this category. As with proprietary
   software, a license is used to tell users what they are allowed to do
   and what not. The difference to proprietary software is that so called
   free software licenses are usually much more permissive, yet the
   author decides what you are allowed to do and what not. So called free
   software is usually attributed to derive from "freedom" instead of
   "free of charge". Still, if the author employs a license telling other
   people what they can or cannot do, this inherently cannot have
   anything to do with freedom.
 .
   Then there is something called public domain, which I consider to be
   an inherently good thing as everyone has eventually the same rights as
   the author. However, releasing a work into the public domain is not
   legally meaningful in many countries. Also, releasing a work into the
   public domain (if possible) depends on the generosity of the author,
   and this decision is made by the author.
 .
   My attitude is even one step further: I am opposed to copyright, thus
   I marked the game with the legally not meaningful phrase of
   "anti-copyright". It is legally meaningless, because in every country
   that has some form of copyright, it would require these laws to be
   abandoned. Still, why do I have this attitude?
 .
   With a license, I would be telling other people what they can or
   cannot do. I am not such a person. I will not make any silly
   restrictions on how you can use this program, I will not tell you to
   put my name on any derived work, I will never ever give "permission"
   for you modify or redistribute the software, because I don't think
   that it is up to me telling you what you can or can't do. You should
   make this decision for yourself.
 .
   I hope this briefly explained the situation.
```
( https://metadata.ftp-master.debian.org/changelogs/main/o/openbve-data/openbve-data_1.4.0.5+dfsg-3_copyright )

Obviously, at this point, she's been gone for ~10 years, and I have had no contact.
The other contributors to the original openBVE code were:
* Anthony Bowden
* Odaykufan (Jens Rügenhagen)
* Paul Sladen

[...]

PR #508, opened 2020-08-11
Change: Attempt to solve memory leak with VAOs

leezer3:
This implements IDisposable on various classes, and changes a lot of the cleanup code in order to try and solve https://github.com/leezer3/OpenBVE/issues/507

### DIFF
--- a/source/LibRender2/BaseRenderer.cs
+++ b/source/LibRender2/BaseRenderer.cs
@@ -232,10 +232,25 @@ namespace LibRender2
 			Loading = new Loading(this);
 			Keys = new Keys(this);
 			MotionBlur = new MotionBlur(this);
+			if (StaticObjectStates == null)
+			{
+				StaticObjectStates = new List<ObjectState>();
+			}
 
-			StaticObjectStates = new List<ObjectState>();
-			DynamicObjectStates = new List<ObjectState>();
-			VisibleObjects = new VisibleObjectLibrary(currentHost, Camera, currentOptions, this);
+			if (DynamicObjectStates == null)
+			{
+				DynamicObjectStates = new List<ObjectState>();
+			}
+			
+			if (VisibleObjects == null)
+			{
+				VisibleObjects = new VisibleObjectLibrary(currentHost, Camera, currentOptions, this);
+			}
+			else
+			{
+				VisibleObjects.Clear();
+			}
+			
 
 			GL.ClearColor(0.67f, 0.67f, 0.67f, 1.0f);
 			GL.Clear(ClearBufferMask.ColorBufferBit | ClearBufferMask.DepthBufferBit);
@@ -326,6 +341,18 @@ namespace LibRender2
 		public void Reset()
 		{
 			Initialize(currentHost, currentOptions);
+			for (int i = 0; i < StaticObjectStates.Count; i++)
+			{
+				ObjectState os = StaticObjectStates.ElementAt(i);
+				os.Dispose();
+			}
+			StaticObjectStates.Clear();
+			for (int i = 0; i < DynamicObjectStates.Count; i++)
+			{
+				ObjectState os = DynamicObjectStates.ElementAt(i);
+				os.Dispose();
+			}
+			DynamicObjectStates.Clear();
 		}
 
 		public int CreateStaticObject(StaticObject Prototype, Vector3 Position, Transformation BaseTransformation, Transformation AuxTransformation, bool AccurateObjectDisposal, double AccurateObjectDisposalZOffset, double StartingDistance, double EndingDistance, double BlockLength, double TrackPosition, double Brightness)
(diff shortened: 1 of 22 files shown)


leezer3:
The GC.Collect are probably un-necessary.

### Other things to do:
This also needs to be added to route viewer.
Main program needs considerable testing, this has only really been tested in Object Viewer. Need to check that this won't make things crash (e.g. when jumping around etc)

leezer3:
This now appears stable-
Reload a file in Object Viewer or Route Viewer, and anfter a few minutes the memory usage will settle back to what it was originally.

One thing I'm not so happy about is that this entire implementation specifically relies on the fact that when something is removed from a List<T> the Dispose() method is not called.
If this changed unexpectedly (I haven't gone looking into the C# roadmap), then things would fall over badly.

cc @s520 @ginga81

s520:
The current implementation implements IDisposable but not the finalizer.
So I think we need to call Dispose somehow after it is removed from the list.

I thought of writing a finalizer according to [the C# guide](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose).
This is because the handle obtained from OpenTK is an unmanaged resource.

However, when I tried to delete the handle in the finalizer, I got an AccessViolationException.
So this idea may be wrong in the first place...

leezer3:
Things are a little complicated....
Now that we're using a single prototype instance with multiple references, this makes the standard dispose / finalize pattern a little complicated (As **Dispose()** can be called multiple times, and takes no account of multiple instances of the reference, and we have no explicit control over where finalize is called & thus GL contexts)
The code I've implimented in this PR uses the standard dispose pattern, but relies on the fact that when we remove something from a **List<T> ** the dispose method is not actually called- https://stackoverflow.com/questions/16181693/removeatx-does-it-dispose-the-element-x
Instead, I've added the DisposeAndClear() extension method, which itinerates down the entire chain of references, disposing *everything* as it goes.

To my mind, this is a little dangerous (and only works on a reload) as it fundamentally assumes that there's no further use for anything. and instead aims to clear all used memory.

shooking:
Hi. Your code is open source right? So some commercial static analysis
firms let you run their tools. If you like I can try to run some free ones?
I do it for my day job and it can find some obscure leaks / potential race .
conditions.

leezer3:
Correct, we're open source, although the licence issue is a little complicated due to previous baggage:
https://github.com/leezer3/OpenBVE/issues/305

Any contributions are welcomed :)

The biggest leak itself (well techinically the visible sympton) is relatively obvious: 
https://github.com/leezer3/OpenBVE/blob/master/source/OpenBveApi/Objects/Mesh.cs#L19
As far as I can see from digging with DotMemory and the VS2010 memory viewer (and I'm no professional) a reference to something is being held somehow by the following list after we issue the Clear() command, which in turn prevents the memory from being cleared:
https://github.com/leezer3/OpenBVE/blob/master/source/LibRender2/Objects/ObjectLibrary.cs#L23
The changes in this PR add an explicit dispose method to all components of the class-chain called by an extension method, which seems to work around the problem, but as noted above I don't like this approach.

LabRatAndy:
Could I offer a suggestion. 

All VAOs appear to be added to the static list list disposeable, by the constructor. 
[https://github.com/leezer3/OpenBVE/blob/18a7f1e7ca54be63c3e807fe65f4bc686c37fbd4/source/LibRender2/openGL/VertexArrayObject.cs#L19](url)
[https://github.com/leezer3/OpenBVE/blob/18a7f1e7ca54be63c3e807fe65f4bc686c37fbd4/source/LibRender2/openGL/VertexArrayObject.cs#L36](url)
However, the dipose method doesn't remove the VAO from this list  and so there would still be a reference to the VAO held in this list even though the underlying openGL resources have been cleaned up by the Dipose method. The GC would still think that this dispoesed VAO is still being used due to it being held in the static disposable list and therefore not collect it when it next runs.

leezer3:
All suggestions greatly appreciated 👍 

Will take a closer look tomorrow at where you're linking to (doesn't work at all on the phone :/ )

leezer3:
Good call.

That list gets used here when the renderer cleans up on close:
https://github.com/leezer3/OpenBVE/blob/master/source/LibRender2/BaseRenderer.cs#L260

Looks like we need to clear it at the same time as a reload. Also going to need to clear the FBO list.

NB:
Before the changes in this PR, the **Dispose()** method for the VAO wasn't getting called at all.
Might be better moving it into a destructor.

leezer3:
So, removing those lists helps with *some* of the problem-
This allows the GC to reclaim the system memory.
What it doesn't do however is free the GPU memory.
https://github.com/leezer3/OpenBVE/pull/508#issuecomment-674188776
I also tried deleting the resources in the finalizer, which as noted above crashes with an access violation exception. I think this is likely being caused by a lack of GL context in the finalizer thread- Will experiment a little.

leezer3:
Much better.

I *think* the other PR now solves this properly by using the destructor and an enqueued list on each render pass, can anyone else please take a look and confirm before I merge?

PR #511, opened 2020-08-18
Change: Different approach to memory leak with VAOs

leezer3:
This takes a different approach entirely to https://github.com/leezer3/OpenBVE/pull/508

If the resources have not explicitly been released, the destructor now enqueues them to be deleted in the next render pass thread.

Appears to solve things properly.
cc @s520 @LabRatAndy

### DIFF
--- a/source/LibRender2/BaseRenderer.cs
+++ b/source/LibRender2/BaseRenderer.cs
@@ -185,6 +185,14 @@ namespace LibRender2
 			set;
 		}
 
+		/*
+		 * List of VBO and IBO to delete on the next frame pass
+		 * This needs to be done here as opposed to in the finalizer
+		 */
+		internal static readonly List<int> vaoToDelete = new List<int>();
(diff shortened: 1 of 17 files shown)


LabRatAndy:
I will have a look later on tonight

s520:
Except for the #512 issue, it looks like there is nothing wrong.
The memory leak does not occur even after performing the procedure of #507 on Intel iGPU.

leezer3:
Thanks:)

Did you mean to remove the texture unload, or was that a mistake?
Haven't checked what's going on there.

s520:
I put them together because Renderer.TextureManager.UnloadAllTextures() was called after Renderer.Reset().
Is it better to incorporate it in Game.Reset?
If it is inappropriate in the first place, put it back.

leezer3:
Ah, missed that you'd moved it :)
That should be right, we just need to purge the texture cache somewhere to account for any updates on reload.

PR #518, opened 2020-08-27
Add: cross-fading to switch between day and night textures

s520:
This PR adds cross-fading mode to CSV objects.

Cross-fading mode changes the transparency of DaytimeTexture as opposed to the transparency of NighttimeTexture.
This behavior is different from the conventional one, so it is disabled by default and can be enabled by command.

The command format is as follows:
`EnableCrossfading, value` (value: true or false)

Closed #425

PR #519, opened 2020-08-27
Crossfading improvements

leezer3:
Slightly improved version of https://github.com/leezer3/OpenBVE/pull/518

This moves the CrossFadeTexture bool to the MaterialFlags enum, and uses the same enum in the parsers, as opposed to a bunch of bools which then set the flags.

Not done any major testing, but appears to work

cc @s520


s520:
I think it's a good idea.
However, I fixed it because the bit flag was not working. (#520)

leezer3:
That would help, changed the old but not the new :/

Issue #498, opened 2020-06-27
parts of Panel2.cfg object blinking randomly from 1.7.1.7


ginga81:
### Description

The brightness changes unintentionally according to the surrounding brightness.
At Series-E6's digital speed bar graph is blinking.
When a train is running, some of panel's brightness, for example the right side's digital speed meter is  is must bright always, but effected from Track.Brightness.

For example, Series-N700A's 7-segments digital number object is works correctly.
 In other hands, LCD digital speed meter, bar graph, digital watch and more's brightness is changes unintentionally to dark.
And about the cause, F81_tec200 found that if **DaytimeImage and NighttimeImage is the same file**, **only load DaytimeImage**.
For example, one of object the bar graph what is brightness changes unintentionally, the file name is bg of a png file.
This object is using the same file day and night.
In other hands, 7-segments digital number object's file is digi_num and digi_num_n, which is different file using properly.

### Reproduction

Route is any, but must set Track.Brightness for checking
Train:
Series-E6
https://youtu.be/yRBr-Dsa1U4

Series-N700A
https://www.youtube.com/watch?v=tBweD9T2b3Y

## Route

Any, but must set Track.Brightness for checking

## Train

Series-E6
https://youtu.be/yRBr-Dsa1U4

Series-N700A
https://www.youtube.com/watch?v=tBweD9T2b3Y

## Logs

This issue is no crash. So, no logs.

### Related information 

Windows 10

My PC's SPEC:

OS: Ubuntu18.04LTS(64bit)
Memory:31.4 GiB
CPU:AMD® Ryzen 7 2700x eight-core processor × 16
Graphic board:Radeon RX 590 Series (POLARIS10, DRM 3.33.0, 5.3.0-51-generic, LLVM 9.0.0)

leezer3:
Should be fixed by https://github.com/leezer3/OpenBVE/commit/1325bd0ec76758cc5ae4a6a2b631155e563ae48d

**Explanation:**
The 2D cab is rendered with lighting disabled, but still uses some material properties, which were only being set by the new renderer with lighting enabled.

ginga81:
At OpenBVE-2020-07-02.zip, blinking issue is fixed.
But, still not fix these area.
These originarities are always bright, but there are interlocking with brightness.
![E6-brightness1](https://user-images.githubusercontent.com/22734699/86378484-af5de680-bcc4-11ea-944a-15fba738a0ad.jpg)
[...]

PR #523, opened 2020-08-31
Fix: issue where panel elements with NighttimeImage defined are rendered with lighting enabled

s520:
In BVE4 and 5, panel elements with NighttimeImage defined are rendered with lighting disabled. (#498)

Currently, it is not possible to specify an emissive color for the panel, so the author cannot substitute this effect.
If we want to address this issue like that of a CSV object, we should add an emissive color setting.
However, this effect is the same from BVE4, so I don't think it needs to be changed.

### DIFF
--- a/source/LibRender2/BaseRenderer.cs
+++ b/source/LibRender2/BaseRenderer.cs
@@ -1016,7 +1016,7 @@ namespace LibRender2
 				Shader.SetBrightness(factor);
 
 				float alphaFactor = distanceFactor;
-				if ((material.Flags & MaterialFlags.CrossFadeTexture) != 0)
+				if (material.NighttimeTexture != null && (material.Flags & MaterialFlags.CrossFadeTexture) != 0)
 				{
 					alphaFactor *= 1.0f - blendFactor;
 				}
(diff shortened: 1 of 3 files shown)


leezer3:
I agree it doesn't need changing for a 2D cab.

Something else I've thought of as a result of this:
A panel.cfg cab (BVE2) should possibly be rendered with lighting disabled too. Before alternate night-time images were possible, it was common for there to be two versions of any given train in a day / night variation, & the **Track.Brightness** command didn't exist & so the whole thing remained at a constant brightness.
Doing this would make the lighting / overall effect 'original' but on the flip side, I'm not sure that it's necessary and in most cases the revised effect looks decent. 
It only really has a major effect if you try to use a BVE2 nightime cab on a route which uses brightness commands. Could add a hack to control this?

Any thoughts on that one?

s520:
I don't think the program can tell if the cab selected by the user is a nighttime cab.
So why not have the user input whether it's a nighttime cab or not when the user selects a train?
This checkbox is clickable only when panel.cfg cab is selected.
The state of this checkbox toggles whether lighting is enabled when drawing the cab.

leezer3:
Correct, not easily autodetectable. (Although for English content could key off night in the folder name & a panel.cfg)

I was thinking more of a hardcoded list, either by adding it to a new database, or by hardcoding them in the parser.
The issue with problem with adding a checkbox is getting it found by the average user

s520:
If full automatic detection is difficult, we should leave the user with a choice.
As an aid, the program will guess from the folder name and change the default state of the checkbox.
Or, as you say, it will check against the database and change the default state.

If the program assists to some extent, the average user should be fine with the defaults.

PR #525, opened 2020-09-15
Changes to Vertex Colors & Emissive Hack

leezer3:
https://github.com/leezer3/OpenBVE/commit/82e43116648c53933d8178c0e0a9f9fa793c8048

Reverts this, and hopefully fixes things a little better this time around.
I've also hidden the internal drawing routines for the cube (As we've created a specific class for it, IMHO it should only have the single draw method)

cc @s520

### DIFF
--- a/assets/Shaders/default.frag
+++ b/assets/Shaders/default.frag
@@ -3,10 +3,11 @@
 in vec4 oViewPos;
 in vec2 oUv;
 in vec4 oColor;
-in vec4 lightResult;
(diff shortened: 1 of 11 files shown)


leezer3:
NB.
This also adds a minor 'improvement' to the CSV route detection routine as follows:
The first 30 lines are read, and checked for the presence of **MeshBuilder** (case invariant).
If this 'special' string is present, we assume the file to be an object, not a route.

PR #526, opened 2020-09-16
Fix: emissive hack

s520:
If you combine textures with emission colors, the new renderer will make the colors brighter.

- Old renderer
Emissive_OR

- New renderer (Emissive branch #525)
Emissive_NR

- New renderer (This PR)
Emissive_NR_Fix

I also removed and cleaned up unused variables in the shader.

### DIFF
--- a/assets/Shaders/default.frag
+++ b/assets/Shaders/default.frag
@@ -19,7 +19,11 @@ uniform bool uFogIsLinear;
 
 void main(void)
 {
-	vec4 finalColor = uIsTexture ? texture2D(uTexture, oUv) : vec4(oColor.rgb, 1.0);
+	vec4 finalColor = vec4(oColor.rgb, 1.0);
+	if(uIsTexture)
+	{
+		finalColor *= texture2D(uTexture, oUv);
+	}
 	if((uMaterialFlags & 1) == 0 && (uMaterialFlags & 4) == 0)
 	{
 		//Material is not emissive and lighting is enabled, so multiply by brightness
--- a/assets/Shaders/default.frag
+++ b/assets/Shaders/default.frag
@@ -3,10 +3,11 @@
 in vec4 oViewPos;
 in vec2 oUv;
 in vec4 oColor;
-in vec4 lightResult;
+in vec4 oLightResult;
 
 uniform bool uIsTexture;
 uniform sampler2D uTexture;
+uniform int uMaterialFlags;
 uniform float uBrightness;
 uniform float uOpacity;
 uniform bool uIsFog;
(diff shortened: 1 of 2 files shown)


leezer3:
Thanks!

Hopefully we're slowly getting somewhere, and might get this properly right at some stage.....
I'll try testing this properly later.

Also requires this reverting:
https://github.com/leezer3/OpenBVE/commit/82e43116648c53933d8178c0e0a9f9fa793c8048

---

I think at this point, we probably need a list of known issues / edge cases that have been fixed in the shader, otherwise things get confused and broken again.

assets/Shaders/default.frag
```diff
@@ -15,36 +16,31 @@ uniform float uFogEnd;
...
-uniform int uMaterialFlags;
 
 void main(void)
 {
-	vec4 textureColor = vec4(oColor);
-	if(uIsTexture)
-	{
-		textureColor *= texture2D(uTexture, oUv);
-	}
+	vec4 finalColor = uIsTexture ? texture2D(uTexture, oUv) : vec4(oColor.rgb, 1.0);
```
leezer3, line 22: Minor:
You're assuming colored vertices disable texturing.

At the minute, I don't think there's anything with colored vertices in the wild (only format that supports them IIRC is the X format with either of the new parsers), but is this a correct assumption?
If I'm reading the DirectX specification correctly ( https://docs.microsoft.com/en-us/windows/win32/direct3d9/per-vertex-color-state ) then this overrides the material properties ambient etc. colors, and I would assume texturing still works in this case.

s520: Oops. I fixed it.

s520: wait a minute.
I need to fix it a little more.

s520:
In older renderers, using two textures to specify an emission color disables lighting.
(It doesn't matter if the two textures are the same.)
This behavior is correct when the hack is enabled, but it is the same when the hack is disabled.
Lighting should be enabled if the hack is disabled.
Is this a known issue?

I looked at examples below in the ObjectViewer on the master branch.

### e.g. 1
```
LoadTexture, Daytime-base.png, Ref-night.png
SetEmissiveColor, 0, 255, 0
```
- Old renderer
DayNight_OR

- New renderer
DayNight_NR

### e.g. 2
```
LoadTexture, Daytime-base.png
SetEmissiveColor, 0, 255, 0
```
- Old renderer
Day_OR

- New renderer
Day_NR

### e.g. 3
```
SetColor, 0, 0, 255,
SetEmissiveColor, 0, 255, 0
```
- Old renderer
Color_OR

- New renderer
Color_NR

leezer3:
I think you've probably found another one....

Logically, I'd expect the emissive color to be respected.

s520:
I haven't been able to find where it's coming from.
Well, it seems better to think about it separately from this PR.

Therefore, there are no changes to this PR.

leezer3:
Merged this into the emissive branch.

I can't see anything immediately broken, still testing before I merge into the main branch.
Have also added a command-line option to the viewers for good measure to make testing all of this a little easier:
```
/enablehacks
```

PR #527, opened 2020-09-18
Change: Add some prequisite checking in the makefile

leezer3:
Adds a little prequisite checking to our makefile
https://bveworldwide.forumotion.com/t1975-linux-compilation-not-registered-task-move#20024

Might be worth thinking about CAKE or FAKE possibly.

cc @s520

### DIFF
--- a/makefile
+++ b/makefile
@@ -3,6 +3,9 @@ MSBUILD := msbuild
 MIN_MONO_VERSION:= "5.18.0"
 MONO_VERSION:= $(shell mono --version | awk '/version/ { print $$5 }')
 MONO_MEETS_MINVERSION := $(shell expr "$(MONO_VERSION)" ">=" "$(MIN_MONO_VERSION)")
+MIN_NUGET_VERSION:= "2.16.0"
+NUGET_VERSION:= $(shell nuget help | awk '/Version\:/ { print $$3 }')
+NUGET_MEETS_MINVERSION := $(shell expr "$(NUGET_VERSION)" ">=" "$(MIN_NUGET_VERSION)")
 
 # Directories
 DEBUG_DIR   := bin_debug
@@ -117,6 +120,12 @@ prequisite-check:
  $(info Checking for nuget....)
  ifeq (, $(shell which nuget))
  $(error "nuget does not appear to be installed on this system.")
+ endif
+ $(info nuget Version $(NUGET_VERSION) found.)
+ ifeq "$(NUGET_MEETS_MINVERSION)" "1"
+ #Nothing
+ else
+ $(error "OpenBVE requires a minimum nuget version of 2.16")
  endif
  $(info Attempting to restore nuget packages (This may take a while)....)
 	
--- a/makefile
+++ b/makefile
@@ -2,10 +2,9 @@
 MSBUILD := msbuild
 MIN_MONO_VERSION:= "5.18.0"
 MONO_VERSION:= $(shell mono --version | awk '/version/ { print $$5 }')
-MONO_MEETS_MINVERSION := $(shell expr "$(MONO_VERSION)" ">=" "$(MIN_MONO_VERSION)")
 MIN_NUGET_VERSION:= "2.16.0"
-NUGET_VERSION:= $(shell nuget help | awk '/Version\:/ { print $$3 }')
-NUGET_MEETS_MINVERSION := $(shell expr "$(NUGET_VERSION)" ">=" "$(MIN_NUGET_VERSION)")
+NUGET_VERSION:= $(shell nuget help 2> /dev/null | awk '/Version:/ { print $$3; exit 0}')
+GreaterVersion = $(shell printf '%s\n' $(1) $(2) | sort -rV | head -n 1)
 
 # Directories
 DEBUG_DIR   := bin_debug
@@ -111,7 +110,7 @@ prequisite-check:
  $(error Mono does not appear to be installed on this system.)
  endif
  $(info Mono Version $(MONO_VERSION) found.)
- ifeq "$(MONO_MEETS_MINVERSION)" "1"
+ ifeq "$(call GreaterVersion, $(MONO_VERSION), $(MIN_MONO_VERSION))" "$(MONO_VERSION)"
  #Nothing
  else
  $(error OpenBVE requires a minimum Mono version of 5.20)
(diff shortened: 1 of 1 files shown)


s520:
We should also add a version check for the NuGet client.
If your client version is less than 2.12, you will get the following error:
```
'NAudio' already has a dependency defined for 'Microsoft.Win32.Registry'.
'SharpCompress' already has a dependency defined for 'System.Memory'.
The 'System.Buffers 4.5.1' package requires NuGet client version '2.12' or above, but the current NuGet version is '2.8.7.0'.
The 'System.Memory 4.5.4' package requires NuGet client version '2.12' or above, but the current NuGet version is '2.8.7.0'.
The 'System.Runtime.CompilerServices.Unsafe 4.7.1' package requires NuGet client version '2.12' or above, but the current NuGet version is '2.8.7.0'.
The 'System.ValueTuple 4.5.0' package requires NuGet client version '2.12' or above, but the current NuGet version is '2.8.7.0'.
'Ude.NetStandard' already has a dependency defined for 'NETStandard.Library'.
The 'System.ComponentModel.Annotations 4.7.0' package requires NuGet client version '2.12' or above, but the current NuGet version is '2.8.7.0'.
The 'System.Reactive 4.4.1' package requires NuGet client version '2.12' or above, but the current NuGet version is '2.8.7.0'.
'System.Threading.Tasks.Extensions' already has a dependency defined for 'System.Runtime.CompilerServices.Unsafe'.
```
Some distributions (e.g. openSUSE) have not updated the client.

I will look at the other parts after tomorrow.

leezer3:
Done.

(Concept worked OK with the first batch, so I've not bothered triggering a build)

Possibly wants an internet connection check too, but if we're not careful this ends up turning into a monster of a script.

makefile
```diff
@@ -1,13 +1,11 @@
...
-ifndef HAS_MSBUILD
-	MSBUILD := xbuild
-endif
+# Version checking
+MSBUILD := msbuild
+MIN_MONO_VERSION:= "5.18.0"
+MONO_VERSION:= $(shell mono --version | awk '/version/ { print $$5 }')
+MONO_MEETS_MINVERSION := $(shell expr "$(MONO_VERSION)" ">=" "$(MIN_MONO_VERSION)")
+MIN_NUGET_VERSION:= "2.16.0"
+NUGET_VERSION:= $(shell nuget help | awk '/Version\:/ { print $$3 }')
```
s520, line 7: At least below 3.4.4, the awk command will not exit.

Also, I don't think you need to escape this colon symbol.
```
awk: command line:1: warning: regexp escape sequence `\:' is not a known regexp operator
```

leezer3: 3.4.4 of nuget do you mean?

I think this would work to add the timeout in that case:
```
$(shell timeout 2 nuget help | awk '/Version:/ { print $$3 }')
```
Need to test with an older version of nuget though.

s520: That's right.
That doesn't seem to be a problem for timeouts.
However, the versions cannot be compared.
Probably the same for mono.

s520: I sent a PR with the above contents fixed.

leezer3: Thanks.

Will test later. Interestingly on hhe Windows Subsystem for Linux (Buster) my solution works, so this may be distribution / awk implementation dependant somewhere. (Deian stuck on GNU awk?)

s520: I have confirmed fedora 32 and openSUSE 15.2.
fedora is a real machine and openSUSE is a virtual machine.
I haven't confirmed it in debian yet.

s520: I have confirmed that it is the same for debian sid on the virtual machine.

s520: I may have misled you.
The versions I have confirmed are summarized in the table.
The result is the same regardless of the type of distribution and whether it is a real machine.
| Version of nuget.exe                                         | Result |
| ------------------------------------------------------------ | ------ |
| [2.8.6](https://dist.nuget.org/win-x86-commandline/v2.8.6/nuget.exe) | stuck  |
| [3.3.0](https://dist.nuget.org/win-x86-commandline/v3.3.0/nuget.exe) | stuck  |
| [3.4.4](https://dist.nuget.org/win-x86-commandline/v3.4.4/nuget.exe) | finish |
| [3.5.0](https://dist.nuget.org/win-x86-commandline/v3.5.0/nuget.exe) | finish |
| [4.1.0](https://dist.nuget.org/win-x86-commandline/v4.1.0/nuget.exe) | finish |
| [4.9.4](https://dist.nuget.org/win-x86-commandline/v4.9.4/nuget.exe) | finish |

leezer3:
```
nuget help 2> /dev/null | awk '/Version:/ { print $3; exit 0}'
```

Done a little more playing today.
I think that works on all nuget versions by deliberately exiting after the first match (The older Nuget help text appears to be missing the terminator character), and redirecting any stderr messages to /dev/null to stop AWK complaining about attempting to write to an unknown file.

Am I right in thinking that the version command isn't on RHEL, hence why you've changed to using sort -rV ?

s520: That method also seems to solve the problem.

Next is the version comparison.
Are you referring to why I changed to using sort-rV?
With the current method, even if the version can be obtained correctly, the version comparison is not working.
```
user@localhost:~/OpenBVE> make
Checking for prequisite system libraries.
Checking for Mono....
Mono Version 6.8.0.105 found.
Checking for MSBuild....
Checking for nuget....
nuget Version 2.8.7.0 found.
Attempting to restore nuget packages (This may take a while)....
nuget restore OpenBVE.sln
...
```

leezer3: Fair enough.

timeout doesn't appear to be on OS-X by default without homebrew, hence why finding a method to get rid of it.

s520: OS X doesn't seem to support `sort -V`. (https://ss64.com/osx/sort.html)
Therefore, we have to rewrite it with the following command:
`sort -t'.' -k1nr -k2nr -k3nr -k4nr`
Can you check it?

leezer3: That works.

(Irritatingly, the OS-X sort command takes -version to provide the version number, not sort by version it seems....)

s520: Sorry.
There was a mistake earlier.
I think this will work...
`sort -t'.' -nr -k1,1 -k2,2 -k3,3 -k4,4`

```
user@localhost:~/OpenBVE> printf '%s\n' 2.8.7.0 2.16.0 | sort -t'.' -k1nr -k2nr -k3nr -k4nr | head -n 1        
2.8.7.0
user@localhost:~/OpenBVE> printf '%s\n' 2.8.7.0 2.16.0 | sort -t'.' -nr -k1,1 -k2,2 -k3,3 -k4,4 | head -n 1
2.16.0
```

leezer3:
```
printf '%s\n' 2.8.7.0 2.16.0 | sort -t. -k 1,1nr -k 2,2nr -k 3,3nr -k 4,4nr | head -n 1
```

https://stackoverflow.com/a/4495368/722627

I *think* that's what we want?

s520: Oh, that's it.
Thanks!

PR #531, opened 2020-09-24
Fix: some issues with the installer

s520:
Also, Appveyor fails to build the tagged commit, so I fixed that setting.
The build process is probably finished before the completion of the compil32.exe used in the script.
This is inferred from the fact that the installer is not recognized as an artifact. (by my some tests)

The artifact settings are for testing purposes, so I'll remove them if you don't need them.


leezer3:
Hang on :/
```
Note that it is still expected to install the OpenAL redistributable provided
by Creative Labs (at http://openal.org/), as that will provide the "router"
OpenAL32.dll that applications talk to, and may provide extra drivers for the
user's system. The DLLs provided here will simply add additional devices for
applications to select from. If you do not wish to use the redistributable,
then rename soft_oal.dll to OpenAL32.dll (note: even the 64-bit DLL should be
named OpenAL32.dll). Just be aware this will prevent other system-installed
OpenAL implementations from working.
```

From the readme.
Let me check whether this works without the Creative installer..

leezer3:
OK, so that still works.
Must only be if you're installing it into the System32 directory and attempting to call it by device name.

Knew there was a reason I hadn't removed it though....

PR #532, opened 2020-09-24
Change: Deploy new release builds to Github also [ci skip]

leezer3:
This is intended to mirror all new builds so that we get the compiled binary artifacts in Github releases.
Needs mirroring to AppVeyor to produce the installer artifact, which since this has hopefully been fixed will think about later.

Minor wondering if we can get the entire site updated automatically by script (as per the translation bot)

https://docs.travis-ci.com/user/deployment/releases/

Not sure we can automatically link to the latest with the current file naming schema, as this would require the same filename:
https://docs.travis-ci.com/user/deployment/releases/

cc @s520


s520:
Site updates may be achieved by monitoring the ReleaseEvent.
https://docs.github.com/en/developers/webhooks-and-events/github-event-types#releaseevent

When a release is released, the bot will mirror its contents and update links and more.

Issue #466, opened 2020-04-26
Transparency trouble with new renderer on 1.7.1.4, loading screen and proccessing bar both not shown only linux


ginga81:
Old renderer is not happen transparency issue.
Loading screen trouble is only on Linux, not happen on Windows.

This trouble happens at my tohoku-shinkansen route.
http://midnightexpginga81.web.fc2.com/touhoku_shinkansen/route_data/Touhogu-Shinkansen-Routedata-fromSendai_v0.06.zip
This screenshots are now creating new version, so difference. but most objects are the same, and also happens.
The cars are all of my released outer-view object, Series-200, 925, E5  and E6.

I tested the daily-builds.
OpenBVE-2020-04-20.zip:OK
OpenBVE-2020-04-21.zip:NG

![new-render](https://user-images.githubusercontent.com/22734699/80294466-12eb0580-87a4-11ea-843b-037ab03f78fd.jpg)
![old-render](https://user-images.githubusercontent.com/22734699/80294468-141c3280-87a4-11ea-860d-4f4e15ed3d94.jpg)

At objectViewer's options_ov.cfg, isUseNewRenderer key change also I tested at 1.7.1.4's.
There are not differense about transparency.
This object that I tested.
Touhogu-Shinkansen-Objectdata-v0.15.zip
https://onedrive.live.com/redir?resid=2E630AFCBBE39699%21159
Object/OpenOBJs/Dike&Wall&Roads/Shinkansen/Pole/PoleR.csv
![new-render-ov](https://user-images.githubusercontent.com/22734699/80298212-e693b100-87c4-11ea-9742-e78ef8d4b2e1.jpg)
![old-render-ov](https://user-images.githubusercontent.com/22734699/80298215-ea273800-87c4-11ea-8cc4-21d2afb9d360.jpg)

Please check these troubles.

leezer3:
Will do.

I suspect something funny in the drivers (default state maybe?) as I did quite a bit of testing on Mac with this and definitely didn't see any glitches :(

s520:
When NewRenderer is enabled, ObjectViewer's key display is not correct to begin with.
After April 21:
![キャプチャ](https://user-images.githubusercontent.com/29241703/80299892-0205b880-87d3-11ea-91b9-aa0b61578e67.PNG)

Probably the cause is https://github.com/leezer3/OpenBVE/commit/4ab05b45fde9df1663049bb94159956948268293.

The confirmed environment is as follows:
[...]

PR #539, opened 2020-09-29
Fix: issue with partial transparency

s520:
This PR fixes the issue pointed out in https://github.com/leezer3/OpenBVE/issues/466#issuecomment-619619294.
This issue occurs in at least v1.4.3.0 and later main programs and v1.7.0.0 and later viewers.
This PR reverts the drawing of partially transparent faces to the viewer's method up to 1.6.0.0.

- v1.4.3.0 (Main program)
1 4 3 0-main

- v1.6.0.0 (RouteViewer)
1 6 0 0-route

- v1.7.0.0 (RouteViewer)
1 7 0 0-route

- This PR (RouteViewer)
1 7 2 0-route_fix

- This PR (Main program)
1 7 2 0-main-fix


leezer3:
Very nice.

I'll take a proper look later or tomorrow

leezer3:
If I read what you've done right, I'm not entirely sure this isn't just hiding the issue in this case.

Essentially, if I understand what you've done, this code is pulling the transparent and semi-transparent faces into two disparate lists & drawing the semi-transparent faces separately afterwards.
The issue with the linked file is that the drawing order is wrong https://github.com/leezer3/OpenBVE/issues/466#issuecomment-622055633

It shouldn't matter which list they're in.

Our real issue is where it selects the vertices to sort by here:
https://github.com/leezer3/OpenBVE/blob/master/source/LibRender2/Objects/ObjectLibrary.cs#L245
It's using a simple implementation of the painters algorithm, which breaks when dealing with overlapping polygons.

leezer3:
Definitely hiding the issue.
You'll see what I mean if you drop this tree with a replacement PNG texture in:

[trees2-snow.zip](https://github.com/leezer3/OpenBVE/files/5307839/trees2-snow.zip)

leezer3:
See what you think of this, it's a hack based on https://www.alecjacobson.com/weblog/?p=2750

```C#
if (Interface.CurrentOptions.TransparencyMode == TransparencyMode.Performance)
{
	SetBlendFunc();
	SetAlphaFunc(AlphaFunction.Greater, 0.0f);
	GL.DepthMask(false);
	foreach (FaceState face in VisibleObjects.AlphaFaces)
	{
		face.Draw();
	}
}
else
{
	SetAlphaFunc(AlphaFunction.Greater, 0.0f);
	GL.Disable(EnableCap.CullFace);
	GL.DepthFunc(DepthFunction.Less);
	foreach (FaceState face in VisibleObjects.AlphaFaces)
	{
		face.Draw();
	}
	GL.Enable(EnableCap.CullFace);
	GL.CullFace(CullFaceMode.Front);
	GL.DepthFunc(DepthFunction.Always);
	foreach (FaceState face in VisibleObjects.AlphaFaces)
	{
		face.Draw();
	}
	GL.Enable(EnableCap.CullFace);
	GL.CullFace(CullFaceMode.Front);
	GL.DepthFunc(DepthFunction.Lequal);
	foreach (FaceState face in VisibleObjects.AlphaFaces)
	{
		face.Draw();
	}
	GL.Enable(EnableCap.CullFace);
	GL.CullFace(CullFaceMode.Back);
	GL.DepthFunc(DepthFunction.Always);
	foreach (FaceState face in VisibleObjects.AlphaFaces)
	{
		face.Draw();
	}
	GL.Disable(EnableCap.CullFace);
	GL.DepthFunc(DepthFunction.Lequal);
	foreach (FaceState face in VisibleObjects.AlphaFaces)
	{
		face.Draw();
	}
	ResetOpenGlState();
}
```

You also need to add the following to the **ResetOpenGlState** function:
```
GL.CullFace(CullFaceMode.Front);
```

This will be slower at the minute on the old renderer, as display lists are broken. (and it obviously adds 3 draw passes for each face)
I suppose we could possibly add this (or another approach) under the quality option, and keep the current as the middle.

s520:
I'll take a look at it over the weekend.

s520:
My change may be wrong, but I've tried your suggestion and it doesn't seem to solve the problem.
test1
test2


Can you push your changes to GitHub?

Well, I would like to investigate techniques such as order-independent transparency. (for both performance and quality)

Also, we should change the transparency option item names to "Performance" and "Quality".
It's hard to tell what the current names "Sharp" and "Smooth" are.

PR #542, opened 2020-10-03
New: Implement windscreen wipers

leezer3:
Implements windscreen wipers independant of plugins.
This is compatible with / uses the same principles as OS_ATS & the UK train system plugins.

## Principle:
An array of **N** drops is available, each of which is visible (true) or invisible (false)
These drops should be added to the area swept by the windscreen wiper in order from L-R, in random vertical positions.
The wiper itself has a position value from 0 (left) to 100 (right)
Each frame, a random drop is added to the windscreen & lit.
As the wiper moves, if the drop it is passing over is lit, it is toggled to unlit.

### TODO:
* ~~Add in the animated variables for use in panel2 / animated windscreens. Probably **RainDrop[N]**~~ Using randomly generated drops.
* ~~Couple in the existing WindscreenWiper keys.~~
* ~~Add panel generator.~~ Randomly generating a drop set each time the panel is loaded. Not perfect, but makes it much easier to deal with.
* ~~Add configuration. (Raindrops section in panel2 & train.xml?)~~ Adds a simple windscreen section to Panel2.cfg (probably means a separate generator is irrelevant)
* ~~Plumb in the animated variables~~
* Future: Do something for panel.animated ??
* Add to XML panel & sound parser.

### Possible Improvements:
* Drops from all plugins are static. I suppose we could *possibly* add vertical movement on panel2 based trains.
* ~~Use built-in drop images.~~ Built in drop images used if nothing supplied.

---
### Panel2.cfg
New section: 
**[Windscreen]**

This supports the following arguments:
* **TopLeft** : The top left co-ordinate of the drop area
* **BottomRight** : The bottom right co-ordinate of the drop area
* **Layer** : The panel layer on which the drops appear
* **NumberOfDrops** : The total number of drops (positions are randomly generated each game launch)
* **DropSize** : The size of a drop in pixels
* **RestPosition** : The rest position (Left, right or -1 , 1)
* **HoldPosition** : The hold position (Left, right or -1 , 1)
* **WipeSpeed** : The speed in seconds for one pass (left to right) of the wiper.
* **DaytimeDrops** : A comma-separated list of files to use for the daytime drop images. (If not specified, built-in will be used)
* **NighttimeDrops** : A comma-separated list of files to use for the nighttime drop images. (If not specified, built-in will be used)
* **HoldTime** : The time in seconds the wiper pauses when using intermittant wiping.

New global panel2 variables:
* **WiperPosition** : The current position of the windscreen wiper, from 0 to 100.

### Sound.cfg
New section: 
**[Windscreen]**

This supports the following arguments:
* **RainDrop** : The sound played when a drop hits the windscreen
* **WetWipe** : The sound played when 20% or more of the total drops are visible
* **DryWipe** : The sound played when 20% or less of the total drops are visible
* **Switch** : The sound played when the wiper switch is pressed

### Routefile
New command:
**.Rain _Intensity_** Intensity should be a number from 0 to 100 representing the current rain intensity.

## Further notes:
Trains equipped with these wipers also recognise the legacy beacon based rain commands. (Beacon 21)
If one of these is present in the routefile, all .Rain commands will be ignored.

If a routefile uses the .Rain command, this is incompatible with legacy OpenBVE versions / trains at the minute.


leezer3:
Starting to take shape.
For the minute at least, the windscreen lives in the train panel. (Logical as it's an animation property of the panel)

Unfortunately, at the minute I haven't got a good solution for generating similar 3D drops on a panel, so this is 2D only at this point.

Next job to sort out is hooking in an appropriate set of routefile commands.
Best bet is probably to do the following:
* React to legacy OS_ATS / UKTrainsys rain beacons.
* Add an entirely new routefile command, **.Rain _intensity_** When hitting this, we probably also need to send out a backwards compatible beacon signal (arrg.h...)

leezer3:
[Raindrops video](https://www.youtube.com/watch?v=sCE2Iu8PeE4)

leezer3:
Now in the main builds, just need to add some documentation to the thing.

Issue #528, opened 2020-09-19
Input device plugins and hold brake


MarcRiera:
### Summary

I am currently developing an input device plugin for the "Densha de GO!" controllers for the PS1. No problem regarding the controller themselves, with adapters they show up as USB gamepads and then it is just a matter of assigning the correct button combinations to OpenBVE commands. Once it is mature enough (soon, I hope!), I will send a pull request to get it in the main source repo.

While developing, I have noticed two issues regarding the use of input device plugins with trains with a hold brake.

### Issues

1. Currently, input plugins have no way of knowing whether the train they are controlling has a hold brake or not. The command to apply or release the hold brake (`OpenBveApi.Interface.Translations.Command.HoldBrake`) works, but depending on the train it will be needed or not. In-game, OpenBVE presents the hold brake as an extra notch between N and B1, so it makes sense to bind it to that physical position in a real controller. The presence or not of a hold brake thus affects the binding of the rest of brake notches.

2. When using the hold brake application command directly, OpenBVE currently does not change the brake notch. This way, if the brake notch is B1 and the hold brake is applied, the regular brakes will still be applied. This is inconsistent with how the feature is presented to the user in-game (as mentioned before) and is not limited to input device plugins, as it affects any use of `OpenBveApi.Interface.Translations.Command.HoldBrake`.

### Proposed solutions

1. Extend the arguments in the `SetMaxNotch` method, currently informing about the maximum notches for the driver. This would be backwards incompatible with existing plugins. There is at least one existing plugin outside of the main repo (https://github.com/TetsuOtter/BIDSSMemLib) that would require changes, but fortunately the changes would be minimal. To prevent further issues if we need to expand this again in the future, it would be wiser to replace it with something equal or similar to `SetVehicleSpecs` for train plugins, with driver notches instead of "internal" notches.

2. Ensure we move the brake notch to 0 when applying the hold brake.

I can take care of both fixes once we agree on the solutions (specially the first aspect).

leezer3:
My initial instinct would be to add something like a new SetTrainSpecs function with an appropriate struct and leave the old one alone entirely.

This preserves the backward compatability & would allow for future expansion if another controller required a more esoteric property.

The other option would be to implement it into the main tree completely, same basic way as a RailDriver, although that might be tricky if it's not easily detectable.
Doing that you'd have full visibility of the internals of everything.

@s520 thoughts?

s520:
I support the first idea of leezer3.

[...]

PR #543, opened 2020-10-03
Densha de GO! controller plugin

MarcRiera:
This pull request implements an input device plugin for Densha de GO! controllers, as mentioned in #528. Technical documentation on the controllers can be found [here](https://gist.github.com/MarcRiera/075b5651039942b2380dc51f0ec1cbc9).

image

## Features
* Compatible with classic (non-USB) Densha de GO! controllers for consoles: Nintendo 64 (TCPP-20003), Sega Dreamcast (TCPP-20003), Sega Saturn (TC-5175290) and Sony PlayStation (SCPH-00051, TCPP-20001, TCPP-20002, TCPP-20008). Naturally, **a USB adapter is required**.
* Initial (untested) compatibility with DGOC-44U and DGC-255 (USB controllers by Unbalance).
* Automatic calibration via a wizard. The user is asked to follow a sequence of button presses and the plugin automatically detects the configuration for the adapter. This only needs to be done once or after changing the adapter.
* Several options regarding how the power and brake notches are mapped between the physical and virtual handles.
* Command mapping to physical buttons.
* Localization support.

## Limitations
* The controllers have their own way of being detected on consoles but this is lost when using adapters. For this reason, autodetection is not possible and the plugin will accept any USB controller. Disable the plugin when not in use to avoid interference.
* Some adapters map the direction buttons on a standard PlayStation controller to a hat. Due to how the Densha de GO! controllers work, **N** and **P4** on the power handle become undistinguishable. The plugin takes this into account and will guess the correct notch most times, but can become confused specially when changing notch too fast.
* For a similar reason, because the Dreamcast controller needs to press the four direction buttons at the same time, it will not work with adapters which map them to a hat. Only adapters mapping them to independent buttons will work with OpenBVE.

## Future work
* Confirm compatibility with USB controllers by Unbalance.
* Add support for USB controllers for the Sony PlayStation 2.

### DIFF
--- a/OpenBVE.sln
+++ b/OpenBVE.sln
@@ -96,6 +96,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SanYingInput", "source\Inpu
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DefaultDisplayPlugin", "source\InputDevicePlugins\DefaultDisplayPlugin\DefaultDisplayPlugin.csproj", "{3D833D81-4F14-4A68-B0DE-6469B8CF4E62}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DenshaDeGoInput", "source\InputDevicePlugins\DenshaDeGoInput\DenshaDeGoInput.csproj", "{F6FBD153-9D9F-4D6C-B08D-0C798F106B32}"
+EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Texture.Tga", "source\Plugins\Texture.Tga\Texture.Tga.csproj", "{F714A9FA-1C4B-456B-BACF-D6763223A81C}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AssimpParser", "source\AssimpParser\AssimpParser.csproj", "{77CCAE62-D374-468E-98A0-DD8ADB461C10}"
@@ -551,6 +553,10 @@ Global
 		{8DAA1CF4-A29A-42CE-8649-34E0FBC0D97C}.Release|Mixed Platforms.Build.0 = Release|Any CPU
 		{8DAA1CF4-A29A-42CE-8649-34E0FBC0D97C}.Release|x86.ActiveCfg = Release|Any CPU
 		{8DAA1CF4-A29A-42CE-8649-34E0FBC0D97C}.Release|x86.Build.0 = Release|Any CPU
+		{F6FBD153-9D9F-4D6C-B08D-0C798F106B32}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{F6FBD153-9D9F-4D6C-B08D-0C798F106B32}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{F6FBD153-9D9F-4D6C-B08D-0C798F106B32}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{F6FBD153-9D9F-4D6C-B08D-0C798F106B32}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -582,6 +588,7 @@ Global
 		{C841EA02-4918-4162-ADC1-C09F5BC28AF9} = {16553295-E70F-4596-AA78-848EEA576C4A}
 		{8D21C5C6-73E9-42E0-92FB-63500E321836} = {16553295-E70F-4596-AA78-848EEA576C4A}
 		{F49789F2-97F3-45B3-BC85-F4E09C0D120D} = {16553295-E70F-4596-AA78-848EEA576C4A}
+		{F6FBD153-9D9F-4D6C-B08D-0C798F106B32} = {851ADF1E-4500-4A14-B4AA-C1087A6C537E}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {50B52A67-69B1-4B19-A59C-FF058FD9EBDF}
--- a/OpenBVE.sln
+++ b/OpenBVE.sln
@@ -557,6 +557,14 @@ Global
 		{F6FBD153-9D9F-4D6C-B08D-0C798F106B32}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{F6FBD153-9D9F-4D6C-B08D-0C798F106B32}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{F6FBD153-9D9F-4D6C-B08D-0C798F106B32}.Release|Any CPU.Build.0 = Release|Any CPU
+		{F6FBD153-9D9F-4D6C-B08D-0C798F106B32}.Debug|Mixed Platforms.ActiveCfg = Debug|Any CPU
+		{F6FBD153-9D9F-4D6C-B08D-0C798F106B32}.Debug|Mixed Platforms.Build.0 = Debug|Any CPU
+		{F6FBD153-9D9F-4D6C-B08D-0C798F106B32}.Debug|x86.ActiveCfg = Debug|Any CPU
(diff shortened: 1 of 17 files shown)


leezer3 (commented):
Presumably at the least you could check for number of buttons / axis?
Would eliminate a lot of controllers.

Unfortunately  IIRC the OpenTK controller name is broken else that might have been testable against.

MarcRiera:
> Presumably at the least you could check for number of buttons / axis?
> Would eliminate a lot of controllers.
> 
> Unfortunately IIRC the OpenTK controller name is broken else that might have been testable against.

We could add a check for a minimum and maximum amount of buttons, but not very specific; the capabilities reported may vary between adapters. We cannot rely either on what I mentioned some time ago about the axis position, as it seems to be specific of my adapter. In fact, even if OpenTK correctly reported the name of the device, it would be useless in this case, because the adapter would report the same regardless of the PlayStation controller connected to it (my adapter shows up as "PlayStation 3 controller").

Fortunately, selecting a device, calibrating it and testing it is very easy and user-friendly.

leezer3:
Minimum should be easy?

The controller notch mappings simply don't work unless sufficient buttons are present, and so if the adaptor is unsuitable, blocking it is probably better.

Name wise, I was thinking more along the lines of deliberately blocking things actually, but a valid point.

leezer3:
Another minor point:
As significant 'new' code, the licence probably should be clarified. (So as not to leave this in the current messy situation in the future)
Ideally I'd say BSD-2 or similar, but let me know if you've got any other thoughts.

MarcRiera:
Added the license to the files. By BSD-2 I understand you the refer to the Simplified BSD (BSD with two clauses). Let me know if you meant something else; I am totally fine with any BSD license for this.

Regarding the checks, I have added a restriction to a minimum of 10 buttons plus one hat or 12 buttons, which is the bare minimum the plugin needs. For the maximum, I have set it to 20 buttons, which is one button more than my adapter in "PS3" mode and enough for any valid adapter mapping everything to buttons (a DualShock has 17 buttons or so).

leezer3:
Yeah, that's great 👍 
BSD-2 is essentially as close as we can get to Michelle's original intentions, whilst still having something that's recognised as an actual licence, not a mess.

Some further reference on the licence stuff, which was a quite major discussion last year: https://github.com/leezer3/OpenBVE/issues/305

Can's see any reason why AppVeyor should fall over at the minute, possibly because the new project doesn't exist on the master branch, but that's only speculation.
Travis is OK, and it builds locally on Linux / Win.

MarcRiera:
@leezer3 Despite changing the plugin to ITrainInputDevice, it seems that OpenBVE is not passing the vehicle specs properly. The default specs for the plugin are 5 power notches, 8 brake notches and no hold brake, and they stay the same regardless of the train used. Removing the default specs and trying to load the plugin results in an unhandled error, probably due to unassigned variables, which is what makes me think `SetVehicleSpecs` is not being called at all.

leezer3:
My fault.
Checking the plugin info class, not the plugin itself. Commit I've just pushed to your branch should sort.

(n.b. Opening your branch with Visual studio adds a bunch to the SLN file. Maybe this is what's missing to allow CI to work)

MarcRiera:
Thanks, it is working now. Tested with a bunch of different trains and the notches map as expected.

I will take a look at what you mention about Visual Studio and push any additions to the solution file. I mainly develop under Linux and use Visual Studio only when designing Windows Forms (there is no way to do it visually on Linux as far as I know).

leezer3:
Thanks!

Is this ready for merge now (still marked as draft)?

I can't really test beyond the fact that thing builds and doesn't crash as obviously I don't have one of these (much as I'd quite like one to fiddle with, but money, money, money........)

MarcRiera:
It's nearly ready; last week I found documentation about two USB controllers for the PC versions of the games and I'm adding initial support. I don't have these controllers to test them (they are rarer and more expensive than the PlayStation ones), but I'll ask on online forums to see if I can find someone to confirm it works.

I've done extensive testing and fixed any bugs I found. The plugin ignores unsupported controllers, stops polling a supported controller if it is disconnected and does not lock the virtual notches. It should be at least as stable as the other included input plugins.

leezer3:
Good stuff :)

The PC ones will probably work, assuming they present as a relatively standard joystick / pad.
If they're something more exotic, they may or may not do something.

Another marginally related thought-
I saw a game recently which auto-paused if the controller disconnected.
Mildly tempted to add this as an option.

MarcRiera:
They are standard and have a VID/PID, which we cannot use with OpenTK at the moment, but might be possible in the future. Meanwhile, I'm using specific button/axis/hat count for detection.

The PS2 versions of the controllers (USB too) should also be usable, but I have found little documentation and other compatible PC software seems to rely on unofficial drivers. I'll ask someone with one of those to try https://gamepad-tester.com/ and report me what it detects.

> Another marginally related thought-
> I saw a game recently which auto-paused if the controller disconnected.
> Mildly tempted to add this as an option.

I'd add a warning as well to make sure the user knows what has happened,  to continue

leezer3:
Now, that's a good idea.

Probably ought to check for whether we're actually using the thing for any controls, then pause / prompt as appropriate.

Will try and knock something up over the next few days.

leezer3:
https://github.com/opentk/opentk/pull/1180

That fixes the joystick reconnect issues under Linux, although we'll see what OpenTK think of it. 
(Haven't tested the patched version with OpenBVE at the minute; We may have more issues on our end to deal with)

MarcRiera:
Good, I will make sure the code supports reconnection so it works properly when that gets merged.

Still working on the USB controller (needs additional UI due to the extra buttons), but it turns out that all the classic console controllers (Nintendo 64, Sega Dreamcast, Sega Saturn and Sony PlayStation) work the same internally and should work with this. Check the gist linked in the OP for more technical details. I have renamed everything in the code which applies to all consoles and previously thought that was PlayStation-specific.

MarcRiera:
Ready to merge!

leezer3:
Marginally experimental PR added to your branch, see what you think:
https://github.com/MarcRiera/OpenBVE/pull/1

MarcRiera:
It is marginally related, but the latest commit fixes #528 (the brakes are now released when applying the hold brake directly with a command and the brake notch is B1).

I have made all the changes to make the plugin work again, except the buttons, which do not work for some reason. The problem is definitely with the dictionary, but I have no idea what it is. Will keep investigating...

leezer3:
See my commented line above.

You want to get the stuff from the InputTranslator into the dictionary you're actually checking the pressed / released states against.

leezer3:
https://github.com/leezer3/OpenBVE/commit/c3ac97b85af2b29d2d83d4df3ccb4445a11df16f

This is the major part of reconnection sorted.
I'll try and deal with the actual reconnection process in-game next week.

MarcRiera:
Great! I have updated the OpenTK reference to 3.3.1 and I can confirm that `IsConnected` now seems to work (previously, it was not reliable). Nothing different regarding input after reconnection: the plugin now detects the controller is connected again, but the device reports no further input Not sure if this should be working.

The device dropdown now also displays the USB device name. I am not using it for device detection, though; I found out OpenTK generates joystick GUIDs based on vendor/product so last week I implemented a more reliable check.

I finally solved the button issue with a more simple approach and simplified other parts of the code with messy conditions. Other than the reconnection issues, I think there is nothing else missing.

leezer3:
Good, that caused me enough trouble :)

Planning to look at reconnection in-game next week with any luck.

Merge this now, or wait until that's sorted?

MarcRiera:
You can merge it already, it is fully functional.

leezer3:
I think this branch solves the issues:
https://github.com/leezer3/OpenBVE/commit/6de517079efcb11f1b4c9b227b214b041b55e87c

TLDR:
In some cases (primarily multiple sticks & Linux), the OpenTK joystick index can change on reconnect, although it shouldn't do this....
This switches us to use a GUID dictionary when mapping joysticks, as opposed to the raw index.

MarcRiera:
Ok, this will require a small change to the plugin, which currently stores the raw index in the config file. I know SanYingInput uses some sort of GUID but I think it is not the one provided by OpenTK. Maybe that one requires a change for reconnection as well.

leezer3:
Good point, hadn't thought of that directly.

I'm working on a State from GUID function for OpenTK which will simplify things in my branch somewhat, but that may take a while.
If it works with your modification, I think merge that and we're relatively OK in the short term.

Will also take a look in SanYinInput, but worse-case scenario, reconnection doesn't work, which it doesn't anyway :)

PR #544, opened 2020-10-07
Change: Unify vector parsing

leezer3:
This branch attempts to unify the Vector parsing code considerably.

Most places parsing a Vector simply called the **NumberFormats.TryParseDoubleVb6** method repeatedly on each component of the vector, and had their own **Interface.AddMessage** call with an appropriate error message.

This PR creates several new helper functions in the NumberFormats file which perform the function in a unified way.

### Benefits:
* Single set of methods for parsing vectors.
* Gets rid of a bunch of duplicated code.

### Drawbacks / Issues
* This unfortunately changes the structure of some error messages. Most users don't worry about these, but may confuse developers slightly?
**New:** Error X is invalid in AddVertex - Coordinates at line 2 in B:\test\FCMB 2000\3D\Cabina\AgullaPressioB.csv
**Old:** Error Invalid argument vX in AddVertex at line 2 in file B:\test\FCMB 2000\3D\Cabina\AgullaPressioB.csv
* Should these be moved to be vector constructors?
* Slightly nasty hack to enable the messages to be added by the API.
* More places can use this, e.g. TrainEditor2

@s520 any thoughts?
Ended up running this up as I looked in the Panel2 parser in preparation for adding the windscreen properties & decided I could get rid of some of the duplicated code.


source/OpenBveApi/Math/Math.cs
```diff
@@ -179,6 +183,83 @@ public static bool TryParseDoubleVb6(string Expression, double[] UnitFactors, ou
...
 		}
 
+		/// <summary>Parses a Vector2 formatted as a Visual Basic 6 string</summary>
+		/// <param name="Value">The string value</param>
+		/// <param name="Key">The key value</param>
+		/// <param name="Section">The section</param>
+		/// <param name="Line">The line</param>
+		/// <param name="FileName">The filename</param>
+		/// <param name="ExpectedArguments">Whether two arguments are expected</param>
+		/// <returns>True if parsing succeeds, false otherwise</returns>
```
leezer3, line 193: Typo, need to remove these

source/OpenBveApi/Math/Math.cs
```diff
@@ -179,6 +183,83 @@ public static bool TryParseDoubleVb6(string Expression, double[] UnitFactors, ou
...
+		public static Vector2 TryParseVector2(string Value, string Key, string Section, int Line, string FileName, bool ExpectedArguments = false)
```
leezer3, line 194: Add an Arguments variant, as per the Vector3

source/OpenBVE/Parsers/Panel/PanelXmlParser.cs
```diff
@@ -781,80 +582,39 @@ private static void ParsePanelNode(XElement Element, string FileName, string Tra
...
-										}
+										Origin = NumberFormats.ParseVector2(Value, Key, Section, LineNumber, FileName, true);
+										OriginDefined = true;
 										break;
 									case "initialangle":
-										if (Value.Length != 0 && !NumberFormats.TryParseDoubleVb6(Value, out InitialAngle))
-										{
-											Interface.AddMessage(MessageType.Error, false, "ValueInDegrees is invalid in " + Key + " in " + Section + " at line " + LineNumber.ToString(Culture) + " in " + FileName);
-										}
+										InitialAngle = NumberFormats.ParseDouble(Value, Key, Section, LineNumber, FileName);
```
leezer3, line 589: Should this be converted to radians?
Nasty suspicion it should....

s520:
I'm busy this week, so I'd like to see it next week.

leezer3:
No trouble.

There are a lot more places and similar bits that can be added (color parser for one, used in the objects, panels etc), & I'd consider this low priority cleanup work at best.

leezer3:
This PR is now about 50% of the way towards replacing the parse calls.

A lot of the rest of these use the negative response to set some sort of default value. (where the standard double / int default of 0.0 doesn't apply)
Need to think about the correct way to do this....

PR #552, opened 2020-11-03
Change: Alter travis file [ci skip]

leezer3:
https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

We'll hit this whenever it comes in....
Can probably get away with running only the upload build on OS-X as the testing itself will be done on Linux.

Possible alternatives to investigate:
* CircleCI
* Github Actions

Other potential changes to make:
* Cut back further on builds triggered. We can probably loose the makefile test builds (as the publish builds are roughly equivialant)
* Chase Travis for OSS credits. Very unclear as to how this will actually work in practice.

---

As a rough rule of thumb, let's assume 1 commit per day (averaged out including PR builds etc).
It's unclear as to what exactly they're going to charge for, as 90% of the build time is setup of their image (not our code running)

A typical build takes anywhere from 10min to 30min, depending on the Travis load. 
Of this ~1min is the actual build time & the rest is setup, which is somewhat mad, especially as it looks like they're going to charge credits for the whole shebang.

Therefore taking a median travis build time of 15min & currently the single build change from this:
* 750 credits per build * 30 days in month = ~22,500 credits per month just for OS-X
* 900 credits for the rest of the builds * 30 days in month = ~27,000 credits per month
TOTAL: ~50,000 credits per month.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,9 @@ linuxbuild.*
 macbuild.*
 /mac/
 installers/debian.deb
+OpenBVE-*.zip
+OpenBVE-*.deb
+OpenBVE-*.dmg
 
 #Debian Build directory
 installers/debian/usr/lib/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,21 +16,10 @@ matrix:
       script: make debian
       if: tag IS present
 
-    #Test builds
-    ##Use makefile##
+    #Further test builds on different Ubuntu / Mono versions and using the SLN file directly
     
-    #Note: Only run upload builds on MacOS https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
-    #Let's test all variations on Linux, but can probably get away with only running an upload build on OS-X
-    #TODO: Consider possible alternatives
-    - os: linux
-      dist: xenial
-      mono: latest
-      script: make all-release
-    - os: linux
-      dist: trusty
-      mono: 5.20.1
-      script: make all-release
-    ##Use .sln file##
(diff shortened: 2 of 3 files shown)


leezer3:
cc @s520 (hopefully you're OK and not too busy)

Second commit drops to ~41,000 credits / month. (all assuming the setup time is charged for)
If setup time is *not* charged for, then use is ~5,000 credits, which would likely put in the free tier comfortably.

I'm going to try and get a CircleCI account running and see what happens there.....

leezer3:
https://azure.microsoft.com/en-gb/services/devops/pipelines/

Azure Pipelines as an alternative maybe?

leezer3:
More minor fiddling-
The makefile now produces our final output name when running a task which includes compression, as opposed to this being a Travis step. Will allow slightly easier migration away.

Azure is looking quite decent at the minute, and possibly less likely to drop us in it.
Need to setup some pipelines for Linux and OS-X workers & see if we can get it running the makefile.

leezer3:
So, Azure Pipelines seems to work very nicely (after a few minor hiccups to do with SSH authorisation)

https://github.com/leezer3/OpenBVE/tree/azure-pipelines

The one flaw in this setup is that we can only get the latest version of Mono easily, which Travis gives us.

I haven't investigated this fully, but I'm not especially happy moving the Linux uploads to the current Mono version whilst Debian is somewhat behind on stable.
Need a bit of thinking; I suppose we could do something like keeping Travis for the Linux compatability build *only*.....

PR #556, opened 2020-11-15
Use GUIDs for Densha de GO! controllers

MarcRiera:
The Densha de GO! input plugin now uses and stores GUIDs to identify controllers instead of raw indices, which may change after disconnecting and reconnecting a controller.

May be further simplified in the future if it becomes possible to poll devices directly by GUID instead of index. For the time being, it works and should allow disconnecting and reconnecting controllers safely.


leezer3:
Merged this and the other reconnect branch of mine-

Thanks!

Let me know if you've seen any other glitches- Intending on pushing a build some time this week or next.

Issue #209, opened 2017-12-23
Implement Mechanik Parser/migration tool for OpenBVE 


afarlie:
BVE Cornwall has a small number of ancient routes for Mechanik, which no longer runs easily on some modern systems.  

It would be nice to have a way to render/play the small number of existing legacy routes in a modern simulator like OpenBVE,

Sirflankalot:
I am currently working on a project that will include tools for easy modernization of routes. I only just encountered these old Mechanik routes, but they are definitely on my todo list for this project. Though, I don't want to get too ahead of myself and say dates.

MarcRiera:
I was actually thinking about Mechanik routes in OpenBVE a while ago, and the format is well-documented. It's messy, it's stuff from 20 years ago, but there are only a few commands. There's an old but great German website with a small tutorial on how to develop Mechanik routes: http://mechanik.peterulrich.de/trasa.html (it's in German, but I've managed to follow it using machine translation).
The Mechanik format has the advantage of being based on the player's rail just like BVE. Curves aren't used frequently because the rotation had to be calculated manually by the route developer, but it shouldn't be too difficult to generate radius values from offset values. The disadvantage of the format is that the concept of 'object' (as in BVE) doesn't exist; textures are placed directly. I guess a good idea for a Mechanik to BVE converter would be to generate an object for each texture declaration, and then check for duplicates.
While I'm not 100% sure I could be able to write such a converter myself, I'm quite familiar with Mechanik and I could do some testing and reverse engineering to at least write proper documentation for the trasa.dat format.

Sirflankalot:
That's a lovely bit of documentation and from the looks of it the machine translation did perfectly fine. 

That texture issue seems like an issue from the outside but it really wouldn't be a problem at all. Your method would be easy to implement and fast. 

> While I'm not 100% sure I could be able to write such a converter myself, I'm quite familiar with Mechanik and I could do some testing and reverse engineering to at least write proper documentation for the trasa.dat format.

This would be **AMAZINGLY** helpful for getting this kind of thing implemented.  If I had proper documentation for it, I could possibly roll a parser in a couple days. Still wouldn't be the highest thing on my TODO list, but it would get done.

leezer3:
This is as close to official documentation as there ever was (Czech, although Google translate doesn't auto-detect it for some reason):
http://mechanik.wz.cz/edit/soubory.php

I've briefly played with this in the past, but haven't got anything to a reasonably working stage.

Caveats:
[...]

PR #591, opened 2021-03-03
New: Implement Mechanik route parser plugin

leezer3:
This PR implements a plugin to parse routes from the legacy DOS based simulator Mechanik.

Closes https://github.com/leezer3/OpenBVE/issues/209

### Working:
* All faces are now shown correctly.
* Basic curves. (not widely used by Mechanik routes)
* Works in both Route Viewer and the main simulation.
* Loading screen image is used as the route image in the main screen.

### Issues:
* Correction after a large curve isn't quite right yet. (Track may be offset by ~0.1m from intended place, some routes are worse than others)
* Sound events not properly implemented yet.

### Not Implemented:
* Not implementing a cab parser just at the minute. Due to the way Mechanik worked, everything was a variant on the supplied EN57 cab image, and most stuff was just a simple textual overlay. Probably will not implement.
* Neutral sections. Experiment here for CSV / RW routes: https://github.com/leezer3/OpenBVE/pull/501 perhaps overcomplicated, with voltages etc. (Athough that was built after talking to someone who specifically uses a plugin for *all* of that....)


leezer3:
Don't know if we *really* need the property database.
IIRC Terra Australis is the only route with station names visible / in-game, and I'll need to think of a description for our known routes.

Default trains might be interesting to find too :/
I'd be inclined to use BVE2 standard stuff for these (not complex exterior things), simply as the original game obviously never had such luxuries.
On the flip side, later stuff is likely to be more available.

leezer3:
Added database entries for a bunch of French stuff, which has station names / times in the readme.

This at least makes the database somewhat more worthwhile.
I'm tempted to drive the other routes using AI to get the time betweeen stations and create a timetable. This would make the in-game timetable / display better, but gets a little further away from the original :/

leezer3:
Added a modules database and another entry to the known routes.
Should be able to add the door sides into the database too. Not really necessary, but let's make the user experience the best possible.

Other thoughts:
I'm wondering about detecting the module control strings in the final output file, and using these to issue a correction and seeing if that helps. Still haven't figured out exactly what rules the original plays by in this case....

leezer3:
Few more updates.
Signalling is definitely glitchy when viewed from outside / behind the train, but it was never intended to be viewed in this manner.

Issue #594, opened 2021-03-14
[DdGO! controllers] Power notches not detected with certain adapters


MarcRiera:
### Description

Certain PlayStation to USB adapters (made by Elecom, other brands may be affected too) have an unconventional way of mapping the D-pad buttons, which are used for the power notches. These adapters map these buttons to a hat switch, which does not allow pressing left and right simultaneously.

OpenBVE already supports such adapterswhen pressing left and right simultaneously results in only left reported as pressed. However, the Elecom adapters report center (no direction) and the input plugin gets confused.

(Received this bug report via email, leaving it here for future reference and to remind me of fixing it later).

MarcRiera:
Fixed in #600

PR #600, opened 2021-03-26
Improvements for DdGO! controllers

MarcRiera:
This PR adds the following:

- Fix for #594. Has the downside of the power notch jumping P1>N>P2, but it is noticeable only when moving the handle slowly and is not critical.
- Allow controllers to define a notch count different from P5/B8 (needed to support other controllers with more and less physical notches). The UI has been updated accordingly.
- Compatibility with PS2 Type II and Shinkansen controllers.

Todo:

- ~Fix handle mapping for the new controllers (there is something broken with the emergency brake).~
- ~Finish support for the Type II controller~
- ~Move the input read loop to an independent async task to increase responsiveness.~
- ~Add PS2 controller reconnection support (not working in-game for some reason).~
- ~Add a help form with OS-specific instructions on how to configure the PS2 controllers.~


MarcRiera:
Got the Shinkansen controller working: https://www.youtube.com/watch?v=7j0nyys7v9g

In summary: the three PS2 controllers use a HID interface, but it is hidden from the OS by a vendor-specific device class. This means we cannot use the standard OpenTK input way to access them unless OS-specific drivers are developed. Instead, I am using [LibUsbDotNet](https://github.com/LibUsbDotNet/LibUsbDotNet) to communicate directly with the devices, so the input device plugin serves as a multiplatform userspace driver. There are some special requirements on first usage (generic driver on Windows and udev rules on Linux) but they will be documented in the plugin config window itself.

The Shinkansen controller in particular has a display for current speed, current speed limit and door state (open/closed). The current speed is already passed in ElapseData, but there is no way to access the current speed limit or the door state. @leezer3 Is it possible to add this without breaking other input plugins?

leezer3:
Should be.
Will look in detail tomorrow.

Adding stuff is trivial (only really need to make sure you get the name 'right first time), it's renaming or removing stiff which is breaking.
C# is so much more forgiving as a managed language in this regard :)

leezer3:
Actually, it might be better to just add an appropriate door change method to the interface & cache it internally:
https://github.com/leezer3/OpenBVE/blob/master/source/OpenBveApi/Interface/Input/InputDevice.cs#L49

That would maintain the maximum backwards compatability, and mirror the train plugin API.
Obviously, that's the way it is due to legacy BVE4, but we might as well be consistant.

leezer3:
Something else as a secondary query; Can you reproduce this at all?
https://bveworldwide.forumotion.com/t2065-2d-panels-not-displaying-in-recent-builds#20540

MarcRiera:
> 
> 
> Actually, it might be better to just add an appropriate door change method to the interface & cache it internally:
> https://github.com/leezer3/OpenBVE/blob/master/source/OpenBveApi/Interface/Input/InputDevice.cs#L49
> 
> That would maintain the maximum backwards compatability, and mirror the train plugin API.
> Obviously, that's the way it is due to legacy BVE4, but we might as well be consistant.

No problem. If you want to mirror SetBeacon for the speed limits it's also fine, it's very simple to get the current and upcoming limit with this function and the location from ElapseData.

> 
> 
> Something else as a secondary query; Can you reproduce this at all?
> https://bveworldwide.forumotion.com/t2065-2d-panels-not-displaying-in-recent-builds#20540

I'll take a look at the latest builds and check if that happens (I haven't noticed it during development, but this branch isn't up to date with master and the issue seems to occur on very recent builds).

leezer3:
My basic thought though is to keep everything as consistant as possible between different implementations / parts of the API.

As I'm sure you know, inconsistancy is rife, and there are often 3+ implementations of the same thing in each of the viewers etc, which is why I've been pushing everything into shared plugins.
Makes everything so much easier to maintain.

leezer3:
Looking good so far, let me know when it's ready to merge.

Please also note that I'm planning to merge this *soon* (probably start of next week) as the next major internal feature:
https://github.com/leezer3/OpenBVE/pull/606

I don't think it'll affect anything (other than needing to mark as AnyCPU rather than x86 only), but it's a large enough potential change that I'm giving you a heads-up in this PR.

MarcRiera:
> Looking good so far, let me know when it's ready to merge.
> 
> Please also note that I'm planning to merge this _soon_ (probably start of next week) as the next major internal feature:
> #606
> 
> I don't think it'll affect anything (other than needing to mark as AnyCPU rather than x86 only), but it's a large enough potential change that I'm giving you a heads-up in this PR.

Thanks for the heads-up! I don't think it'll affect anything either, but will be ready to fix any breakage.

I'll need some more time to get this ready to merge. I've got the PS2 controllers working with a background worker but for some reason reconnection is not reliable (it works in the config window but not in-game). I suspect it's some sort of race condition, but I need to fully trace the bug. It was previously working fine without the background worker, but that created substantial input lag.

Other than that, only the OS-specific config for Windows and Linux is missing.

leezer3:
You may need to try a pure thread as opposed to a background worker.

Background workers are somewhat intertied with WinForms & the WM Message pump, and the main game window isn't a WinForm. More likely to be an issue on non-windows too.

MarcRiera:
> You may need to try a pure thread as opposed to a background worker.
> 
> Background workers are somewhat intertied with WinForms & the WM Message pump, and the main game window isn't a WinForm. More likely to be an issue on non-windows too.

Got rid of the background worker and used a task. Everything now works as expected, with the task receiving input in a loop and ending cleanly if the controller is disconnected or the user selects something else. If there is a reconnection, the task is started again.

I've added the SetSignal, SetBeacon and DoorChange functions to the plugin with the necessary code for the Shinkansen controller. The main program part of this is missing (the functions don't get called yet); do you want me to implement the functions in a PR or do I leave it to you? DoorChange looks very easy to add but SetSignal and SetBeacon have additional code near them and I'm not entirely sure I'll add them properly. I have only considered these three functions because the rest used in train plugins don't really make sense in the context of input device plugins.

leezer3:
Either or, it makes no real odds :)
IIRC the SetSignal / beacon code you're looking at is for finding the next red section.
Would need to go back and look at the API docs to figure exactly what, but it probably wants to be left there to maintain compatability.

MarcRiera:
This is now ready to merge. The plugin config window is now cleaner:

imatge

A help window has been also added for basic troubleshooting and to provide the OS-specific steps (there's nothing for Mac, should work out of the box).

imatge

leezer3:
They aren't standard USB.
You'll need to copy in the udev rules file I think.

We might be able to do this automatically, would have to look and see if there is a documented interface for requesting root permissions.

ginga81:
Ok!
Thank you!
What do I do?

leezer3:
@MarcRiera will have to comment, he's done most of the work on this, and actually has one or two of these :)

leezer3:
Yeah, I only just noticed that the box was going missing on resize (anchor set to top not bottom).

Oddly, my Linux VM crashes when opening the DenshaDeGo input config form:
```
5/11/2021 4:09:52 PM
OpenBVE 1.7.560.1268-leeze Crash Log
Program is running on the Linux backend
Current screen resolution is: Windowed 960px x 600px 
The exception caught was as follows: 
System.Threading.ThreadExceptionEventArgs  at System.Environment.get_StackTrace () [0x00000] in <533173d24dae460899d2b10975534bb0>:0 
  at OpenBve.CrashHandler.UIThreadException (System.Object sender, System.Threading.ThreadExceptionEventArgs t) [0x00000] in <abf410090ee0420294540cccf7971155>:0 
  at System.Windows.Forms.Application.OnThreadException (System.Exception t) [0x00000] in <6d635ac3dc1c4424ad385ded79f1e868>:0 
  at System.Windows.Forms.NativeWindow.OnThreadException (System.Exception e) [0x00000] in <6d635ac3dc1c4424ad385ded79f1e868>:0 
  at System.Windows.Forms.NativeWindow.WndProc (System.IntPtr hWnd, System.Windows.Forms.Msg msg, System.IntPtr wParam, System.IntPtr lParam) [0x00000] in <6d635ac3dc1c4424ad385ded79f1e868>:0 
  at System.Windows.Forms.XplatUIX11.DispatchMessage (System.Windows.Forms.MSG& msg) [0x00000] in <6d635ac3dc1c4424ad385ded79f1e868>:0 
  at System.Windows.Forms.XplatUI.DispatchMessage (System.Windows.Forms.MSG& msg) [0x00000] in <6d635ac3dc1c4424ad385ded79f1e868>:0 
  at System.Windows.Forms.Application.RunLoop (System.Boolean Modal, System.Windows.Forms.ApplicationContext context) [0x00000] in <6d635ac3dc1c4424ad385ded79f1e868>:0 
  at System.Windows.Forms.Form.ShowDialog (System.Windows.Forms.IWin32Window owner) [0x00000] in <6d635ac3dc1c4424ad385ded79f1e868>:0 
  at System.Windows.Forms.Form.ShowDialog () [0x00000] in <6d635ac3dc1c4424ad385ded79f1e868>:0 
  at OpenBve.formMain.ShowMainDialog (OpenBve.formMain+MainDialogResult initial) [0x00000] in <abf410090ee0420294540cccf7971155>:0 
  at OpenBve.Program.Main (System.String[] args) [0x00000] in <abf410090ee0420294540cccf7971155>:0 
Current program memory usage: 99.73mb
System memory free: 0mb
```

Exception stack isn't really showing anything useful, just that we're crashing when showing it as a modal dialog.
I'd guess something in the constructor, but I'm looking into that at the minute.

ginga81:
Basically, I can't press the config button.
What do I do?
noconfig

leezer3:
The version linked above should solve that also.

ginga81:
Oh!
Each day's zip, but has been changed the window size issue!
I am sorry.
But crashed again.
`2021/05/12 6:18:28
OpenBVE 1.7.560.1229-vsts Crash Log
Program is running on the Linux backend
Current screen resolution is: Windowed 960px x 600px 
The exception caught was as follows: 
System.Threading.ThreadExceptionEventArgs  at System.Environment.get_StackTrace () [0x00000] in <533173d24dae460899d2b10975534bb0>:0 
  at OpenBve.CrashHandler.UIThreadException (System.Object sender, System.Threading.ThreadExceptionEventArgs t) [0x00000] in <79b794376f9b456890af181df218a8b8>:0 
  at System.Windows.Forms.Application.OnThreadException (System.Exception t) [0x00000] in <6d635ac3dc1c4424ad385ded79f1e868>:0 
  at System.Windows.Forms.NativeWindow.OnThreadException (System.Exception e) [0x00000] in <6d635ac3dc1c4424ad385ded79f1e868>:0 
  at System.Windows.Forms.NativeWindow.WndProc (System.IntPtr hWnd, System.Windows.Forms.Msg msg, System.IntPtr wParam, System.IntPtr lParam) [0x00000] in <6d635ac3dc1c4424ad385ded79f1e868>:0 
  at System.Windows.Forms.XplatUIX11.DispatchMessage (System.Windows.Forms.MSG& msg) [0x00000] in <6d635ac3dc1c4424ad385ded79f1e868>:0 
  at System.Windows.Forms.XplatUI.DispatchMessage (System.Windows.Forms.MSG& msg) [0x00000] in <6d635ac3dc1c4424ad385ded79f1e868>:0 
  at System.Windows.Forms.Application.RunLoop (System.Boolean Modal, System.Windows.Forms.ApplicationContext context) [0x00000] in <6d635ac3dc1c4424ad385ded79f1e868>:0 
  at System.Windows.Forms.Form.ShowDialog (System.Windows.Forms.IWin32Window owner) [0x00000] in <6d635ac3dc1c4424ad385ded79f1e868>:0 
  at System.Windows.Forms.Form.ShowDialog () [0x00000] in <6d635ac3dc1c4424ad385ded79f1e868>:0 
  at OpenBve.formMain.ShowMainDialog (OpenBve.formMain+MainDialogResult initial) [0x00000] in <79b794376f9b456890af181df218a8b8>:0 
  at OpenBve.Program.Main (System.String[] args) [0x00000] in <79b794376f9b456890af181df218a8b8>:0 
Current program memory usage: 98.67mb
System memory free: 0mb
`
noconfig4

zbx1425:
I think it might be confusing if the checkbox for enabling a plugin cannot be seen when the window is maximized.
Is that to be fixed?

leezer3:
It should have been yesterday.

zbx1425:
> It should have been yesterday.

Sorry 😐...

leezer3:
Now that's odd :/

I've checked the daily built copy above in a decompiler, and it definitely has the 'right' code in it, but it crashes.
The same code built locally on Windows doesn't.

I'm presuming this must be an event loop problem, but I can't immediately see how. Will do some more thinking.....
My initial wondering would be replacing the timer with our internal timer implementation, but if it's a concurrency or race in the event loop this might not do anything anyway.

leezer3:
Might be getting somewhere.
It appears this crash is libusb related.

Please try creating the symlink as described here:
https://github.com/LibUsbDotNet/LibUsbDotNet#linux-users

ginga81:
Wow!
The TypeII detect and works!
But, the main game is trouble.

I tested v1.7.563.1082-vsts 64-bit.
I done:

cd /lib/x86_64-linux-gnu/
sudo ln -s ./libusb-1.0.so.0 ./libusb-1.0.so
reboot pc
run OpenBVE,and enable the Densha de GO device plugin
Press 'config' button
Press the link of 'My controller is not detected or does not work properly'
download 99-ps2-train-controllers.rules
sudo cp ./99-ps2-train-controllers.rules /etc/udev/rules.d/
reboot pc
connect the Densha de GO Type II joystick
run OpenBVE,and enable the Densha de GO device plugin
Press 'config' button
the all levers and buttuns (without foot switch because I cannot find this at my room) is works!

But, I run the route data, some troubles are occurs.
May be at least, we have to set any button action (for example HORN_PRIMARY) to button of A,B,C or D.

If do not do, the main window is freeze and stop all program, but I think may be 1,2,4,8,16,32 and 64 seconds waiting, 1 second per moves.
If we set to the button action the another buttons, same action of the case of not set nothing.
I wait from 1 minute, after then, freeze continue, but I think may be will stop 2 min or more.

If the button is set to A,B,C or D, the main window is freeze 1 second per 2 seconds.

cannot work P5 lever.

MarcRiera:
> Wow!
> The TypeII detect and works!
> But, the main game is trouble.
> 
> I tested v1.7.563.1082-vsts 64-bit.
> I done:
> 
> cd /lib/x86_64-linux-gnu/
> sudo ln -s ./libusb-1.0.so.0 ./libusb-1.0.so
> reboot pc
> run OpenBVE,and enable the Densha de GO device plugin
> Press 'config' button
> Press the link of 'My controller is not detected or does not work properly'
> download 99-ps2-train-controllers.rules
> sudo cp ./99-ps2-train-controllers.rules /etc/udev/rules.d/
> reboot pc
> connect the Densha de GO Type II joystick
> run OpenBVE,and enable the Densha de GO device plugin
> Press 'config' button
> the all levers and buttuns (without foot switch because I cannot find this at my room) is works!
> 
> But, I run the route data, some troubles are occurs.
> May be at least, we have to set any button action (for example HORN_PRIMARY) to button of A,B,C or D.
> 
> If do not do, the main window is freeze and stop all program, but I think may be 1,2,4,8,16,32 and 64 seconds waiting, 1 second per moves.
> If we set to the button action the another buttons, same action of the case of not set nothing.
> I wait from 1 minute, after then, freeze continue, but I think may be will stop 2 min or more.
> 
> If the button is set to A,B,C or D, the main window is freeze 1 second per 2 seconds.
> 
> cannot work P5 lever.

I can recreate this with my Shinkansen controller, which worked previously without issues. The config form is fine, but there are issues in-game. I will take a look.

Issue #608, opened 2021-04-20
[Minor]TFO: motor sound problem


enjoysoftware:
### Description

The Track Following Object XML file has a `<Decelerate>` syntax. If you use that syntax to decelerate a train, the motor sounds B1 and B2 tracks will not play.
Instead of the B1 and B2 tracks not playing, the P1 and P2 tracks are played in reverse.
Some trains have different deceleration and acceleration sounds. Such trains make strange motor noises when decelerating at the TFO.

## Route

Any routes

## Train

Any trains

### Related information 

* Windows, Ubuntu

Issue #598, opened 2021-03-25
[TrackFollowingObject]The object goes off the rail


enjoysoftware:
### Description
I'm trying to run an oncoming train using a TrackFollowingObject.
However, if I specify -1 in the `<Direction>` definition and the train is backing, and I try to change the rail, the train will come off the rail.

#### Route file
~~~csv
.....
200,.Rail 1;3.8;0;1
300,.Sta .......
450,.Rail 2;0;0;0,.Rail 3;3.8;0;0
500,.Rail 2;3.8;0;0,.Rail 3;0;0;0,.RailEnd 2,.RailEnd 3

1500,.Sta .....
~~~
### XML file
~~~xml
...
<Stop>	
        	<Decelerate>0.00</Decelerate>
        	<StopPosition>800</StopPosition>
        	<Doors>0</Doors>
        	<StopTime>00.0010</StopTime>
        	<Accelerate>3</Accelerate>
        	<TargetSpeed>60</TargetSpeed>
        	<Direction>-1</Direction>
        	<Rail>0</Rail>
      </Stop>	
      <Point>	
        	<Decelerate>3.00</Decelerate>
        	<Position>500</Position>
        	<Accelerate>0</Accelerate>
        	<TargetSpeed>50</TargetSpeed>
        	<PassingSpeed>50</PassingSpeed>
        	<Rail>3</Rail>
      </Point>
      <Point>	
        	<Decelerate>0.00</Decelerate>
        	<Position>450</Position>
        	<Accelerate>0</Accelerate>
        	<TargetSpeed>50</TargetSpeed>
        	<PassingSpeed>50</PassingSpeed>
        	<Rail>1</Rail>
      </Point>
      <Stop>	
        	<Decelerate>3.00</Decelerate>
        	<StopPosition>330</StopPosition>
        	<Doors>1</Doors>
        	<StopTime>00.0010</StopTime>
        	<Accelerate>0</Accelerate>
        	<TargetSpeed>0</TargetSpeed>
        	<Direction>-1</Direction>
        	<Rail>1</Rail>
      </Stop>	
...
~~~
If you run the above code, it will look like this video.

https://user-images.githubusercontent.com/47489953/112427989-db9f2300-8d7d-11eb-9f61-f93c66f0d108.mp4

How a train go off the rail depends on the length of the train.
This issue does not occur if I specify 1 in the <Direction> definition.
[...]

PR #609, opened 2021-04-20
Fix: multiple issues with TFO

s520:
This PR fixes multiple issues with TFO.
- Change: the TFO operation plan to be calculated only when reading the route
  - Previously, if the player jumped between stations, it would be calculated again.
- Fix: the method of calculating the acceleration of TFO
  - Calculation formula of acceleration was incorrect. (Fix: #608)

I'm currently working on fixing #598.

**UPDATE: 2021/04/21**
- Fix: TrackFollower did not consider the Invalid TrackElement when moving backwards
  - Changed to take this into consideration when moving backward as well as when moving forward. (Fix: #598)

### DIFF
--- a/source/OpenBVE/Parsers/Script/TrackFollowingObjectParser.cs
+++ b/source/OpenBVE/Parsers/Script/TrackFollowingObjectParser.cs
@@ -122,7 +122,7 @@ namespace OpenBve
 			{
 				if (Program.CurrentHost.Plugins[i].Train != null && Program.CurrentHost.Plugins[i].Train.CanLoadTrain(TrainData))
 				{
-					//Program.CurrentHost.Plugins[i].Train.LoadTrain(Encoding.UTF8, TrainDirectory, ref currentTrain, ref Interface.CurrentControls);
+					Program.CurrentHost.Plugins[i].Train.LoadTrain(Encoding.UTF8, TrainDirectory, ref currentTrain, ref Interface.CurrentControls);
 				}
 			}
 
--- a/source/OpenBVE/Parsers/Script/TrackFollowingObjectParser.cs
+++ b/source/OpenBVE/Parsers/Script/TrackFollowingObjectParser.cs
@@ -120,12 +120,18 @@ namespace OpenBve
 			AbstractTrain currentTrain = Train;
 			for (int i = 0; i < Program.CurrentHost.Plugins.Length; i++)
 			{
-				if (Program.CurrentHost.Plugins[i].Train != null && Program.CurrentHost.Plugins[i].Train.CanLoadTrain(TrainDirectory))
+				if (Program.CurrentHost.Plugins[i].Train != null && Program.CurrentHost.Plugins[i].Train.CanLoadTrain(TrainData))
 				{
-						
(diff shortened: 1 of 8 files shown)


s520:
#598  was a problem due to the position of `RailEnd` on Route CSV and `<Point>` on TFO.
Block with `RailEnd` is Invalid TrackElement.
If TFO switches the rail index of TrackFollower to the rail where `RailEnd` is written, the behavior of TrackFollower is broken because TrackFollower moving backward doesn't consider Invalid TrackElement.

We need to find out that this change does not destroy anything else.

leezer3:
I don't think it should.

The original code only uses TrackFollowers on Rail0.
We've added multiple rails to the equation, and this / the TrackFollowerFunction are the only places that use it.

Possibly need to think about junctions though.
https://github.com/leezer3/OpenBVE/pull/329

That is an incomplete branch, but perhaps we need to think about a RailIndexChange event, as opposed to switching then in the TFO itself?

s520:
Hmm... I'll have a look at that branch and think about it.

leezer3:
It's not so much for this bug fix, as opposed to a later improvement.
(and the branch is seriously out of date with master)

The switch logic would need to be coded into the routefile, and then I suppose you would need time based switch events (?) in the TFO plan to set them as appropriate, and then reset afterwards.
The advantage would be getting the sim closer to operating as per a 'real' railway, bit on the flip side it'd require a rework of the TFO concept & likely not be backwards compatible.

s520:
I feel bad. I can not grasp the intent of your remarks.
Are you negative to https://github.com/leezer3/OpenBVE/pull/609/commits/fd346449c1e90a4b196800f9e162d8c72b982090 itself, or are you negative because it may affect more than just TFO?

leezer3:
None if those, nothing wrong with your code I can see.

I was referring to the potential effects of the linked incomplete branch.

s520:
Thank you. I seem to have misunderstood.

I have fixed the additional issue pointed out [here](https://github.com/leezer3/OpenBVE/issues/598#issuecomment-824702428).

leezer3:
No problem, thanks :)

s520:
I forgot to uncomment the test.
This PR is now ready for review.
Please review it.

leezer3:
I've added a couple of small improvements:
https://github.com/leezer3/OpenBVE/commit/ebad2e4ae78f32f9495126a8a68dd4c2ac4cfd60
https://github.com/leezer3/OpenBVE/commit/fbf1d43e2e69e685840190968ee22c2446d6c6fd

These are intended to address the following situation:
A train contains a train.dat file, but which has an unrecognised BVE format string in. (I'm not aware of any, but I've only found about 2 of the **BVE1210000** format files, and I don't want to rule out others, especially as my collection is more Western focused)
With the changes from this PR, we completely failed to load the train &  subsequently crashed attempting to assign AI.

The first disables the start panel if the train format is unsupported.
The second adds another format- **UnknownBVE** & converts the new function into an enum return as opposed to a void (neater I think).

It might want a check for non-zero cars / handles here, or a check that the loading of the train succedded:
https://github.com/leezer3/OpenBVE/blob/master/source/OpenBVE/System/Loading.cs#L299

s520:
I see. Thank you for fixing it.

PR #624, opened 2021-04-28
Fix: some issues with the path of texture file in DirectX file

s520:
- Fix: a text encoding issue in Assimp X file parser
  - Assimp X file parser didn't consider the character encoding of the text format.
- Fix: the issue where some object files specify absolute paths
  - For example, the texture is specified as an absolute path on the 1490, 1499, and 1532 lines of `railway\object\marumado\5250.x` on the route published [here](http://marumado.umic.jp/marumado_go_download.php).


leezer3:
Thanks!

Have fixed the original X parser to handle the Marumado objects and these type of texture references for good measure.

PR #629, opened 2021-05-10
Fix: the issue of incorrect version display in the installer version

s520:
Since we started to generate the installer with Azure Pipeline, the version information of the main program is incorrect.
https://bveworldwide.forumotion.com/t2100-wrong-version-name-in-about-page-in-installer-version

This PR fixes the issue.

The following content was my misunderstanding. sorry.
~~Also, the version information of the installer itself is corrupted.~~
~~In the part below, it seems that the script is not given version information.~~
~~https://github.com/leezer3/OpenBVE/blob/3ed0b0d5050e697b18a01813d9090cd74d433c28/azure-pipelines-release.yml#L36-L39~~

~~It seems that `$(git describe --tags)` is not evaluated at run time.~~
~~https://dev.azure.com/leezer3/OpenBVE/_build/results?buildId=392&view=logs&j=a29a2371-0793-5e14-39b9-20eb0beef0db&t=44cf1151-9065-5991-d458-54cd926c5c60&l=10~~


s520:
I misunderstood because the file version of the installer properties is 0.0.0.0.
This hasn't changed.

s520:
I changed it to try the default way of naming the version if the git command fails. (git doesn't exist, etc.)

PR #631, opened 2021-05-12
Fix: some issues in ObjectViewer

s520:
This PR fixes some issues in ObjectViewer.
- Fix: the issue that the train setting is not reflected in the train load from extensions.cfg
  - It stopped working after we changed the train loading function to a plugin.
- Fix: the issue where extensions.cfg is not loaded again after changing options (F8)
  - This is the issue that has existed since extensions.cfg was able to load.
- Fix: the issue where some saved options aren't reflected at startup
  - AntiAliasing Level and Anisotropic Filtering Level were overridden at startup.

Also, some project dependencies were missing, so I fixed them.

**UPDATE: 2021/05/24**
- Add: mode of optimization performed on the object in ObjectViewer
  - I have added three options: none, low (viewer's default) and high (main program's default).
- Fix: the issue of deformation of Extensions.cfg in the Z-axis direction
  - `TranslateZFunction`, `RotateZFunction`, etc.
- Fix: the issue where train setting window don't close at the end of the program
  - This is a issue that can occur on Linux.
- Add: the function to shift focus to the train setting window that is already open
- Change: the setting values are kept even if the simulation is disabled
  - Fix: #632
- Change: actually simulate the brake pressure when the train is loaded from extensions.cfg
- Add: the renderer statistics in ObjectViewer
  - Close: #637


leezer3:
Thanks!

Will take a proper look later.

s520:
OK, I understand.
When I have implemented the features suggested by the Japanese community, I would like to look into OpenGL Main Menu.

s520:
Currently, this PR fixes part of #632.
After reloading, you need to press the Apply button again.
It seems that it was applied automatically before.
So I need to fix this PR a little more.

leezer3:
Looking at it, I'm not actually sure we have a good solution there without some quite radical redesign.
Semantically at least, my initial thoughts would be that the train.dat (or alternative) should be the source of the values.

Some possible thoughts:
* Disable the Enable Train checkbox when loaded as train?
* Perhaps things such as the brake type, number of notches etc. should then be locked.
* Allow altering of the driver notch, and from there 'fix' the brake cylinder etc. so that it returns real simulation data, rather than the current dummy values.

s520:
The current PR locks some parameters (the brake type, number of notches etc.) when the object is loaded from extensions.cfg.
In addition, when disabling the Enable Train checkbox when loaded as train, I changed it to initialize only unlocked values.

Therefore, I think that our opinion is almost the same up to the second point.
I need to think more about the third point.

s520:
I changed so that the setting values are kept even if the simulation is disabled. (Fix #632)
If you want to initialize the setting values, click the reset button.

leezer3:
As always, thanks.

Please say when this is finished; We could probably do with another small point release before the GL Menu and animated GIF code are added.

s520:
OK.

By the way, Is `emergencyReservoir` in [the document](https://github.com/leezer3/OpenBVE-Documentation-Hugo/blame/master/content/en/objects/native/animated/_index.md#L528-L529) a typo for [`equalizingReservoir`](https://github.com/leezer3/OpenBVE/blob/619b8728127dfdd36e8a37d373741ab4309ebff0/source/OpenBveApi/FunctionScripts/FunctionScript.cs#L669-L677)?
If this is a typo, I need to fix the label of the train settings window.

leezer3:
Now that's interesting.

The original documentation definitely uses emergencyReservoir:
https://github.com/leezer3/OpenBVE/blob/3fd4d729564fb4de83c6494d0aa929c5984b18f5/documentation/HTML/object_animated.html#L440

I agree though this looks like a typo, albeit a very old one.
There's definitely no **emergencyreservoir** in any code, and using it results in an error, so I think it can be safely got rid of.

s520:
Ready for review.
Please review this PR.

PR #633, opened 2021-05-13
Fix: the issue where daily build would work triggered by PR

s520:
This PR fixes the issue where daily build would work triggered by PR.
I thought it wasn't intentional behavior.
If it is intentional behavior, please close this PR.

### DIFF
--- a/azure-pipelines-release.yml
+++ b/azure-pipelines-release.yml
@@ -8,6 +8,8 @@ trigger:
     exclude:
       - '*'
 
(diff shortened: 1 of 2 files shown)


s520:
It seems that the release build was also working.
https://dev.azure.com/leezer3/OpenBVE/_build?definitionId=2

s520:
It looks like we need to change `condition: and(succeeded())` to `condition: succeeded()`.

Issue #614, opened 2021-04-25
TouchElement cannot work at 1.8.0.1 and crash OpenBVE


ginga81:
### Description

When a TouchElement press, OpenBVE is crash.
OpenBVE version is 1.8.0.1.

### Reproduction

Route:Any. But you can use bundled sample route.
Train: Please use this sample Train.

## Route

Any. But you can use bundled sample route.

## Train

 Please use this sample Train.
[tutorial-train-1.8.0.1-touch-crash.zip](https://github.com/leezer3/OpenBVE/files/6371559/tutorial-train-1.8.0.1-touch-crash.zip)

## Logs

2021/04/25 17:23:41
OpenBVE 1.8.0.1 Crash Log
Program is running on the Linux backend
Current screen resolution is: Windowed 960px X 600px 
Current routefile is: ---/A0直線デモルート.csv
Current train is: /Train/Tutrial-train
Current train plugin is: OpenBveAts.dll
1 error(s)
The current routefile caused the following exception: 
System.NullReferenceException: Object reference not set to an instance of an object
  at OpenBve.Graphics.Renderers.Touch+<>c__DisplayClass16_0.<LeaveCheck>b__0 (System.Int32 x) [0x00004] in <c1a242e00db74bdeb84d35e7431d4b1d>:0 
  at System.Linq.Enumerable+WhereArrayIterator`1[TSource].MoveNext () [0x0002c] in <5b415632df1f4365ae2242b1a257bb5b>:0 
  at OpenBve.Graphics.Renderers.Touch.LeaveCheck (OpenTK.Vector2 Point) [0x0017b] in <c1a242e00db74bdeb84d35e7431d4b1d>:0 
  at OpenBve.MainLoop.mouseUpEvent (System.Object sender, OpenTK.Input.MouseButtonEventArgs e) [0x0003f] in <c1a242e00db74bdeb84d35e7431d4b1d>:0 
  at (wrapper delegate-invoke) System.EventHandler`1[OpenTK.Input.MouseButtonEventArgs].invoke_void_object_TEventArgs(object,OpenTK.Input.MouseButtonEventArgs)
  at OpenTK.NativeWindow.OnMouseUp (OpenTK.Input.MouseButtonEventArgs e) [0x00000] in <791d08df1a7948a4b0f6aebfd6cbb1ae>:0 
[...]

Issue #564, opened 2020-12-14
Not reflected the Informations Default Displaying Plugin (for testing)'s parameters.


ginga81:
### Description

I set this screenshots settings.

![勾配などが反映されない](https://user-images.githubusercontent.com/22734699/102114048-1dfebd00-3e7d-11eb-9881-d50bfc55dd55.png)

But, not reflected these parameters.
When I set to all parameters, but the display is 
I set to the clock, speed is kmph, DistNextStation is Km, and fps.
But the displaying is stop is miles, gradient and speed is mph.

![Screenshot from 2020-12-15 02-10-29](https://user-images.githubusercontent.com/22734699/102114911-1ee41e80-3e7e-11eb-9166-2ef308bc583e.png)

OpenBVE's version is
v1.7.416.1319-travis
1.7.2.0 is correctly.

Could you please confirm this?

### Reproduction

This is some options trouble, so the route is any.

## Route

This is some options trouble, so the route is any.
So, no links.

## Train

The train is any.

## Logs

This issue is no crash, so no logs.

### Related information 

My PC's SPEC:

OS: Ubuntu18.04LTS(64bit)
Memory:31.4 GiB
CPU:AMD® Ryzen 7 2700x eight-core processor × 16
Graphic board:Radeon RX 590 Series (POLARIS10, DRM 3.33.0, 5.3.0-51-generic, LLVM 9.0.0)

leezer3:
Will take a look.

I have a suspicion this might need another setting enabling, but I can't remember off the top of my head.

ginga81:
This trouble has been fixed at 1.8.0.2.
So this issue is closed.

PR #636, opened 2021-05-15
Fix: Issues with DenshaDeGoInput on hosts with missing LibUsb symlink

leezer3:
I'm putting this as a PR / proposal at the minute, as it makes a breaking change to the InputDevicePlugin interface.
I don't believe that there are any other InputDevice plugins available at present, so this is relatively low risk.
The non-breaking alternative would be to keep a separate set of platform detection code in DenshaDeGoInput as previously.

### Changes:
* Adds a HostInterface to the InputDevice plugin load call. I've only used this for platform detection at the minute, but I was thinking of error returns if / when a failure mode is encountered.
* Wraps the LibUsb code in a try / catch block to prevent it taking down the main form.
* Detects DllNotFoundException in this try / catch, and changes the help window troubleshooting text appropriately.

### Potential Improvements:
* Log invalid data packets etc. in the main game log (hence the addition of the host interface)

cc @MarcRiera


s520:
For reference, there are other input device plugins available.
https://github.com/TetsuOtter/BIDSSMemLib

leezer3:
In that case, we don't want to alter the interface.

Pity, will have to see whether I can come up with a backwards compatible plan.
Ought to link that plugin (and any others) from the main site too.

s520:
That's the only publicly available plugin I know.
In addition, there is one private plugin. (https://twitter.com/zhang_bx/status/1390990728394805250)

MarcRiera:
I've tested this with my controllers and noticed no regression. LibUsb controllers, however, seem to glitch in-game (the main program freezes and the controller is unresponsive). This is unrelated to this PR, as it already happened before. I need to investigate, but it could be related to the recent internal improvements.

Regarding the interface change: we recently added support for SetSignal, SetBeacon and doors without issues, as they were added to the newer ITrainInputDevice instead of IInputDevice. The latter is what TestuOtter's plugin uses, and the former was initially created to accommodate specific needs of DenshaDeGoInput. I'm not sure which one is using the private plugin, but if we apply the improvement to ITrainInputDevice I think it should be fine. That'd leave us with two different interfaces that should probably become one, but we'd have time to notify TetsuOtter and the other developer about the necessary changes and even ask them if there's anything else they need.

leezer3:
Sounds reasonable-ish.

FWIW, I'm also considering moving the filesystem to be a host interface property by default.
It's the logical ish place for it to live as a property of the host, but on the flip side we risk bloating that interface.....

leezer3:
Reverted the host interface change. 
I couldn't see a decent place to put it without adding a new method, so I've resorted to pulling it out of the filesystem via reflection instead. Not exactly brilliant, or something I'd really reccomend, but there we go.

I've also wrapped all our libusb calls in try / catch blocks, and added the error logging which was the other reason I was thinking of the HostInterface.
It'd be marginally interesting to see if this now stops your freezing (I wouldn't expect it to mind)

MarcRiera:
> Reverted the host interface change.
> I couldn't see a decent place to put it without adding a new method, so I've resorted to pulling it out of the filesystem via reflection instead. Not exactly brilliant, or something I'd really reccomend, but there we go.
> 
> I've also wrapped all our libusb calls in try / catch blocks, and added the error logging which was the other reason I was thinking of the HostInterface.
> It'd be marginally interesting to see if this now stops your freezing (I wouldn't expect it to mind)

Interestingly, it has stopped freezing in-game. No apparent feature regression (disconnecting and reconnecting the controller still works). I'm not entirely sure why it stopped working in the first place, I guess a connection error during polling somehow stopped the function and created the freeze. In any case, given that the current code is safer with try/catch, I'd treat it as fixed. Let's see if @ginga81 can test it as well with the Type II.

leezer3:
Odd.
That makes me think timing or concurrency issue.

Maybe add a lock / timeout to the libusb calls...

MarcRiera:
> Odd.
> That makes me think timing or concurrency issue.
> 
> Maybe add a lock / timeout to the libusb calls...

Currently, we ask the device for input synchronously and wait indefinitely. This takes place on a task and the main loop can only ask for input again once the task is finished (to avoid asking repeatedly, as the device only answers when there's a state change). A timeout in the LibUsb.Read function wouldn't hurt, but I wouldn't make it too low; otherwise, it might be a bit of a waste.

ginga81:
Can I this at nightly build?
Or do I build from source?
If I will build, where the source code from download?
I cannot found 'Download zip'.

leezer3:
> Currently, we ask the device for input synchronously and wait indefinitely. This takes place on a task and the main loop can only ask for input again once the task is finished (to avoid asking repeatedly, as the device only answers when there's a state change). A timeout in the LibUsb.Read function wouldn't hurt, but I wouldn't make it too low; otherwise, it might be a bit of a waste.

But is this thread safe is the question?
We've got both the controller update and the polling for connected devices. The connected devices poll is on the main thread, whilst the controller update is on it's own task & therefore separate thread.

As I read the [libusb documentation](https://www.cs.unm.edu/~hjelmn/libusb_hotplug_api/mtasync.html) accessing from different threads, one *may* be oblivious when an event arrives depending on the timing, and whether the other holds the filesystem lock.
This sounds somewhat like what was happening, in that random events appear to have been eaten. 
I haven't looked into libusb.net to see what it does about concurrency. 

Thus, we might want a lock and a wait-loop in the controller update thread?

---
### Other stuff:
I think we need to add libusb to the debian package dependancies. 
We could probably also add the udev rule into that too to make install nearly seamless. I don't think there's anything against shipping a udev rule in a deb, certainly other stuff does it (and presumably there's nothing else using the controller on Linux we could upset with this?)

--- 

@ginga81 No nightly builds at the minute, you'd have to build this branch from source :)

ginga81:
I tested before 
https://github.com/leezer3/OpenBVE/pull/636/commits/cae36d549b09fc0ee44b21d18af56678be60c317
The test window's status is OK.
The response is very nice.
The game window is stops at least 1 minute or more.
If the case of bad, the train moves about 1 sec per 1 min.
If the case of good, the train moves about randomly 1 sec or more.
This case's random waiting time is 0 sec or more about 2 sec.
I cannot debuging run from monodevelop. 
If run from monodevelop, unhandle exception occurs.
So, build only.

MarcRiera:
> 
> But is this thread safe is the question?
> We've got both the controller update and the polling for connected devices. The connected devices poll is on the main thread, whilst the controller update is on it's own task & therefore separate thread.
> 
> As I read the [libusb documentation](https://www.cs.unm.edu/~hjelmn/libusb_hotplug_api/mtasync.html) accessing from different threads, one _may_ be oblivious when an event arrives depending on the timing, and whether the other holds the filesystem lock.
> This sounds somewhat like what was happening, in that random events appear to have been eaten.
> I haven't looked into libusb.net to see what it does about concurrency.
> 
> Thus, we might want a lock and a wait-loop in the controller update thread?

In the plugin there is the main loop (`OnUpdateFrame()`) which checks which controllers are connected, calls specific code depending on the model (OpenTK.Input for PS1/PC and LibUsb for PS2) and sends input to the main program. In the case of PS2 controllers, the specific code is simply running the polling task and reading input from a byte array, which is updated by the task as data arrives. Besides the timeout we mentioned earlier, this would seem fine.

However, after reading the documentation, the issue may be happening not when asking for input (in the task, which in theory should not lock the main program even if it waits indefinitely, because it is on a different thread) but right before, when checking which controllers are connected. This is also done via LibUsb on a function called directly from the main loop. The function tries to open a device to see if it exists, which means it is connected. If this function is getting blocked because the device is busy being asked for input with no timeout, it may well be the origin of the freezes.

A lock sounds like a good idea, and according to the documentation it should be possible to allow the main loop to check the controller status. But it may be easier to move everything to a task that handles everything at once and exposes the data/connection status to the main loop, like what OpenTK.Input does.

> ### Other stuff:
> 
> I think we need to add libusb to the debian package dependancies.
> We could probably also add the udev rule into that too to make install nearly seamless. I don't think there's anything against shipping a udev rule in a deb, certainly other stuff does it (and presumably there's nothing else using the controller on Linux we could upset with this?)

This seems fine to me. In any case, all the udev rules do is change the device permissions to allow non-root raw access. There is no system driver change, everything is done in userspace.

leezer3:
@ginga81 @MarcRiera 
Any chance of building / testing the latest commit?
I might have gone overboard with the locks, but better safe than sorry.....

ginga81:
I build and test the newest branch.
If we start the game after displaying the connection test screen, it will stop while loading data and OpenBVE will not be able to play
We can play by starting up and playing without a connection test, but every time you operate the controller, it moves 'one frame' and waits.
While we do not move the controller's lever, all of OpenBVE process will stop and the keypress will not be accepted.
However, the DenGoController's reaction speed is good when moving.

leezer3:
@MarcRiera 
I've hacked this into a single thread input loop instead, plus re-written a few bits.
It *probably* works, but I've not got one of these things. It doesn't have the initial display / pedal state setup you had either at the minute.

MarcRiera:
> @MarcRiera
> I've hacked this into a single thread input loop instead, plus re-written a few bits.
> It _probably_ works, but I've not got one of these things. It doesn't have the initial display / pedal state setup you had either at the minute.

I've just tested it and on startup the Shinkansen controller goes crazy with the rumble and the display turns on. Probably a side effect of changing the initial setup for the data packets. Once I fix it I'll take a look at how it works in-game.

By the way, I'm unable to start a game if I build the whole main program with the latest commit from this branch. I've had to build the input plugin and used it in binary form on a nightly build from May 11th. On the route and train selection screen, I can select a route but the default train doesn't appear and train selection doesn't work. Browsing any train manually simply does nothing, the train is not loaded and the Start button is greyed out. I suspect it might have something to do with this: https://github.com/leezer3/OpenBVE/pull/635

leezer3:
That wouldn't cause that effect.
It's only in the panel parser (and not even triggered by the preview loading code)

Nothing in here that should cause that sort of effect either, it's only changes to the plugin.
Are you certain you tried a rebuild all, as opposed to just a branch checkout & rebuild?
I've seen that cause some odd issues.

MarcRiera:
> That wouldn't cause that effect.
> It's only in the panel parser (and not even triggered by the preview loading code)
> 
> Nothing in here that should cause that sort of effect either, it's only changes to the plugin.
> Are you certain you tried a rebuild all, as opposed to just a branch checkout & rebuild?
> I've seen that cause some odd issues.

My fault. I've cleaned up the project folder, rebuilt everything and tracked down the issue: there's something wrong with the LibUsb thread that is started on plugin load. With the plugin disabled, no issues. With the plugin enabled, unable to select train. Commented the line starting the thread and again, no issues.

With the thread enabled, the controller is detected by the plugin, but the model is not identified (the function doing that is in InputTranslator, because initially the LibUsb code was called from there). This causes the rumble/display issues, because LibUsb isn't sending the proper byte array to the controller. Not sure if it's related to this, but the controller isn't listed in the config form either.

Currently we support just two PS2 controllers. Next week I'll receive the third PS2 controller, which is undocumented, and I'm in contact with someone who has a MTC controller (from another Japanese series of games, Train Simulator, but closely related to Densha). That controller has swappable cassettes that change the amount of notches and each one changes USB behaviour (even VID/PID change!). Given the potential for these ~6 new devices, all via LibUsb, to make things even more complicated, I was thinking about splitting the PS2 controller code in two:

1. LibUsb code: loop in a thread/task doing everything affecting the USB devices in one go. Check if that device is connected, read input and send output. Instead of hardcoding controller-specific parameters here, make code generic and pass any parameter from the PS2 code.
2. PS2 code: controller-specific details (which bytes should be received, which bytes should be sent, capabilities, etc.).

What do you think? It's some work, but I could give it a good push during the weekend and I'll probably find out what's wrong with the thread.

leezer3:
That would seem logical to me.
Abstract controller class, and then specific implementations?

Interesting that the thread is locking up, it doesn't do that on Windows although that's with no supported controller.....

MarcRiera:
> That would seem logical to me.
> Abstract controller class, and then specific implementations?
> 
> Interesting that the thread is locking up, it doesn't do that on Windows although that's with no supported controller.....

Yes, abstract controller class expanding what you already added in `Ps2Controller.UsbController.cs`. In the PS2 code we'll then create a list of supported controllers with settings (VID, PID, bytes to be read, etc.) that will be passed to LibUsb. Just like currently, only one device of each type at the same time. I'm still not sure how I'll sort changing between different models of LibUsb controllers, but I'll have another controller next week to test this.

leezer3:
Ultimately, you're the one with these things in your hands and implementing it :)

I tend to prefer the abstract model with derived classes, as that's the 'correct' C# way to do it (and normally makes adding something else easier down the line), but at the end of the day do it however you want to!

leezer3:
@ginga81 see if this branch now works OK for you after Marc's changes.
You're the only other person with one of these :)

ginga81:
I tested the this branch, and I have Ryojo-con, so also this tested.
Even if I enable the DenGo plugin and save it, it will be disabled every time at end of the game.

In rare cases, the response does not reach the notch specified for 1 second, but it is generally smooth.
The pedal works normally by short-circuiting the 3.5mm analog jack and testing it as a substitute for the pedal.
At the game window, P5 and B8 notch is not work, but config's status test is correctly.

If you return to the menu after driving once, the operation will not be accepted
At this time, even if the device plugin remains enabled, it will not be recognized.
To enable, have to re-execute OpenBVE.

Also, I have Ryojo-con USB joystick, so I connected.
I re-installed new udev.rules.
It works perfectly with horn pedal!
I think a bit tremling B0 and B1.
So, the power notch is off at the every trembling.

From B0 to B1's area is a bit narrow for me.
At the game window, Ryojo-con's P4 notch is no responce. but at the config'test is shown and corrctly.

MarcRiera:
Thank you for testing!

> I tested the this branch, and I have Ryojo-con, so also this tested.
> Even if I enable the DenGo plugin and save it, it will be disabled every time at end of the game.

I can recreate this, but I'm not sure why it happens. I'll take a look.

> In rare cases, the response does not reach the notch specified for 1 second, but it is generally smooth.
> The pedal works normally by short-circuiting the 3.5mm analog jack and testing it as a substitute for the pedal.
> At the game window, P5 and B8 notch is not work, but config's status test is correctly.

Which train are you using to test the Type II? If the train has less than P5/B8 notches, it will not reach P5/B8.

> If you return to the menu after driving once, the operation will not be accepted
> At this time, even if the device plugin remains enabled, it will not be recognized.
> To enable, have to re-execute OpenBVE.

I will test this again.

> Also, I have Ryojo-con USB joystick, so I connected.
> I re-installed new udev.rules.
> It works perfectly with horn pedal!
> I think a bit tremling B0 and B1.
> So, the power notch is off at the every trembling.
> 
> From B0 to B1's area is a bit narrow for me.
> At the game window, Ryojo-con's P4 notch is no responce. but at the config'test is shown and corrctly.

Again, it's possible the train has less than P4 notches. Regarding the "trembling", I'll make the B0 notch a bit wider.

ginga81:
This sample has P8B8.
https://github.com/leezer3/OpenBVE/issues/614
B8 is the another train using as you say, hadn't B8. sorry.
B8 is works well.
Ryojo-con's maximum brake is B6. no EB. 
Also this sample train, not work Ryojo-con's P4.

Option settings memory is now working.
But ,at the past time, this build wasn't memory settings.

And this build has the another issue.
Informations Default Displaying Plugin(for Testing)'s displaying settings is different.
This was fixed at the past, but happens again.
https://github.com/leezer3/OpenBVE/issues/564

At once this build is run, after then, the all OpenBVE build is this display settings. Even then, options_ddp.cfg is not change.
To recovery, I have to run 1.7.1.8 and have to a route play.
I don't know why.

I doubt that the option settings memory issue is maybe relate this troble.

PR #641, opened 2021-05-29
LibUsb fix/rework

MarcRiera:
This makes LibUsb controllers work again and reorganizes some stuff.

Done:

- Make Type II/Shinkansen controllers work again.
- Expand generic class for LibUsb controllers and put controller-specific settings on top of it.
- Add timeout when reading input data.
- Rework controller code into classes
- Fix Unbalance controllers on Windows
- Add support for Ryojouhen controllers


MarcRiera:
This is ready to merge (and #636 will be ready too).

I've created a generic controller class used by all models, which simplifies a lot of things. Each model then builds upon this class with model-specific code. The LibUsb code is generic and the PS2 controller class passes the required parameters for each model.

I found two issues which are now fixed:

- The GUID format for OpenTK joystick changes between OS (it's created by SDL2 and it was unified on some version but I think the OpenTK version we use is based on a version before the change). This caused Unbalance controllers to be completely undetected on Windows. I've added a platform check; if we ever upgrade OpenTK, we'll be able to simplify the code.
- We generated a LibUsb issue if there was a problem reading input from the controller. Instead of stopping the LibUsb thread completely in this situation (which doesn't cause a real exception), we consider the controller "disconnected". This allows disconnecting and reconnecting the controller in-game (it originally worked, but the recent changes prevented it).

leezer3:
OpenTK is the latest. (Version 4 onwards pulls out the entire Linux / Mac native backends in favour of GLFW. It's still not particularly pretty several years after having done that)
I'll take a look at the GUIDs in there assuming I remember.

PR #644, opened 2021-06-04
Fix: the issue where station information is not displayed

s520:
This PR fixes the issue pointed out at the following URL:
http://bveworldwide.forumotion.com/t1876-routeviewer-not-showing-station-info


leezer3:
Thanks!

This fixes the issue, but it got me wondering if the station index should actually be stored as a property of the track-follower itself?
This would allow us to test whether we're in the station in animation functions, which could be quite useful for trains.
For example, only open the doors on cars which are in the station platform.

The trouble is with that, I can't see a good way to implement it.
https://github.com/leezer3/OpenBVE/blob/master/source/OpenBveApi/Routes/Events.cs#L29

We *could* change the event parameters to take a TrackFollower instead of the train, car etc
This though would involve a re-work of all the events, and would add quite a few additional little bits of code.

s520:
I agree with your idea of taking TrackFollower as a parameter.
There will be more code changes, but it will be worth doing.

s520:
I added current station index to TrackFollower.
Please check it.

leezer3:
Looks good to me, thanks!

PR #653, opened 2021-06-13
Additional fixes for DenshaDeGoInput

MarcRiera:
Fixes for additional issues mentioned in #636.

Implemented:
- Unload input device plugins after saving the program's settings. Otherwise, they are marked as disabled and not saved to the list of enabled plugins (affected all plugins).
- Tweak Ryojouhen controller notches (should fix B0/B1 and EMG).
- Properly unload LibUsb controllers so they're not locked when going back to the main menu

To do:
- Possible issues with certain notches?

@ginga81 Could you please try this branch? OpenBVE will now remember the plugin is enabled and the Ryojouhen brake handle should work better. I haven't found any issue with P4, it works fine on my PS2 Ryojouhen controller with the sample train you provided before.


ginga81:
May be my fault, I re-download zip and compiled.
But, still happen all troubles.
I don't know how to the  newest blanch download...
I am very sorry...

MarcRiera:
@ginga81 Use this: [OpenBVE_DenshaRefactor.zip](https://github.com/leezer3/OpenBVE/files/6644262/OpenBVE_DenshaRefactor.zip)

ginga81:
Thank you very much!
I tested this compiled version.
But at this version, Type-II's P5 and Ryojo-con's P4 wasn't action.
EB is works at testing window with Ryojo-con, but game window wasn't.
The testing status window's setting and enable remain is maybe correct.

MarcRiera:
I've found and fixed the issue when going back to the main menu. It should work now: [OpenBVE_DenshaRefactor_v2.zip](https://github.com/leezer3/OpenBVE/files/6644592/OpenBVE_DenshaRefactor_v2.zip)

The only issue left are the notches. If I understand correctly, in the config window everything is correct, but in-game it isn't. This makes me think it's related to the train and the plugin settings.

- Convert notches: if disabled, the train matches the controller (P1 to P1, P2 to P2, etc.). There is no conversion. For example, with a P8/B8 train and a Ryojouhen controller, the limit is P4/B6. If enabled, the train skips some notches. The power handle would be Released>P2>P4>P6>P8, for example.
- Keep minimum and maximum: when converting, the minimum and maximum notches are respected. With the same example, the power handle would be Released>P1>P4>P6>P8.
- Map hold brake to B1: when enabled, if the train has a hold brake, it is always on B1.

Hope this is clear. Thanks for helping debug this!

ginga81:
I couldn't understand what these options means.
But now, by your information, I understand completely!

>The only issue left are the notches. If I understand correctly, in the config window everything is correct, but in-game it isn't. This makes me think it's related to the train and the plugin settings.

As you say, in the config window, works perfectly, but in-game is not.
If the options of 'convert notch' is disable, Type-II cannot assign P5.Only works P0 to P4.
As the same, Ryojou-con's P4 was't work. From P0 to P3.

At the another all options, as your says, it works perfectly at the Type-II and Ryojouhen controller!

You are  welcome!
I am the only one tester of  the world!
I am honored!

ginga81:
Sorry, I have a question about Ryojou-hen controller.
https://www.youtube.com/watch?v=wwjrXV_C2gU
Ryojou-hen controller's blake is such as this video.
The brake lever's basic position is center.
Move to the left, brake pressure is down and can drive.
Move to the right, brake pressure is up and stop.
But currently, this brake is such as two-handle type's.
How to change to as this tram's brake?

MarcRiera:
> At the another all options, as your says, it works perfectly at the Type-II and Ryojouhen controller!

Great to hear it's working!

> Sorry, I have a question about Ryojou-hen controller.
> https://www.youtube.com/watch?v=wwjrXV_C2gU
> Ryojou-hen controller's blake is such as this video.
> The brake lever's basic position is center.
> Move to the left, brake pressure is down and can drive.
> Move to the right, brake pressure is up and stop.
> But currently, this brake is such as two-handle type's.
> How to change to as this tram's brake?

Currently it's not possible to use it like this. The controller behaves as in Densha de GO! Professional 2 and Densha de GO! Final, with 6 brake notches. I will add it later (separate PR). Thanks for the suggestion!

@leezer3 This is ready to merge, everything is now working as expected for both of us.

PR #675, opened 2021-07-22
New: Package install for OpenGL menu

leezer3:
This PR starts implementing a package installer for the OpenGL menu.
It uses the same file browser logic as per the main menu, but shows all files & pulls out the associated Windows icon (as whilst the 'default' extension for a package is ZIP, some people have used OBP etc.)

### Changes:
* New _PackageManagement_ GL submenu.
* Installs packages (no dependancy management)
* **OperationCompletedEvent** raised by API when a package operation completes, as opposed to relying on Winforms worker results.
* Adds scroll cursor and scroll bars to GL textboxes

### TODO:
* Test on Apple- Is the method used to pull out our icons working here? (suspect not)
* ~~Hook up the uninstall menu.~~
* Add clicky scroll targets to GL textboxes, as well as mousewheel scroll.
* Port the WinForms menu to use the new **OperationCompletedEvent** so we do things the same everywhere.
* Mouse movement in GL textboxes should likely be handled via an internal method as this is much more re-usable elsewhere than in the menu.
* **?? Probably Separate PR ??-** The options menu needs adding, so we can control where we're putting this stuff on the filesystem.


leezer3:
packagemenu

Basic previews added.
Still very much WIP, no installation or anything yet. Need to add these, before thinking about an installed list and uninstall. Dependancies *may* be complex, as this will (probably) require a new menu type.

Note that the description issue is because this is narrower than the WinForm install dialog. I obviously used newlines when I created this one in order to make it fit right.

leezer3:
Somwhat getting there.

I've added an operation completed event to the packages API, and basic extraction. This ignores dependancies, reccomendations etc.
Next job will be to actually use the database, and to add a view installed option.

This also (marginally) depends on the Options menu being added, so that people can choose where they want to put stuff.
Screenshot supplied in https://github.com/leezer3/OpenBVE/issues/680 also suggests that this might want testing at much larger screen resolutions for scaling issues.

Other marginal needs:
*really* wants a scrollbar or a couple of arrows on the textbox, or at least changing the cursor to the scroll icon when above it.

leezer3:
packagemenu2

Doesn't look like much, but having a scrollbar and appropriate mouse cursor makes the description textbox much more usable.

leezer3:
This is now reasonably functional for basic package operations.

### Missing Bits:
* Dependancy management
* Warning if package already exists
* Polish
* Related options (To be added in separate GL menu options PR)

PR #682, opened 2021-08-03
[DenshaDeGoInput] Fix button repeats and improve brake behaviour

### LINKED ISSUE #681: long horn eternaly continue 0.1ms sounds while the Densha-de Go! plugin enable
### Description

When we are using a long horn sounds using train, and when we are enable Densha de Go plugin settings is enable (If the controller is connect or not, the result is the same), the long horn sound is continue to sound.
This sound is long about 0.1ms per 1 sec.
And while this issue is happening, the long horn settings keyboard cannot sound.
the another normal keyboard sound sounds correctly.

Normally, even the plugin is enabled, the horn have to sounds when we are only keyboard pressed.
But, even we are not press keyboard, this issue is continueing the long horn sounds.
And, while enableing the Densha de Go plugin, long horn keyboard is not work.

### Reproduction

Train:
Please use F81_tec200's Series-E4.
https://www.youtube.com/watch?v=-bjdICkOPD0

## Route

Any

## Train

Please use F81_tec200's Series-E4.

## Logs

This issue is not crash.
So no any logs.

### Related information 

My PC's SPEC:

OS: Ubuntu18.04LTS(64bit)
Memory:31.4 GiB
CPU:AMD® Ryzen 7 2700x eight-core processor × 16
Graphic board:Radeon RX 590 Series (POLARIS10, DRM 3.33.0, 5.3.0-51-generic, LLVM 9.0.0)

Ihave...
Densha de Go Type-II (USB)
Densha de Go Ryojo-hen controller(USB)
(But, this isuue is not need these)

MarcRiera:
Previously, the plugin handled button repeats internally. This apparently worked well but caused issues like #681 and the code could be simplified. This PR:

1. Closes #681.
2. Improves support for trains with an automatic air brake. When using these trains, the brake handle will reflect the real behaviour more accurately.

### DIFF
--- a/source/InputDevicePlugins/DenshaDeGoInput/Config.cs
+++ b/source/InputDevicePlugins/DenshaDeGoInput/Config.cs
@@ -239,19 +239,19 @@ namespace DenshaDeGoInput
 			}
 
 			// Set command boxes
-			buttonselectBox.SelectedIndex = DenshaDeGoInput.ButtonProperties[(int)InputTranslator.ControllerButton.Select].Command;
-			buttonstartBox.SelectedIndex = DenshaDeGoInput.ButtonProperties[(int)InputTranslator.ControllerButton.Start].Command;
-			buttonaBox.SelectedIndex = DenshaDeGoInput.ButtonProperties[(int)InputTranslator.ControllerButton.A].Command;
-			buttonbBox.SelectedIndex = DenshaDeGoInput.ButtonProperties[(int)InputTranslator.ControllerButton.B].Command;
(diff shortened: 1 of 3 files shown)


ginga81:
Thank you very much for quickly fix!
Japanese all reported users are checked and works correctly and  reported to me!

ginga81:
I also checked TYPE-II and Ryojou-hen Controller's automatic air brake new action!
This is my image!
Thank you very much for adding!

MarcRiera:
You're welcome. Nice to hear it meets your expectations!

PR #683, opened 2021-08-05
New: Add lighting change events

leezer3:
This PR adds lighting change events to the CSV routefile set:

### New Commands:
* **Structure.DynamicLight** - Loads a dynamic lighting set to be switched to.
* **Track.AmbientLight** - If dynamic lighting not in use, changes the ambient light color.
* **Track.DirectionalLight** - If dynamic lighting not in use, changes the directional light color.
* **Track.LightDirection** - If dynamic lighting not in use, changes the light direction.
* **Track.DynamicLight** - Changes to a dynamic lighting set loaded via _Structure.DynamicLight_

### TODO:
* ~~Test the thing, haven't done this yet....~~
* ~~Add an appropriate thingy for when events are shown in Route Viewer.~~
* Document.


leezer3:
Needed a minor tweak so the lighting model was actually re-calculated with single lights, but this does seem to work now....

Will sit on this for a few days before merging in case I think of something else that wants to be done in the meantime :)

PR #684, opened 2021-08-09
WIP: MSTS consist / vehicle parser

leezer3:
Heavy work in progress here, highly likely to explode unexpectedly.....

This will load (some) MSTS consists to a working visual train.
Not even thought about physics and stuff yet.
The new renderer is absolutely required for these to get a sensible framerate, and even this is still a little slow. We need to detect somehow if the proportion of texture in use for any given face uses alpha.

Unfortunately, the inbuilt consist / vehicle parser seems to have been very forgiving of errors in the textual block, which means that I'm still very much rehacking the base parser as I go along.....
Proabably want to cherry-pick the parser changes into the mainline branch.

msts01
30-odd van train from the 3DTS Cambrian route.

msts02

Class 47 & TEA tankers from Making Tracks' Network South East route.


leezer3:
Have picked out the fixes to the format parser, through piped brake type and the renderer tweaks and rebased this back onto the head of master.
Probably going to end up doing this a lot, as there are too many half-features we need to get the whole thing working.....

leezer3:
msts03

Took a little more bashing, but we can now load a MSTS consist into the main game.
Nothing really works (yet), so no physics, controls etc. but if nothing else we could just load a default set of acceleration curves. 

Large detailed consists are rather slower than I'd like though, although this is poly-count related as opposed to anything else. We still need to find some speedups in the GL3 renderer.

leezer3:
A little more work done.

Cabviews now load, but need a lot of components integrating. Camera restriction also appears to be broken.

If wheelspin is disabled, we can actually move our trains. No sound, brakes or power notch simulation at the minute mind, so not exactly useful. 
Need to also consider refactoring the acceleration curves a bit.
Also highlights the fact that wheels don't rotate :/ Need to investigate that properly- I believe that an animation frame for wheels pulls out the wheel radii from the engine properties, and presumably twiddles via speed and this. Would wheel radii be useful for animated files?!

leezer3:
Animation changes really want pulling to master when completed & working- This is a common method of doing things.

Haven't tested performance yet, perhaps this would be better in shader when completed?!

Unsure how complete the rest of this is at the minute....

leezer3:
API changes pulled out and merged into master, just to try and keep things in better sync.

leezer3:
This branch will now load the new hierarchy objects.
All complex parts appear to be where they should in a static object, which is encouraging.....

However, it doesn't get them into train exteriors, as the **CarSection** is tied to the current animated object. Still fiddling, heavy WIP.

leezer3:
Abolutely.
The whole point is to allow any hierarchy of parts to animate as a whole, and it's basically the standard animation format for any modern game.

If we add an appropriate controller (for valve gear, this would primarily be the wheel radius), then the animations actually work correctly in the current build.

(It'd be reasonably simple to add the wheel radii to train.xml so that a MSTS shape had working wheel animations, but the issue is that wheels can be attached to parent objects and stuff, which isn't yet really there)

----

Unfortunately though, we don't yet have many of the actual controllers implemented in-game.
It's somewhat of what we'd call a **chicken and egg** problem: To actually implement these controllers in any useful way, we need to implement the 'missing' subsystems. 


MSTS has a reasonably readable textual format, and I'm basically currently using it to start to implement these missing systems, as a known working prototype.
Nothing very useful is likely to come out of this in the short and probably medium terms, but it's all interelated with moving away from being a simple liner simulator to something 'better....

leezer3:
squashed

This is just the squashed faces for the image in the first post.

PR #686, opened 2021-08-11
[DenshaDeGoInput] Add support for Zuiki controller

MarcRiera:
Last week, a new controller was released for the Nintendo Switch: https://www.zuiki.co.jp/mascon/

This PR adds support for this controller. It's a HID-compliant USB controller, so nothing strange this time. Because this controller has more buttons than usual (it's basically a Switch Pro Controller with no right stick and a handle instead of a left stick), only the usual buttons are reported in the input plugin configuration window. However, as it's a standard controller, the remaining buttons can be assigned like any other joystick in the "Controls" section.


leezer3:
Thanks.
Do you know if this reports a sensible (detectable?) joystick name, or whether its just showing up as a generic switch controller?

If detectable, we might possibly be able to do something so these are a little more obvious in the main menu, something you obviously can't do with the PS stuff.

MarcRiera:
It has a standard VID/PID (0f0d:00c1) but the name reported is "One Handle MasCon for Nintendo Switch". So definitely detectable as long as joystick names are working.

Has it ever been considered to store joystick identification when assigning controls? As far as I know, currently we store the joystick index and button/axis number, regardless of the joystick. These indices change between controllers, so it's not very useful, specially considering most people will probably have only one or two joysticks they play with. Making these settings controller-specific would probably help when configuring additional buttons like in this case.

leezer3:
That's a good start then.
I've added a controller specific image in the main form controls as a starting point. I'm wondering about adding a clickable link underneath, which triggers the Densha config form (basically just one-word configure or something); Could also link this to the RailDriver config form, and simply set to empty in all other cases.....

With regards to joystick indicies-
I think the single joystick thing is probably why it's never been changed. Assuming that (most) people only have the single controller, this will always end up at Index 0.
It's probably usual to keep things connected at all times around reboots, which on Windows at least here keeps consistant numbering,

Only real thing with changing that I suppose would be that it'd likely be a breaking change in the config file format, unless we wanted to keep a dictionary of indices or something.
I suspect we've probaby got enough people using multiple versions that this might cause an issue.

MarcRiera:
I think adding a link to the input plugin is a good idea that could be extended to the rest of identifiable controllers (the Unbalance ones are also standard HID and have specific VID/PIDs.

The change from index to VID/PID/name was a suggestion of possible improvement but we can save it for the future after thinking well how to better implement it and see if anyone reports any issue because they are using multiple controllers.

MarcRiera:
I've made the changes required to enable the input plugin if it's disabled (plus changed the detection of the plugin to filename, it seems more simple).

Let me know if there's anything else I can help with to get this merged.

leezer3:
If testing off the filename, it's ideally going to need a case invariant check (just in case), which I've tweaked in.

Mergable I think, will just check to see the builds don't explode.....

MarcRiera:
Thanks!

PR #687, opened 2021-08-14
Fix: Crash when checking for update without an internet connection

Kenny-Hui:
Currently the program will crash with an Unhandled Windows Form Exception if a connection to the server cannot be established.
This PR should fixes it, though I still have a few questions:

image
Is that suppose to be a typo?
The original code comment has mentioned that "The internet connection is broken...", so I assume we can't establish a connection to the server.

Also on the [contributing guidelines](https://github.com/leezer3/OpenBVE/blob/aa9f170147fea0f707e21307f9d7469d036d1aff/Contributing.md), it is mentioned to use "Unix (CR LF) line terminators"  
but as far as I am aware Unix only uses LF, and Windows uses CR LF?

### DIFF
--- a/source/OpenBVE/UserInterface/formMain.cs
+++ b/source/OpenBVE/UserInterface/formMain.cs
@@ -1704,6 +1704,12 @@ namespace OpenBve {
 				}
 
 			}
+			catch (WebException)
+			{
+				//The internet connection is broken.....
+				MessageBox.Show(Translations.GetInterfaceString("panel_updates_invalid"));
+				return;
+			}
 			finally
 			{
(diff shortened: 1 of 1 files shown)


leezer3:
CR-LF Looks to be a typo, although that's a dangerous assumption......

GIT these days seems to automatically handle it (although it may well not have done 5 years ago, I honestly can't remember)
If you copy / paste the entire build dir from Windows to *Nix, you'll get some errors sorted by running Dos2Unix on the thing.

There might also have been a Mono build issue with one or the other, again can't remember :(

PR #689, opened 2021-08-21
Changed macOS openBVE icon to macOS Big Sur style

andylin2004:
This PR is an icon change for macOS systems only such that the icon has been changed to the macOS Big Sur style. This change was made so that the icon would not be seen as too big (relative to other icons in macOS Big Sur). This does not change the functionality of the app at all.


leezer3:
You've tarred the full path (including your home directory)
https://stackoverflow.com/questions/18681595/tar-a-directory-but-dont-store-full-absolute-paths-in-the-archive

Really ought to fix it so the TGZ isn't needed, that was a nasty hack in the first place as getting the permissions right was being a PITA.
Is this an edit to the icns file only; I seem to remember something about padding being added in Big Sur?

andylin2004:
> You've tarred the full path (including your home directory)
> https://stackoverflow.com/questions/18681595/tar-a-directory-but-dont-store-full-absolute-paths-in-the-archive
> 
> Really ought to fix it so the TGZ isn't needed, that was a nasty hack in the first place as getting the permissions right was being a PITA.

Maybe I should save this for future reference.

> Is this an edit to the icns file only; I seem to remember something about padding being added in Big Sur?
This is an edit to the icns file only, as [this](macosicons.com) site has a [template](https://macosicons.com/resources/macos-big-sur-app-icon) for making Big Sur icons.

PR #692, opened 2021-08-28
Fixing scaling issue and disabling full screening on macOS due to scaling issues

andylin2004:
Somehow OpenTK doesn't handle macOS screens well and when you change the resolution in the options pane, the resulting resolution is half of the expected resolution, so that got fixed.

Also somehow, unlike the Windows version of openBVE, the contents do not stretch in full screen, so I had to disable changing resolution in full screen (the numbers will change, but the resolution will not). This was reported to be an OpenTK issue, so I may investigate on which version of OpenTK we are using.

This was tested on a MacBook Air M1, so I'm not too sure how this will play out on non-retina screens (such as a M1 Mac mini or an older iMac).


leezer3:
See if this helps the fullscreen issues- I've added a specific resolution change command, as some backends don't seem to do this:
https://github.com/leezer3/OpenBVE/commit/75066894f733770da4296644d6e95aa0fc9deaf6

Retina might be tricky; Multiplying the resolution by 2 isn't any good for the X220 hackintosh I've got here (1366 x 768)
Detect the retina display by resolution maybe?

I've got a 2560x1440 primary monitor I could probably hook up to and see how it looks, but it's only on High Sierra.

andylin2004:
Which OpenTK version are we using?

leezer3:
Doesn't surprise me, but I'd really rather not go there.

4.0 is a (essentially) complete backend rewrite / switch from custom to GLFW.
It's also had numerous teething issues, which makes me very wary.

andylin2004:
The changes you made were on the right track, however the window ended up moving around (such that upper left corner of window was not at 0,0 and in some cases, moved off screen). Somehow, when I call on DisplayDevice to render things at certain resolutions such as 720x450, 825x525, 840x525, 1024x640, 1280x800, 1440x900 and 1680x1050, the display will run at 2x scaling. I'm still trying to figure this out.

leezer3:
That's really funky :/

Will take a dig into the OpenTK 4 source later and see if anything interesting is going on there with regards to Retina.

andylin2004:
I don't want to start switch statementing based on resolutions either, as I know there are older Macs that natively run at these resolutions (MacBook Air and MacBook Pro 15" 1440x900, MacBook Pro 13" 1280x800, and MacBook Pro 15" 1680x1050).

leezer3:
https://github.com/opentk/opentk/issues/47
The retina resolutions would appear to be returned in the list as half-values.

Presumably therefore we could itinerate in reverse through the list. We either find our resolution doubled (valid retina) or our original resolution, telling us the display is not retina.

leezer3:
https://forum.xojo.com/t/detecting-a-retina-display/29602
Cocoa code for detecting display scale factor.
Might take a look into patching OpenTK.

leezer3:
Untested, quickly hacked together branch for OpenTK to add a scale factor property to the display device:
https://github.com/leezer3/opentk/tree/DisplayScaling

Windows definitely works, will try and test the others next week sometime.

andylin2004:
> Untested, quickly hacked together branch for OpenTK to add a scale factor property to the display device:
> https://github.com/leezer3/opentk/tree/DisplayScaling
> 
> Windows definitely works, will try and test the others next week sometime.

Did you build a NuGet package for this? Would like to test it on macOS and for some reason it doesn’t really build on macOS (and I can’t really resort to windows, as boot camp isn’t an option)

leezer3:
You probably need to build using the build.sh first IIRC, something to do with generating the GL bindings,
First attempt was a little wonky though. Nearly there, but not quite.
[OpenTK.zip](https://github.com/leezer3/OpenBVE/files/7114902/OpenTK.zip)

The attached appears to work for me, although I haven't got a Retina display.

andylin2004:
> You probably need to build using the build.sh first IIRC, something to do with generating the GL bindings,
> First attempt was a little wonky though. Nearly there, but not quite.
> [OpenTK.zip](https://github.com/leezer3/OpenBVE/files/7114902/OpenTK.zip)
> 
> The attached appears to work for me, although I haven't got a Retina display.

Ah, realized that I could directly add a .dll to a project and bypass NuGet. I got the scaling to work in windowed mode (fullscreen is a bit tricky as the current scaling implementation relies on the static scaling value, which can overshoot at times), but I am considering a different implementation where the pixels of the content of a window scale up instead of the content handling the expanded space like there is more space in full screen (in windowed mode, this implementation would be reused but the window pixels would be doubled in area, with no UI real estate changes), as the UI still treats a retina screen like a regular display.

andylin2004:
I fixed the fullscreen window overshooting issue. It turns out that the macOS values related to calculating how much things scale by is dynamic (initially, the value was based off of the values that was detected on initial load).

leezer3:
Good stuff.

I'm going to leave merging this for the minute (not to mention the commit trail is getting a little messy!) until we see what OpenTK want to do. Still thinking about refactoring the Linux display driver a lot too, although I think your changes may well have made getting the DPI easier.

Worst case scenario, I'll just stuff up a fork on NuGet with the appropriate changes in there and we can start pulling that.

PR #696, opened 2021-10-03
Additions to the CSScript animation C# script interface

zbx1425:
## Changes
- An argument list can be passed to the constructor of the script class, in a format similar to URL query string. This enables reusing the same script file for multiple purposes, such as doors on both sides, which requires 4 individual script files before.
Example: 

```ini
[Object]
Position = 0, 0, 0
States = test.csv
TranslateZScript = test.cs?value=1
```

```csharp
using System;
using System.Collections.Generic;
using OpenBveApi.Hosts;
using OpenBveApi.Interface;
using OpenBveApi.Math;
using OpenBveApi.Trains;

public class OpenBVEScript {

	private int val;

	public OpenBVEScript(Dictionary<string, string> args) {
		val = int.Parse(args["value"]);
	}

	public double ExecuteScript(AbstractTrain Train, int CarIndex, Vector3 Position, double TrackPosition, 
		int SectionIndex, bool IsPartOfTrain, double TimeElapsed, int CurrentState) {
		return val;
	}
}
```
- CSScript and FunctionScript now all implement the AnimationScript interface, eliminating the need of special-case handling
- Usage of scripts expanded to Rotate and TextureShift

### DIFF
--- a/source/OpenBveApi/FunctionScripts/AnimationScripts.cs
+++ b/source/OpenBveApi/FunctionScripts/AnimationScripts.cs
@@ -1,4 +1,4 @@
-﻿using OpenBveApi.Math;
+using OpenBveApi.Math;
 
 namespace OpenBveApi.FunctionScripting
 {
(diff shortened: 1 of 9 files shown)


leezer3:
Thanks, now that's a good idea.

MSTS animations are on the list of things to implement, and adding this is a very good way of abstracting out the animation class.
Will have a proper look / test later.

PR #698, opened 2021-10-21
Change: Abstract the motor sounds, work on loading stuff from XML

leezer3:
### Changes:
* Motor sounds are now an abstract class, with an inherited update model.
* BVE5 motor sound tables may now be loaded via XML
* OpenBVE performance curves may now be loaded via XML
* TrainEditor2 motor sounds simulation model merged into the base TrainManager, and fixed. (Appears to have broken not long after implementation)
* Importing BVE2 motor sounds into TrainEditor2 now works.

### TODO (?):
* ~~Update CarXmlConvertor with the performance curves, easy.~~
* Is it worth copying the current motor sound table format into an appropriate file and adding a header, similar to how BVE5 has essentially done it?

### Rationale
Implementing MSTS and BVE5 motor sounds will require a different approach to the current motor sound tables.
This will allow this to be implemented easily, and hopefully for that matter added into the editor in future.


leezer3:
For good measure, I've added XML loading of standard BVE performance curves.
This would let us have differing performance properties per motor car in the train, and would also let us now load a complete train from XML. (n.b. this isn't reccomended at the minute, and the plugin / format will need some minor mods to accomodate this fully)

This is now reasonably mergable, and should probably be done so before starting on anything else. (Otherwise it'll just get out of sync with master and become a total PITA to get in)

Next major aim should be to pull out the BVE5 performance tables. There is an implementation for this in the following older PR: https://github.com/leezer3/OpenBVE/pull/154/

Issue #702, opened 2021-11-07
Speed limits are not properly enforced


MarcRiera:
### Description

Speed limits are apparently not being properly enforced on the user/AI driver. This applies to both track and signal limits.

1. If currently driving on a track section with a 30 km/h speed limit, for example, no warning message is shown when the speed limit is exceeded.

2. When a limit/signal with a lower speed requirement is passed above the required speed, a warning message is shown, as expected. However, if the driver reduces the speed (enough to make the message disappear), and then accelerates over the limit, no warning message is shown (as in case 1).

3. The built-in AI driver respects the limits on approach/pass, but ignores them once the train is inside the limited section. This also happens when driving over the limit manually and then enabling the AI driver.

### Reproduction

This issue happens generally, not a specific combination of route/train. It is very noticeable on FEVF with EM A1 (try driving with the AI from the beginning, it will happily drive over 15 km/h and 30 km/h on the first section, with no warning messages shown on screen). Link: https://fevf.altervista.org/pg008.html

leezer3:
My AI appears to be working OK at the minute. Still playing....
You may sometimes see a very small (~2km/h or so) overshoot / undershoot in hitting the speed limit, which is as designed; Michelle's original code tries to implement a small amount of quasi-human randomness into stuff such as precise AI target speeds.

The speed limit message is interesting. I've just tested on FGC-
We seem to miss the overspeed message for the initial limit (20km/h), but the second and subsequent limit messages definitely trigger. 

Something's definitely funny, but I don't think speed limits have been touched at all recently.

leezer3:
I think the commit above should solve the messages problem, by detecting if the train falls below the required speed and resetting the message trigger. 
First section issue was due to the train exceeding the speed on fast-forwards at game start, and the message being ignored as it was in minimal simulation mode.

I can't reproduce any AI issues though- This appears to be behaving exactly as expected.

MarcRiera:
> My AI appears to be working OK at the minute. Still playing.... You may sometimes see a very small (~2km/h or so) overshoot / undershoot in hitting the speed limit, which is as designed; Michelle's original code tries to implement a small amount of quasi-human randomness into stuff such as precise AI target speeds.
> 
> The speed limit message is interesting. I've just tested on FGC- We seem to miss the overspeed message for the initial limit (20km/h), but the second and subsequent limit messages definitely trigger.
> 
> Something's definitely funny, but I don't think speed limits have been touched at all recently.

[...]

PR #703, opened 2021-11-08
Change: Improve the behavior of the AI on a downwards gradient

leezer3:
https://github.com/leezer3/OpenBVE/issues/702

@MarcRiera 
This I think should solve the issue. Makes the AI a little bouncy on the brakes in this regard (might want a bigger interval, but that would make the speed graph bounce instead), see what you think


MarcRiera:
It seems to have improved, but on FEVF the bug persists. The AI driver seems to go a bit crazy on the brakes, moving the handle constantly between B4 and B7, but this isn't enough to keep the train from speeding up slowly until 45 km/h, where the AI driver finally goes up to B12. When the train is around 32 km/h, the AI driver goes back to the crazy braking and the whole situation repeats.

leezer3:
Try that and see what you think now.

Lot of the trouble is that the brakes on the most recent version of the EM A1 are terrible, and so things keep overshooting.

PR #714, opened 2021-12-11
Fix: UI Thread Exception not written to log

zbx1425:
This PR addresses these two issues:
1. `t + Environment.StackTrace` calls `ThreadExceptionEventArgs.ToString()`, which simply returns the string `ThreadExceptionEventArgs` instead of the full message and stack trace, and prevented the actual exception message from being written to log. By the way, according to player reports, "Unhandled Windows Forms Exception" is incredibly common. Their causes are partially unknown due to this issue.
2. In rare circumstances `null` can be present in `Interface.CurrentOptions.RecentlyUsedTrains`, and raises exception in formMain_Load resulting in OpenBVE being unable to start.

### DIFF
--- a/source/OpenBVE/System/Functions/CrashHandler.cs
+++ b/source/OpenBVE/System/Functions/CrashHandler.cs
@@ -68,8 +68,8 @@ namespace OpenBve
 	        }
             try
             {
-                MessageBox.Show("Unhandled Windows Forms Exception");
-                LogCrash(t + Environment.StackTrace);
+                MessageBox.Show("Unhandled Windows Forms Exception.");
+                LogCrash(t.Exception.ToString() + Environment.NewLine + Environment.StackTrace);
             }
             catch (Exception exc)
             {
--- a/source/OpenBVE/UserInterface/formMain.cs
+++ b/source/OpenBVE/UserInterface/formMain.cs
@@ -218,11 +218,11 @@ namespace OpenBve {
 			if (MechanikRouteIcon != null) listviewRouteRecently.SmallImageList.Images.Add("mechanik", MechanikRouteIcon);
 			for (int i = 0; i < Interface.CurrentOptions.RecentlyUsedRoutes.Length; i++)
 			{
-				if (Interface.CurrentOptions.RecentlyUsedRoutes[i] == null)
-				{
-					continue;
-				}
-				ListViewItem Item = listviewRouteRecently.Items.Add(System.IO.Path.GetFileName(Interface.CurrentOptions.RecentlyUsedRoutes[i]));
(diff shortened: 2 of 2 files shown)


leezer3 (commented):
Thanks.

I might actually think about re-writing this bit completely. 

Checking the extension is a crappy method now we can ask each plugin in turn whether they support a file- I'm rather tempted to get the plugin to pass back the icon in the CanLoad call. Issue with that is obviously that I should have thought of that when implementing the interface, and it would either be a new method or a breaking call change, both if which we really ought to avoid if possible :/

zbx1425:
Thanks!

------
Some random thoughts about error reporting:
1. The "report error" dialog should really be localized, and might be better if it can pop up automatically at a restart following a crash - especially currently there is still a lot crashes. In the local (Chinese) community developers had to instruct the players "Locate 'report bugs' at the left-bottom corner, click the second button from the top and screenshot the log" again and again and again.
2. Logs are not that detailed, specifically many exception handlers just log the message without the stack trace. This made it tedious to inspect a crash, especially when it's not that easy to reproduce. I understand that saving length is critical but I think stack trace is quite useful.
3. Compressing the log ZIP in LZMA is not quite optimal, since many compression softwares on Android lack support of LZMA, which made it difficult for developers to analyze players' logs when they're out.
4. Sending error reports should be easier. Since there's no telemetry, I think making the error report dialog pop up automatically, and provide useful information on how to ask for help, or specifically how to send the log ZIP to bveworldwide / a developer / some other collecting service (maybe set up one that doesn't require registration?) would help new players a lot.

-----
Some random thoughts of plugins:
1. Since there isn't a way to easily install third-party plugins, and there aren't any currently, I personally is a bit doubtful about its necessity. Surely it's nice to split the functions into modules, but the plugin API is thing to maintain a unnecessary backward compatibility of, and I think is sort of a drag in some occasions (like this). And if we're going to support variable length blocks like that in BVE5, that might require major work in the internal route structure, how are we going to keep the route loading plugin interface backward-compatible at that point?
2. Also, if the folder list is to be generated according to plugin function calls, I'm personally a bit worried about performance in folders with a large amount of items.
3. Also (*2), downgrading from a version of Mechanik support to a version without will crash, since the new Mechanik dll is not deleted during this process. Surely uninstalling and reinstalling will fix, but I think it can be done better.

leezer3:
Added a small please report to the crash handler dialog.

Not being localised is entirely deliberate unfortunately, as sometimes crashes can occur before (or during) the loading of language files.

n.b. I've seen essentially no reports of this error in the wild. It really suggests something odd with HK / Chinese stuff if it's being seen regularly (I'd guess that there's a character somewhere something doesn't like), and would be interested to see a crash log to see if we can get to the bottom of this.

Issue #395, opened 2019-09-23
[Suggestion] built-in Uchibo objects for complement use, want to freely use it for freeObj


ginga81:
When I am going to release the TrackFollowingObject's sample route, I choosed my Tohoku-Shinkansen's objects.
Certainly, I think that the most of users are can install and check it.
But I thought that OpenBVE has already have a lot of objects for complement use for Uchibo Objects.
Additionaly, the railway crossing sound has already builtin!

So, I want to suggest that the built-in Uchibo objects and sound can use for multiple numbers and different object category with freeobj, rails and sound.
When we want to use built-in Uchibo's object, the Rail and FreeObj's number and filepath is set to the same number and filepath as Uchibo.
This is naturaly result at the case for complement use.
But, when we are creating the new original route, FreeObj, Rail number, and filepath want to change freely.

In my suggestion, virtual Uchibo files can referlence is good solution.

complement use:
With Structure
    .WallL(3) uchibo\BridgeL.csv
    .WallR(3) uchibo\BridgeR.csv
    .Rail(22) uchibo\tr_bal300r.csv
    .FreeObj(46) Uchibo\lim45r.csv

My Idea:
With Structure
 .FreeObj(123) uchibo\BridgeL.csv
 .FreeObj(124) uchibo\BridgeR.csv
 .Rail(1) uchibo\tr_bal300r.csv
 .FreeObj(1) Uchibo\lim45r.csv

At my idea, if the same real file is put at Railway/Object, it is conflict.
So necessaly to check the exist real file.
But, if we can freely use built in Uchibo object and sound, at the case of TFO's sample route is easy to install (using package system).
Not only for complement use, if we can use for create the new route, these objects are more useful.
Please consider about this.

leezer3:
I'm sure something can be figured out, although they weren't initially intended for this purpose.

Perhaps something like the following:
```.FreeObj(0) $BuiltIn\Uchibo\BrigdeL.csv```

Where the *$BuiltIn* represents the built-in object path.
The potential issue I see here is that if the object referenced is missing, we'd have to decide whether checking the main object folder is a good idea or to simply error out.

Let me think about it and play with things a bit.

ginga81:
Yes!
I imagined such as this too!
I think that at the help document and tutorial, built-in object is useful for the sample.
This feature is now not hurrying, at the future of the good timing, please consider this.

In this case, all of $Builtin objects are built in from the another real folder.
[...]

PR #715, opened 2021-12-16
New: Add a package type for custom plugins

leezer3:
This PR adds a package type and associated plumbing to install and load custom user plugins via the Package Install menu.

Some (minor) flaws if custom plugins are installed, which probably should be addressed at some stage:
* No mediation between two plugins which can load the same type of file.
* No way to enable / disable specific plugins. (Possibly wants something for all plugins as per input device plugins??)
* No current way for a plugin to modify the icon for content it supports.

cc @zbx1425

### DIFF
--- a/assets/Languages/en-US.xlf
+++ b/assets/Languages/en-US.xlf
@@ -606,6 +606,9 @@
 					<trans-unit id="type_other">
 						<source>Other</source>
 					</trans-unit>
+					<trans-unit id="type_plugin">
+						<source>Plugin</source>
+					</trans-unit>
 					<trans-unit id="type_select">
 						<source>Please select the package type:</source>
 					</trans-unit>
--- a/source/OpenBVE/Game/Menu/Menu.cs
+++ b/source/OpenBVE/Game/Menu/Menu.cs
@@ -475,6 +475,10 @@ namespace OpenBve
 									installedFiles = string.Empty;
 									Manipulation.ExtractPackage(currentPackage, Program.FileSystem.OtherInstallationDirectory, Program.FileSystem.PackageDatabaseFolder, ref installedFiles);
 									break;
+								case PackageType.ContentPlugin:
+									installedFiles = string.Empty;
+									Manipulation.ExtractPackage(currentPackage, Program.FileSystem.PluginInstallationDirectory, Program.FileSystem.PackageDatabaseFolder, ref installedFiles);
+									break;
 							}
 						}
 						break;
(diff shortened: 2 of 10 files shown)


zbx1425:
Thanks for the support!

About enabling/disabling plugins, I think route package dependency should be taken into account. For example, a route package that requires a route-loading plugin to be installed and activated to run. If the plugin can be disabled or is disabled by default (like input device plugins), this may cause some confusion.

About the API breaking changes, maybe adding a new non-abstract function, or use something like LoadProperties from IRuntime?

By the way, about the current file system configuration of package installation location, I think allowing to set different paths for routes and trains is a weird choice, and probably making the registry global is weird too, since it makes it possible for a user to create a false "dependency met" situation where the dependencies are installed but are not at the same place. For example, if a user installed package A which is a dependency of package B, then decided to change the directory due to disk space etc., and then installed package B, the dependency will be met but route in package B won't be able to find objects in package A.

leezer3:
Don't necessarily disagree with any of those in principle, but they probably come under future plans & could require route modification :)
The whole package system was conceived as a very simplistic zip extractor really, which it does well enough.

I think the next thing to do us to allow package references for objects, e.g.
**{GUID}\Objects\tree.b3d**
e.g.
https://github.com/leezer3/OpenBVE/issues/395

zbx1425:
I am thinking of something different, probably `Option.IncludePackage({GUID})` that loads the paths all files in a package into a lookup table which is checked against when a file used in the route does not exist. Otherwise it would be a bit of an effort to migrate an existing route, especially considering you won't know the GUID of a dependency package before packing, and there will be more manual checking if there are multiple dependencies. Probably `Option.IncludePackage({GUID}, {Path Prefix})` to avoid conflict.

For example, if A is train package contains train `AAA`, while B uses train AAA but is installed at a different directory, `Train.Folder(AAA)` will fail, but if `Option.IncludePackage(A), Train.Folder(AAA)` is used then the path `AAA` will be automatically checked and converted.

Also, referencing with GUID seems to open the possibility of other methods of managing packages, such as extracting them each in individual directories to avoid path conflicts (in a folder other than LegacyContent? Since it's a different system in this way). It is incompatible with the current structure but I think it's a nice thing.

zbx1425:
I thought it over again, and I personally thought my idea is to
1. Don't provide seperate settings for paths for Train, Railway and Other packages, just provide one setting for a base path, and install them at BasePath/Railway and BasePath/Train.
2. Ensure all packages are installed in the same directory, and move all files over when changing the paths.
to avoid confusion to end users, who one cannot expect to have a full understanding on the implementations.

Although I'm definitely not qualified to complain about this and that, honestly speaking, I think allowing installing them in different directories in the first place (and the decision to provide seperate settings for paths) is quite weird from the first place. Real-world users might rarely find installing them in different directories useful, and these false "dependency met" situations are likely to create more problems.
Disk space might be a reason that one wants to install routes in different partitions or disks, but that would require a more elegant solution rather than the currently implementation. For example, two packages on seperate disks sharing a same dependency package cannot be handled this way.

Backward compatibility is a valid concern. But keeping this as-is also has major concerns (these false "dependency met" situation effectively makes the dependency system useless; and changing installation directories can be a major concern).

PR #728, opened 2022-01-11
New: Write our own bitmap decoder

leezer3:
This PR implements a written from scratch bitmap decoder.

Currently a work / testing in progress, but this is intended to reduce our dependancy on external libraries a little more, and eliminates a bunch of rather nasty array conversion from the original code.
It also (should!) returns correctly return limited 256, 16 etc. color pallettes , rather than the hacky solution implemented at present. 

Biggest probable downfall for this would be whether it's actually necessary. Bitmaps are a deceptively simple format, and already well implemented in Mono etc.

Reference images:
https://entropymine.com/jason/bmpsuite/bmpsuite/html/bmpsuite.html


leezer3:
This now (largely) works, but for images using the default color table it's messed up :/

Still only really experimental, but may be useful in the future.....

leezer3:
I can't immediately find anything broken with this.

Interestingly, the reduced color table algorithm also works better with the new decoder. (e.g. file attached below) 
No immediate reason why this should be the case, so Windows must be interfering somewhere.
[abusimbel.zip](https://github.com/leezer3/OpenBVE/files/8659945/abusimbel.zip)

Marginally considering trying to implement this (or re-port libpng) for good measure, as that would then allow us to dump System.Drawing completely:
https://github.com/JoshDullen/libpng.net

PR #755, opened 2022-03-16
New: Implement sprung handle types

leezer3:
A relatively common feature of prototype power / brake handles is a spring feature.
This gradually returns the handle to a 'default' position after a set period of time. 

This PR implements this, along with some missing XML plumbing for setting the number of power / brake notches.

### Significant Changes:
- Allows the driver car to be set via the train.xml
- Allows the number of brake notches to be set via XML using the _Train --> Car --> Brake --> Handle --> Notches_ **NOTE:** The number of brake notches is only set via the driver car (values set on other cars will be ignored)
- A brake handle spring delay in seconds may be set in XML via _Train --> Car --> Brake --> Handle --> SpringTime_ - Each time the timer elapses, the brake will be increased by one notch. Interacting with the handle resets the timer.
- The maximum notch a sprung brake handle returns to may be set in XML via _Train --> Car --> Brake --> Handle --> MaxSprungNotch_
- Spring type may be set via XML _Train --> Car --> Brake --> Handle --> SpringType_ with available values as follows:
   1. _Unsprung_ - No spring return
   2. _Single_ - Spring return timer is only reset when this handle is operated
   3. _AnyHandle_ - Spring return timer is reset when any power / brake / reverser handle is operated
   4. _AnyKey_ - Spring return timer is reset when any key is pressed. (think about whether to limit this to security keys etc, or whether it should be global??)

### TODO:
- Sprung reverser handles???

### Future Design Issues to Consider:
Really, we ought to have a per-car handle. If in the future, decoupling or changing the physical driving cab is implemented, this would allow the 'new' driving cab to have correct handle types etc.
This could *possibly* be done relatively easily via changing the current handle location to be a reference to the actual handle within the current driver car.
However, this could equally create a total and utter mess with existing content, and anything not specifically designed for this.

_Possibly_ could do something like aliasing **DriverCar** to the inital driver car, and using **CurrentDriverCar** as the reccomended variable going forward, but that's just got a different set of messes to contend with.

Still a WIP, and not added to the documentation yet.

Loosely related to https://github.com/leezer3/OpenBVE/pull/684 as MSTS allows this handle type.

### DIFF
--- a/source/CarXMLConvertor/Convert.ExtensionsCfg.cs
+++ b/source/CarXMLConvertor/Convert.ExtensionsCfg.cs
@@ -449,6 +449,8 @@ namespace CarXmlConvertor
 						newLines.Add("<ReadhesionDevice>NotFitted</ReadhesionDevice>");
 						break;
 				}
+				newLines.Add("<Power>");
+				newLines.Add("<Notches>" + ConvertTrainDat.PowerNotches + "</Notches>");
 				newLines.Add("<AccelerationCurves>");
 				foreach (ConvertTrainDat.AccelerationCurve curve in ConvertTrainDat.AccelerationCurves)
 				{
@@ -461,6 +463,7 @@ namespace CarXmlConvertor
 					newLines.Add("</OpenBVE>");
 				}
 				newLines.Add("</AccelerationCurves>");
+				newLines.Add("</Power>");
 			}
 			else
 			{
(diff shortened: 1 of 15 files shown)


leezer3:
Largely complete other than implementing the AnyKey reset method.

**NOTE:**
This makes a small change to the XML format-
The _Acceleration_ node containing the acceleration curves is now moved to be a child node of a new _Power_ node, so that we can control the power handle notches / spring identically to the brake handle.
The original position for the acceleration node is still supported, but once this is merged, the preferred location should be updated in the documentation.

I don't believe many (if any) released trains use this format yet, and it's not complete, so possibly consider dropping this, but it made it into a release build :/

leezer3:
Documented, and probably reasonably ready to merge....

Sit back and think for a little bit first though in case something has been broken I've not noticed.

PR #794, opened 2022-06-08
Refactor handling of AnimatedObjects

leezer3:
This PR refactors AnimatedObjects so that they are a subset of the overarching **UnifiedObject**

Required by https://github.com/leezer3/OpenBVE/pull/684 in order to get the **HierarchyAnimatedObjects** working correctly, as these require a totally different update method.

### TODO:
* Need to slim down and combine the methods.
* Provide a virtual __Update()__ method within the **UnifiedObject**
* Can we get rid of some more of the excess parameters in some of these methods?
* Possibly convert a bunch of the AnimatedObject booleans to a flags enum??

### DIFF
--- a/source/LibRender2/Trains/CarSection.cs
+++ b/source/LibRender2/Trains/CarSection.cs
@@ -39,7 +39,7 @@ namespace LibRender2.Trains
 				Groups[0].Elements = new AnimatedObject[a.Objects.Length];
 				for (int h = 0; h < a.Objects.Length; h++)
 				{
-					Groups[0].Elements[h] = a.Objects[h].Clone();
+					Groups[0].Elements[h] = a.Objects[h].Clone() as AnimatedObject;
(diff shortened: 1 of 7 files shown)


leezer3:
Hmm...

Structural issues are a major problem here :/
Far too much depending on being able to access deep within the CarSection / Elements.

This might want a rethink / major rework of the entire CarSection approach...

leezer3:
https://github.com/leezer3/OpenBVE/tree/CarSectionRefactor

Marginally tested rework of the CarSection, to be used as an alternative to this. Appears to work 'better' but I'm not at all convinced with the entire ObjectState concept :/
Quite a bit of reference chucking about too, which I'm not thrilled by....

I *think* this one should actually allow the HierarchyAnimatedObjects to be used in a CarSection, but may well break in new and interesting ways.....

PR #799, opened 2022-06-18
Allow to load dependencies of train plugins

HengYeDev:
I am currently writing a plugin which uses a few DLL dependencies. The LoadFile method does not allow to load these dependencies, but the LoadFrom method does.


leezer3:
Are you sure this is needed?

This should already work (indeed BVEC_ATS already loads a couple of external dependancies), and the basic difference between LoadFile and LoadFrom is what happens in terms of *further* dependencies down the chain not found in the GAC.

HengYeDev:
I believe so, I had a dependency that depended on other dlls and it wouldn't work with the current openbve

HengYeDev:
by the way, I am using linux (mono) if that makes a difference...

HengYeDev:
Give me a few days to push my plugin up, but I'm using EmbedIO https://unosquare.github.io/embedio/

leezer3:
Interesting :)
The plugin interface isn't really designed to run a webserver (it's single threaded / blocking, so I'm presuming you're spawning a background thread?), and we've definitely seen issues with this sort of thing before. IIRC in that case it was someone launching an embedded browser for an in-game login, but essentially anything that wants to control its own threadloop may be problematic.

HengYeDev:
I think embedio uses its own thread, it works great for me to display a speedometer and some variables on a browser window and I haven't found any performance issues so far

Issue #789, opened 2022-05-06
[Suggestion] my collection of suggestions


adfriz:
This suggestion not an urgent things but its better if this implemented.

1. Fog density
  Can change the overall density of an fog
2. Font color using hex code
  Can use hex color code for determining the text color
3. Change ingame font
   Can change the ingame font to font that already installed on PC
4. Resize track marker image
   Can resize the image of track marker (usually placed at top right on screen) so the size not depends to only image resolution.
5. Ambient occulusion
   Dark shading that make object have more depth and realistic looks
6. Graphic preset
   Preset like low, medium, high and custom so new player can change it easly

adfriz:
Another suggestion but focused on camera stuff

7.Preserve Camera Postion in Exterior Mode(F2)
  Usually when in exterior mode by default your camera will pivot/anchor at first car (car0), when you change the pivot to other car and to other camera mode such as Free mode(F3) and back to exterior mode you automaticly back to first car.

8. Adjust speed when looking around the camera with mouse
   Not many user know you can control the camera direction with toggle-ing right click on your mouse and move it but the speed is too fast too sensitive, even you only move your mouse very slowly. (I dont know is this just me, because i always use a mouse with 1600dpi)

9. Can change the near clip for the camera
   This feature is usefull if you have an addon that place the camera on a seat like a passenger, because if you can change the near clip the camera wont cutting through object if you close to it. (This feature maybe can cause a bug when Z sorting)

Issue #772, opened 2022-04-03
Compliations of issues/questions from my experience


Kenny-Hui:
## Possible Bugs (Minor)
~**1.** This panel does not have a theme set when initially opening the menu. (Both in Windows & Linux)
This could be fixed by switching to any other tab, then returning to the game selection menu.~
![image](https://user-images.githubusercontent.com/28094366/161421884-0970abc7-3ca4-4837-9835-5380da3087ef.png)

**2.** On Linux, in-game UI fonts often seems to clip through. (nvidia-driver-470)
Though I assume it just has something to do with the Graphics driver?
![image](https://user-images.githubusercontent.com/28094366/161422142-4cc7d99e-79b8-4f55-aea5-777275d0c6d8.png)

## Possible Impactful Bugs (Resolved)
~**1.** After the door width fix of #770, I can't get the train to power anymore. (Whether using a custom plugin or the default ATS)
Reproducible with the same train in #770.~

~**2.** After a car rolls over/flipped, the loop.wav will no longer play on that car.
However after Jumping Stations, the loop.wav remains silence.~

~**3.** The driver position seems to be moved forward after using CarXmlConverter
**Train.dat:**
0, 2.385, -1.825 (0, 2385, -1825)
**Train.xml:**
0, 2.385, 8.425~

~**4.** #767~

## Questions (Resolved)
~**1.** The train.xml allows you to set a custom Notch Description.
The train has 8 Power Notch and 7 Brake Notch, and I added the following to the xml file:~
```
<NotchDescriptions>
	<Power>10deg,15deg,20deg,25deg,30deg,35deg,40deg,45deg</Power>
	<Brake>10deg,15deg,20deg,25deg,30deg,35deg,40deg</Brake>
</NotchDescriptions>
```
~In-Game I am only seeing the Neutral Notch/EMG displaying `10deg,15deg,20deg,25deg,30deg,35deg,40deg,45deg`.
![image](https://user-images.githubusercontent.com/28094366/161423934-5981ea42-9411-4252-b366-3ac10a699fa8.png)
Am I doing something wrong?~

~**2.** I have not tried this for a while now, but a few months ago when I set the Horn Sound to a loopable one (one start sound, one loop sound), it seems there will always be a pause after playing the start sound, before start looping.
Is there a way to make it play smoothly?~

~**3.** When I reload a extension.cfg along with a train.xml in Object Viewer, the following error will pop up:
[...]

Issue #712, opened 2021-12-03
OpenTK.DisplayDevice throws exception in 1.8.2.0+


danfe:
I've tried to update FreeBSD port to version 1.8.2.0 and then 1.8.2.1 but they both won't start now, showing the window that says "Unhandled exception: The type initializer for 'OpenTK.DisplayDevice' threw an exception".  This is probably related to cloning of OpenTK in order to stop depending on the outdated OpenTK package.

Anything I can provide or help to track down and fix this problem?

leezer3:
Can't think how :/
The OpenTK fork doesn't change anything fundamental, and only adds one XServer call, which has been tested on Linux, and ought to be supported on anything remotely more than a potato....

Have you got a full stacktrace for the exception? (Might be in the log)

Will try and spin up a VM later in the week and see if anything happens for me, although BSD isn't something I've tried before.

danfe:
> Have you got a full stacktrace for the exception? (Might be in the log)

No, it's pretty much nothing in the logs:
```
15:19:57  WARNING: Program crashing. Creating CrashLog file: /home/danfe/.config/openBVE/Settings/OpenBVE Crash- 2021.12.03[15.19].log
15:19:58  WARNING: Program crashing. Creating CrashLog file: /home/danfe/.config/openBVE/Settings/OpenBVE Crash- 2021.12.03[15.19].log
  --8<--  --8<--  --8<--
03.12.2021 15:19:58
OpenBVE 1.7.766.438-danfe Crash Log
Program is running on the Unknown backend
Current screen resolution is: Windowed 960px x 600px 
```
In some older logs, it was more rich:
```
15:11:51  Using openGL 1.2 (old) renderer
15:11:51  Initialising game window of size 960 x 600
15:11:53  Game window initialised successfully.
15:11:53  Initializing the default shaders failed- Falling back to legacy openGL.
15:11:53  Initializing the touch shader failed- Falling back to legacy openGL.
15:11:53  Renderer initialised successfully.
```
In the log from 2020.10.27, it also managed to deduce the backend (SDL2):
```
27.10.2020 12:17:19
OpenBVE 1.7.404.146-danfe Crash Log
Program is running on the SDL2 backend
Current screen resolution is: Windowed 960px x 600px 
```

leezer3:
So, I've just installed a fresh FreeBSD virtual machine, using the XFCE desktop environment.

I've then installed the current OpenBVE port. (1.8.1.0)
[...]

PR #805, opened 2022-08-04
New: Allow user selectable font

leezer3:
This PR allows a user selectable font for the main menu and the in-game UI.
Only tested on Windows at the minute, and may end up being restricted to this- Most likely to be an issue is going to be the font selector combox.

Depending on the exact font, rendering glitches can also be seen at the top / bottom of the text in-game occasionally, even in Windows (this is a known issue with the default font on some Linux installs)
Probably when the font glyph is colored in the top pixel row, and it might be worth investigating the glyph set for a glitching font.

font

#789


ginga81:
I would be very grateful if we could be change not only the font type but also the size.
left unders informations are a bit small for me.

leezer3:
Try the HUD size slider, set to large.

This really ought to be renamed, as it actually controls the font size for most things in-game, as well as the OSD.
Will need to take a look into the code to see how well resizing the doors indicator etc. would work dynamically, as that might be a problem.

Could just rename this to font size, and add a few more graduations to it I suppose.....

ginga81:
Oh!
Thank you very much!
I'm sorry to bother your time.

leezer3:
Done a lot of messing with the font drawing routines today.

This PR now fixes the font bleeding under Mono (e.g. https://github.com/leezer3/OpenBVE/issues/772 )
We still have trouble with characters with massive tails on some fonts, but it's much better.

### TODO:
* Font needs to be regenerated in librender2 when changed from the main menu (takes effect after game restart otherwise)
* Mac font is currently hardcoded (due to Unicode glyph set)- Needs to be changed.
* Why do most fonts under Mono / Linux show up as un-renderable? Can we hide these [bunch of squares looks horrid]?

leezer3:
Bit more work to block the non-renderable fonts from being displayed in our font picker- Turns out these are mostly the uncommon Noto variants which use the Western character set to provide more codepoints.....

Need to block anything non-usable on Apple, which I haven't yet tested.

Could possibly actually merge the two (soon to be three) lists.

@danfe
Merging this PR will add the Noto CJK fonts as a soft dependancy, which I'll need to check with regards to FreeBSD to see about the correct font naming. 
Most GUI Linux desktop systems will have these installed already (Chrome / KDE commonly pull them in), but I don't know exactly how common they're likely to be on FreeBSD.
At ~30mb, I'd be inclined to add them as a requirement in your port, but YMMV :)

danfe:
> Merging this PR will add the Noto CJK fonts as a soft dependency, which I'll need to check with regards to FreeBSD to see about the correct font naming. [...] I don't know exactly how common they're likely to be on FreeBSD.

We do have Noto fonts in the ports collection (so the port can depend on them), but CJK are further split into _-hk_ (Traditional Chinese Hong Kong), _-jp_ (Japanese), _-kr_ (Korean), _-sc_ (Simplified Chinese), _-tc_ (Traditional Chinese) packages, probably owing to their size (these fonts are quite beefy).

Overall this looks like a useful feature.  I'll be interested to see if it can remedy the blurriness I've mentioned in the end of issue #712.

leezer3:
Beefy is somewhat of an understatement looking at the port :/
https://www.freshports.org/x11-fonts/noto
~1gb download.

Debian on the other hand is 75mb download, and appears to be in just about any desktop package.
https://packages.debian.org/stretch/fonts-noto-cjk

Much less happy about adding a 1gb dependancy than a 75mb one which is in most places by default, I think you'll have to be the best judge here as to whether it's worth doing this.
FWIW, looking at the latest Cloudflare stats, I'd guess probably 60% + of our userbase is from Korea / Japan / HK, which makes getting proper unicode showing for most of these in the main menu by default is actually a massive deal :)

---

Bluryness:
I can't see much difference TBQH. Cleartype under WinForms is a mess under Windows, let alone anywhere else (heavily dependant on DPI, screen resolution and all sorts of other mess)

Might be better, but it's somewhat hard to tell anything useful after completely changing the font- I haven't patched anything else in this regard.

---

Other Linux / BSD related patches in there whilst I was at it:
* FreeBSD now shows the nag prompt if running as **root** / **sudo** (absolutely no need for this under any circumstances)
* Changed the bug report window to use an internal popup to show logs under Linux / BSD as opposed to a shell launch. (otherwise does nothing unless we've got a default .txt handler, as opposed to under Windows showing the choose app prompt)
* Using the close button on the main form was unreliable, and often crashed under BSD. This appears to be a Mono bug in that the filesystem watcher doesn't like being disposed as things are closing.

danfe:
> Beefy is somewhat of an understatement looking at the port :/ https://www.freshports.org/x11-fonts/noto ~1gb download.

This is a meta-port which includes _all_ fonts by depending on the corresponding basic, emoji, extra, and CJK ports.  It's intended for user convenience (manual installation) and not to be used as a dependency.

> Debian on the other hand is 75mb download, and appears to be in just about any desktop package. https://packages.debian.org/stretch/fonts-noto-cjk

Right.  As I've said, FreeBSD splits it even further (one CJK into sC, tC, H, J, and K).

> Much less happy about adding a 1gb dependency than a 75mb one which is in most places by default, I think you'll have to be the best judge here as to whether it's worth doing this.

Of course no need to depend on that huge meta-package, just on `x11-fonts/noto-jp` and/or other `x11-fonts/noto-??` should be sufficient.

leezer3:
The three fonts we're specifically setting based upon locale at the minute are as follows:
```
Noto Sans CJK JP
Noto Sans CJK KR
Noto Display
```

I believe this would be:
```
x11-fonts/noto-basic
x11-fonts/noto-jp
x11-fonts/noto-kr
```

Noto Display has the Cyrillic character set, the JP version has JP & some CN / common (I believe enough to be OK for our usage), and KR does Hangul.
@Kenny-Hui or another native Chinese speaker using Linux is going to be better placed to comment as to whether it wants the specific SC variant though. As someone with a basic / translator grasp of Chinese, all I can really do is check for missing glyphs and that it is the same as Windows.

danfe:
> Noto Display
> I believe this would be `x11-fonts/noto-basic`

Hmm, per what I see, `noto-basic` includes just the standard Sans and Serif families, Display comes from the `noto-extra` package (please confirm or deny).

Kenny-Hui:
> @Kenny-Hui or another native Chinese speaker using Linux is going to be better placed to comment as to whether it wants the specific SC variant though. As someone with a basic / translator grasp of Chinese, all I can really do is check for missing glyphs and that it is the same as Windows.

From my limited testing, they all seems to render fine without the SC variants. (Well, at least before my installation got screwed up)
Though most of the content I have is all Traditional Chinese, not too sure about simplified

leezer3:
> > Noto Display
> > I believe this would be `x11-fonts/noto-basic`
> 
> Hmm, per what I see, `noto-basic` includes just the standard Sans and Serif families, Display comes from the `noto-extra` package (please confirm or deny).

You're right :/
Comes from the core package in Debian, but plain Noto Sans seems to be OK, so switching to that instead.


> > @Kenny-Hui or another native Chinese speaker using Linux is going to be better placed to comment as to whether it wants the specific SC variant though. As someone with a basic / translator grasp of Chinese, all I can really do is check for missing glyphs and that it is the same as Windows.
> 
> From my limited testing, they all seems to render fine without the SC variants. (Well, at least before my installation got screwed up) Though most of the content I have is all Traditional Chinese, not too sure about simplified

Thanks :)

leezer3:
Now merged.

I'm going to leave this a week or so in the hope that any new glitches in the daily builds get picked up by someone (really.....) but will put out a release after this, as IMHO it's a pretty significant step for new users.

----

### KNOWN ISSUES:
* Some fonts with massive tails (couple of handwriting style ones on my Windows box) may lose the end of the tail in-game- **WONTFIX**
* With some fonts on non-windows platforms, some textboxes in the main form may slightly overlap the borders. I think this is likely a Mono issue- Could possibly increase the spacing in these places slightly to compensate, but not sure it's the best idea, so leaving this alone for the minue.
* There's no single font covering all Unicode codepoints under Linux, and as a result, we need to detect the system locale and pick the most logical one by default. Should be OK, but might need a few more adding- see how we go for the minute.

----

### Other thoughts:
We could possibly use the locale detection code to try and set the native language by default, now we have a good font.
Not sure whether this would be a good idea, and some of our non-english users will need to comment on this- I'm concerned about mis-detections, and what happens in a missing font scenario.
At least, english is relatively widely known?

PR #810, opened 2022-08-30
New: Implement QuadTree visibility

leezer3:
This PR implements the QuadTree visibility algorithm from the abandoned OpenBVE2 demo ( https://github.com/sladen/openbve2 )

### Why is this important?
Using a QuadTree allows us to create intersecting routes. 
Consider the following diagram:
intersecting

This doesn't work on prior versions of OpenBVE, as the visible world objects are calculated based upon the linear track position, and so at the intersection point, Station A would not be visible.

Additionally, using a QuadTree will allow us to much more easily import other formats, as these don't depend on linear visibility.

### CHANGES:
* **Options.ObjectVisibility** may now be set to mode 2- Quad Tree based visibility.

### TODO:
* This is currently only used for static faces.
* The current arrays should be optimised / made simpler by using lists instead.
* ~~Use of the QuadTree visibility algorithm should be controlled (in CSV format routes) by the **Options.ObjectVisibility** toggle.~~
* Test extensively!
* ~~What performance impact does this actually have?~~
* ~~How should the quad distance be controlled? Currently set at a fixed 500m, but this should likely be set to the viewing distance, or perhaps something like half of the viewing distance???~~

### BROKEN:
* Animated objects, I *think* not in the initial quad visibility. (strangely not trains)


leezer3:
Interestingly, at this point the background thread seems to be slower...
Creating the list copies is likely to be costly, but needs to be done to ensure thread safety :/
Perhaps we could do this only when the object lists change via the setter?

There are also some glitches in GL3 mode, which appear to occur when some objects are first shown. Something funky up with the VAO timing, but still investigating this.
Whilst I suppose the background thread could be dropped for the moment, I'd like to get this running correctly in order to be able to offload more stuff onto it in the background,
For example. I believe the heavy texture loading can likely be done in said thread, and pre-load the quad ahead.

leezer3:
Glitches found-
I believe a clone was being kept internally by the list, so our solution is to generate the VBO on list add. Not perfect, but there we go.....

Speed:
Using quad tree mode seems to be anywhere upto about 10% slower, although it handles complex areas better. Not really investigated this at present, we may have some speedups to be found.

PR #814, opened 2022-09-08
New: Steam traction modelling

leezer3:
This is the start of implementing a new steam traction model.
Some basic ideas pulled from the BVEC_ATS / OS_ATS implementation, but this is intended to run solely within the main game.

Not yet implemented into the main simulation, doesn't build properly, untested at the minute....

### Aimed for features:
* ~~Exhaust and live steam injectors.~~
* ~~Cutoff, including ineffective range.~~
* ~~Fire size / area.~~
* ~~Blowoff.~~
* ~~Basic valve gear and stroke.~~
* ~~Blowers.~~

### TODO:
* Acceleration curves need hooking in somehow.
* More research into the numbers and assumptions.
* Actually test this.
* Animated variables to hook into all of this.
* ~~Percentage based power / reverser handles.~~

### **Boiler** - Contains boiler related properties
* **WaterLevel** - Sets the initial water level in the boiler
* **MaxWaterLevel** - Sets the max water level in the boiler
* **SteamPressure** - Sets the starting steam pressure
* **MaxSteamPressure** - Sets the max steam pressure
* **BlowoffPressure** - Sets the pressure at which the blowoff triggers
* **MinWorkingPressure** - Sets the minimum working pressure
* **SteamGenerationRate** - The number of units of water converted into steam per second
* **LiveSteamInjectionRate** - The number of units of water injected per second by the live steam injector
* **ExhaustSteamInjectionRate** - The number of units of water injected per second by the exhaust steam injector
### **Firebox** - Contains firebox related properties
* **Area** - The initial area of the fire
* **MaxArea** - The max area of the fire
* **MaxTemp** - The maximum temperature of the fire
* **ConversionRate** - The number of fuel units burnt per second, and converted into temperature units
* **ShovelSize** - The number of fuel units added per shovelful
### **CylinderChest** - Contains cylinder chest related properties
* **StandingPressureLoss** - The standing pressure loss in units per second
* **BasePressureUse** - The base pressure used per wheel revolution (e.g. a single complete stroke)
* **SniftingValve** - _None_ , _Manual_ , _Automatic_ and _AutomaticRegulatorTriggered_
* **CylinderCockLeakRate** - The number of pressure units leaked per second when open with full regulator
### **ValveGear** - Contains valve gear related properties
* **WheelCircumference** The circumference of the driving wheel of the valve gear (used to define the base wheel position for cranks / pivots)
#### **Crank** - A crank position for use in animation, with the following properties:
* **Radius** The radius from wheel center of the crank
* **Length** The length of the attached crank rod
* **Offset** The number of degrees offset from zero at the start of the simulation
#### Pivot - A pivot position for use in animation, with the following properties:
* **Radius** The radius from wheel center of the pivot
* **Offset** The number of degrees offset from zero at the start of the simulation
### **Tender** - Contains tender related properties
* **CarIndex** - Sets the index of the tender car if a separate car (n.b. this car may carry no other cargo)
* **FuelLoad** - The starting fuel load
* **FuelCapacity** - The starting fuel capacity
* **WaterLoad** - The starting water load
* **WaterCapacity** - The starting water capacity


ginga81:
I want use Japanese Steam locomotive's Bypass-cock.
At the BVEC_ATS, our presentation wasn't still well.
This time, I want to requre.

ginga81:
One more after.
Using something that calculates the rotation of the cylinder rod will not cause wheel slipping.
I would like it to move at high speed in conjunction with the speedometer that the driving wheels are spinning and the rods are also spinning.
I want to show the wheel slipping for My D51, so only used png.
StateFunction can change by slipping speedmeter's speed.

ginga81:
In this book, written the Bypass Unit's explanation.
Please consider and I wish to the implementation this.
bypass

蒸気機関車メカニズム図鑑
細川武志 著
グランプリ出版
ISBN978-4-87687-317-3
Translation quoted from 142P
Translation by F81_tec200.

Steam Locomotive mechanism pictorial book
grand prix Publishing

ginga81:
Additionally, Steam Chamber Air Valve unit also wish to implementation.
Please consider also this and I wish to the implementation this.
In Japanese steam locomotive, this unit's operating steam pressure is about 5 kg/cm2.
Under the 5 kg/cm2, open the valve.
When the open and close, the another each sound is generated.
When this valve is opening and steam pressure is high, the noisy sound generated one time, and steam pressure to 0 to suddenly.
But when the steam pressure is down in advance under the 5kg/ch2, the sound is not generated.
When the valve is opening and add the steam pressure from zero, the valve is close and sound generate one time.
But, not need the 5kg/cm2.

I don't know about the another countly's locomotive(I think that this is using only about Japan only.)
So I think perhaps that it will set the default value is 5 kg/cm2.
But I think that this value will need to valiable.

steamcamber

蒸気機関車メカニズム図鑑
細川武志 著
グランプリ出版
ISBN978-4-87687-317-3
Translation quoted from 134P
Translation by F81_tec200.

Steam Locomotive mechanism pictorial book
grand prix Publishing

leezer3:
Will see what I can do, but bear in mind this is still going to be relatively simplistic simulation.

https://advanced-steam.org/5at/technical-terms/steam-loco-definitions/bypass-valves/

FWIW, bypass / snifting valves (UK term) were never especially common here, now that i know exactly what you're talking about from the diagram. Normal practice here would be to coast with a tiny bit of regulator, which also helps with overcoming the rolling resistance of the train & keeps things lubricated. (everything is lubricated via atomised oil in the steam)

They tended to be used more widely in countries who used steam traction much later, with the associated superheated, higher boiler pressures etc.

The second (air chamber valve) would appear to be a more automatic variant on the same basic idea. Probably what we'd call a drifting valve.

Not quite clear why both were required on the same loco, but shall see where things go.

ginga81:
I already know that overseas, as you said, it is closer to mid and the regulation valve is slightly opened to make it run close to this.
I have drive some England's steam locomotives with Steam's Train Simulator 20xx.
(I like BR/LNER Class J50 that is feel very real.)
But, these are very important elements in more real simulating Japanese SL.
And I think it can be said that these are evolutions unique to Japan.
At my D51, I wanted to include these,  but at 2014, my English  vocablary wasn't good, so I couldn't say more for BVEC_ATS.
So, I was include to D51 was dummy by-pass cock by reverser lever.
In this time, his english and by this book, we can say these acculately.

At Japanese railway museum at Saitama Prefecture, change from order to lottery. 
So, now we can't drive D51 simulator is incredibly impossible.
So, I want to more real simulate.

JunmoreHeavyBox:
Here are a few of my tips...
It seems to me that steam locomotives have only wheel linkages indexed 0 and 1 for wheel linkage objects with angles differing by 90 degrees, there is no such thing as a left or right linkage. Thinking about three-cylinder steam locomotives.
Consider an articulated steam locomotive such as the Big Boy, which requires two sets of wheel rod sets.

Idea Demonstration
```
<WheelLinkages>
    <WheelLinkage><!--WLIndex 0-->
        <Radius>0.75</Radius>
        <Height>0.8</Height><!--Piston centre from ground-->
        <Stroke>0.8</Stroke>
        <LinkageAngle>120</LinkageAngle><!--LkIndex 0-->
        <LinkageAngle>240</LinkageAngle><!--LkIndex 1-->
        <LinkageAngle>0</LinkageAngle><!--LkIndex 2-->
    </WheelLinkage>
    <WheelLinkage><!--WLIndex 1-->
        <!--Data...-->
        <LinkageAngle/ >
        <!--LkIndex 3, 4, 5-->
    </WheelLinkage>
</WheelLinkages>
```

Perhaps some complex animation of the linkage mechanism...
The height difference between the piston and the wheel radius can change the acceleration characteristics and the timing of the sound trigger without the use of animations.

Ideally everything in the wheelbar set is synchronised, while multiple wheelbar sets like two bogies with Speedometer variables may have angular deviations after a certain number of miles.

According to the presentation, piston indexes 0,1,2 are at wheel rod group index 0, piston indexes 3,4,5 are at wheel rod group index 1 and so on if you customise them.
These variables can be used whenever Animated objects are used by the same vehicle.
_WheelRotatei_ The wheel rotation animation, where i represents the index of the wheel rod group to which the wheel belongs.
_WheelLinkageZi_ The animation of the wheel linkage moving along the Z axis.
_WheelLinkageYi_ Animation of the wheel linkage moving along the Y-axis.
_PistonLinkageRotatei_ Animation of the piston linkage rotating along the X-axis.
_PistonLinkageZi_ Animation of the piston linkage moving along the Z-axis.
Where i represents the Linkage index.

If possible, I would like to refine the customisation to a custom radius for each wheel, a custom stroke for each linkage and a custom piston height. However, I have abbreviated the presentation.

JunmoreHeavyBox:
Mention of the custom wheel radius of each wheel brings to mind a geared steam locomotive with piston that rotate faster than the walking wheels. Regardless of the reduction ratio, it is synchronised in the wheel linkage set, as it is the gear-drive rod that transmits the power. Similarly, this type of drive can be connected to the wheels of a coal and water trailer to turn it into a power wagon, but also this type of drive necessitates that the wheels of the coal and water power wagon remain synchronised with the animation of the locomotive wheels in a wheel linkage set.

leezer3:
That's a very good point, although I'll note this is still a heavy WIP and experimental playground at the minute :)
It's basically ported from BVEC_ATS, although I've done considerable work (experimentation with numbers!) on stuff like the fire / boiler simulation properties. 

As far as I'm aware I'm the only person to actually use the valve gear animation (the D51 uses a more 'conventional' frame based approach), and adding too many available variables may confuse matters, especially as this requires manual model parts and positioning.
I've converted the rod positions to an array though, and can probably do the same with the pivot locations. This might cut down on the number of required animation instructions too

JunmoreHeavyBox:
The variables I'm considering are just custom things linked via indexes, similar to the indexes in the Extensions.cfg file... I got confused looking at it, and even using Excel to build it.
I am considering the same steam locomotive with all the animation objects used in the XML carriage file, which has nothing to do with the locomotive's carriage position in the train. But this means that wheel linkage across vehicles in my idea is a pain.

JunmoreHeavyBox:
There are several components that can be considered from this PR.
The use of a wheel linkage to connect the wheels and motor of a non-cylinder locomotive, that is just an animation.
A fireless locomotive differs from a steam locomotive in that it is supplied with a limited amount of compressed steam from a large cylinder, no fire and no water. If without the cylinder, it looks like a battery.

ginga81:
The reason I chose frame-based on the D51 is because it didn't animate in 3D when the driving wheel slipped.
The wheels didn't move at all even though the speedometer was up when the driving wheel slipping.
Therefore, I gave up on the BVEC_ATS driving wheel calculation and decided to create 32 transparent PNGs and animate them.
However, with this, if you run it for a long time, about 30 minutes or more, the position of the driving wheel will shift, so if the driving wheel can be linked when the wheel slips animation even at 3D, I would like to make it in 3D.

JunmoreHeavyBox:
I have a similar, but less obvious, question about the Speedometer variable for animated objects.
No. 7 in #821

JunmoreHeavyBox:
Speaking of wheels synchronised across vehicles, how about adding [carIndex] after the variables for wheels and linkages?

The concept of a wheel and rod set is a bit confusing, I would like to approximate the composition of multiple steam locomotives in the traction model, where the final traction is a composite of the traction of multiple virtual steam locomotives. When any one set of wheels slips it does not interfere with the other wheels that are pulling properly. This is not only true for the animation.

Then just to talk about the "compound expansion steam locomotive", without seeking to implement it. If you let the train plugin control any virtual steam locomotive alone.
https://wikimili.com/en/Compound_locomotive

Issue #815, opened 2022-09-26
[Request] Make each Animated object in the nest have independent variables


JunmoreHeavyBox:
The variables of the Animated object are unified within a section (like a car, or one of the car's two bogies, or a coupler. recently added), This makes the movement of the multiple wheels in this part using the Animated object synchronised, but it doesn't have the linkage like the linkage that drives the locomotive.
Wish there was a switch to have an animation object have separate variables in the nest, which would allow for gaps in the animation of multiple objects in the same section. (The animation gap for each object is not noticeable, I really... request a feature for a little thing.)

Independent Variable Switch.animated
```
[Object]
IndividualSubject = True
```

This can be a lot of thinking about things like the location of objects...

leezer3:
I think you're mis-understanding the way the current speedometer works, or your translation is missing something somewhere :)

We don't really have a concept of induvidual axles / speeds at the minute (another structural issue, but not one that's likely to be easily solved)
All cars have two 'axles', which define the two track based pivot points of the car.
The speedometer is calculated from the front axle of the car it is attached to.

A link-driven locomotive would by defnition have all wheels syncronised, as otherwise it'd tear itself apart.
If you're looking to have slightly non-synchronous wheels on a non-link driven locomotive consider adding / subtracting a small random value based upon speed to give your desired effect. I'd guess something along these lines as a starting point:
```
RotateXFunction = value + [delta * speedometer / RADIUS_OF_THE_WHEEL] + if[speed > 0 ,random[0, 0.0017 * [speed / 10]]  , 0]
```
(Untested, but would give something like 1/10th of a degree or so randomly added per frame at a reference speed of 36km/h)

JunmoreHeavyBox:
I understand that a car has two bogies that each have their own speedometer variables, what I'm thinking is that assuming the bogie is also a car (bogies stacked with vehicles, like an 81XX train), the steering of this trolley The frame part uses a model of wheels. Such a bogie (actually a car) has its own speedometer variable for both wheels.

leezer3:
As far as I can see what you're actually asking for is to get the speed of the wheel rotation, which only has a tangential relationship to the speedometer.

The current setup simply doesn't allow for this to be calculated, hence the standard formula uses the speedometer, and assumes that the wheel moves in tandem with the ground position.
Adhesion / wheelslip are handled internally by simply dropping the applied acceleration output, as opposed to true simulation of the wheel position.

[...]

PR #819, opened 2022-10-09
New: Implement proper sanders

leezer3:
This PR implements 'proper' sanders as an alternative to the BVE readhesion device (this is intended to simulate a control system for an electric motor)

XML only feature.
At the moment, sanders are assumed to double the base adhesion rate. Probably OK, as this might get a little too complex otherwise....

### Usage:
A Car node may contain a **\<Sanders>** node, with the following child nodes:
* Type- controls the type of sanders fitted, from the following _PressAndHold, Toggle, NumberOfShots, Automatic, NotFitted_
* Rate- The application rate of sand in units /s
* SandLevel- The initial level of sand.
* NumberOfShots- The number of shots available (if type is set to _NumberOfShots_)
* ApplicationTime- The time after which a sand application cuts off automatically.
* ActivationTime- The amount of time for which wheelslip must occur before the sanders are activated in _Automatic_ mode

A Sound node may contain a **\<Sanders>** node with the following child nodes:
* Activate- Played once when the sanders are activated with sand available.
* EmptyActivate- Played once when the sanders are activated with no sand available.
* Deactivate- Played once when the sanders are deactivated.
* Loop- Played looped when the sanders are active.
* Empty- Played once when the sand supply empties.

New animated variables:
* Sanders- Returns 1 if the sanders are active for a car, 0 otherwise
* SandLevel- Returns the sand level for a car
* SandShots- Returns the number of sand shots available

### TODO:
* A little more plumbing in the control activation / deactivation.
* Hook up panel variables.
* ???

Follow-up to https://github.com/leezer3/OpenBVE/issues/815#issuecomment-1272571285
@JunmoreHeavyBox

### DIFF
--- a/assets/Controls/Default.controls
+++ b/assets/Controls/Default.controls
@@ -1,7 +1,7 @@
 ﻿; Current control configuration
 ; =============================
 ; This file was automatically generated. Please modify only if you know what you're doing.
-; Default keyconfig version 3- 08.06.2017
+; Default keyconfig version 4- 10.10.2022
 
 POWER_INCREASE, keyboard, Z, 0
 POWER_DECREASE, keyboard, A, 0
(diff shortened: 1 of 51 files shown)


JunmoreHeavyBox:
How do I replenish my sand if I run out of it?
The XML station has not changed for some time and I would like the train to replenish its sandbox at the XML station.

Issue #822, opened 2022-10-15
Sounds (which are written into .animated file) cannot do playbacking if this .animated file is in another .animated file's [Include] section


BPI-919:
### Description

Hi, I tried grouping the [Sound] sections of .animated file(s), for purpose to avoid "code" duplications and also to make the positioning of the sounds within a specific 3D-object me easier. However, I ran into a problem during this process: When it's time to playback, the sounds are not hearable.
Please let you read the information below for the details (which are in the Logs section).

Thanks,
Andrew

## Route

This is a WIP route (Suburban Railway Lines H8 and H9 in the agglomeration of Budapest - HU), it is not downloadable yet. However, I hope that the screenshots will help you to identify my problem.

## Train

It is not a train-specific issue. It occurs with any train.

## Logs

I attached 4 screenshots below:
- In the first one (This is the route file with .csv extension), in line 27, you can see how the second file (which is in the second screenshot) is be included into this one.
- In the second one (H8_Structure.include), in line 359, you can see how the third file (which is in the third screenshot) was declared to this one. Of course, it has been used in the With Track section too, it is believable without screenshot.
- In the third one (This is the main .animated file, which collects the related .animated-s - I wish it could be worked properly...), in lines 2 and 3, you can see how the fourth file (which is in the fourth screenshot) was included to this one.
- In the fourth one (This contains an [Object] section with an "empty" object file and also contains 3 [Sound] sections. The 3 [Sound] sections include the sounds, with the appropriate path. I actually put an [Object] section at the beginning of the .animated file, in matter from this step, I hoped the sounds would be played in the game properly, but unfortunately I didn't hear them...), you can see the "invisible" .b3d file and the 3 problematic [Sound] sections.
Interestingly, if I move any sound (section) into the parent-animated (in which you can find this one by an [Include] section, in this case the parent-animated is: "peron_orskoz.animated"), OpenBVE plays this sound, properly. In the same "session", unfortunately, the other sounds (which stayed at the same location) could not be heard when it was the time to play the sounds.

### Related information 

My PC:
- Name: ASUS K401LB (notebook / Laptop computer)
- OS: Windows 10 Home 64 bit 21H2
- Memory: DDR3L 8 GB (4 GB + 4 GB)
- 1000 GB SSD (C:) + 24 GB SSD (D:)
- Intel Core i5-5200U CPU @ 2.20 GHz
- Intel HD Graphics 5500 & NVIDIA GeForce 940M

OpenBVE Version:
[...]

PR #827, opened 2022-10-31
Fix: Sounds are now playable in animated collections

BPI-919:
Please merge it. This is the solution of #822 .


leezer3:
Thanks :)
Really good to see someone else contributing code!

I'll look tomorrow properly, but I'm relatively certain the sound will need cloning, but that's really minor.

BPI-919:
Thank you. :) In the meantime, I encountered a new problem:
By the two-handled vehicles, the robot pilot (AI) does not work.
I hope, there will be a solution for this as well.

Issue #821, opened 2022-10-14
[Question And Bug?]Collect some train problems


JunmoreHeavyBox:
### Description

Please describe the issue you are experiencing.

1. ~~When the train jumps to any station, the original Noise of the train is not turned off. The Noise played the second time after the reset is superimposed on the sound of the game started before the reset.
Pay attention to the noise volume.
To eliminate the original noise, the train has to be derailed, but only to eliminate the noise of the last time. This issue can be a headache for trains that often use change (jump) stations, and the noise gets louder as the trains are reset more often in the game.~~

https://user-images.githubusercontent.com/114210303/195951793-7469b35c-6c27-4b9f-9da2-2b3de89fe025.mp4

2. ~~The classic problem. In Control Trailer mode (where the driver is not in the power car), it seems that the AI driver is unaware of the wheel skid of the locomotive located at the tail of the train and will only increase the gear of the traction handle to the end. It's not until the AI driver is in the power car that it's aware of the need to consider wheel slippage. I had to use the ReAdhesionDevice to limit traction. Although the ride is uncomfortable, it is a way for the driver to casually top up the traction and still get the train moving.
This question is not necessary for Multiple-Unit trains and "trains with good adhesion coefficients and loads".
But what good is it going to do? Customize RunInterval trains for the #329 ?
I've wanted to bring this up for a long time.~~

3. ~~Some errors have been found in the use of NotchDescriptions.
Master control handle brake position with HoldBrake: There are two unused notches texts between NotchHB and Notch1.
Locomotive brake handle: The text that should be Notch0 uses the text of Notch1.~~
```
<NotchDescriptions>
<Brake /> (EBn-1;HB; ; ;B1;B2;..;Bn-2) Master control handle brake position with HoldBrake
<Brake /> (EBn-1;HB;B1;B2;..;Bn-2) Natural
<LocoBrake /> (B1;B1;B2;..;Bn-1) Locomotive brake handle
<LocoBrake /> (R;B1;B2;..;Bn-1) Natural
</NotchDescriptions>
```

4. The classic problem. It's possible to push the AI train with your own car when bored.
During the process of pushing, the bogie will drift slowly towards the load.

https://user-images.githubusercontent.com/114210303/196785691-0d769ab0-4409-4aa7-8ca9-1adac921d98d.mp4

5. I tried to test the windshield function during 2D cab practice, and something went wrong.
[...]

PR #831, opened 2022-11-05
Fix: Notch related issue

Kenny-Hui:
Some works on #821
This PR *should hopefully* fix the duplicated overlay on the power notch.
(Also fix a crash when the custom notch description is not available for the current notch)


JunmoreHeavyBox:
Thank you very much! (^u^)

leezer3:
Thanks!

PR #833, opened 2022-11-16
Change: Improve Car XML Converter

Kenny-Hui:
This PR fixes some of the bugs and improve the Car XML Converter.  
Nothing major, just found several annoyances when needing to use the program.
## Fixes
- Special characters from **train.txt** not escaped in **train.xml**, leading to in-game crash due to an invalid train xml.
- Encoding is now detected, non-UTF8 encoded **train.txt** should be converted properly.
- `terminatedEarly` not reset between runs, resulting in conversion not working if the first attempt failed.

## Improvements
- Minor changes on the UI, should look slightly nicer now
image
- Support drag and drop files/folders to the textbox. (Will grab parent directory if it's a file, the folder browser dialog is quite tedious to navigate)
- The log box is now read only rather than disabled, should result in easier readability. (Also allows scrolling)

### DIFF
--- a/source/CarXMLConvertor/Convert.ExtensionsCfg.cs
+++ b/source/CarXMLConvertor/Convert.ExtensionsCfg.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Globalization;
 using System.IO;
+using System.Security;
 using System.Threading;
 using System.Windows.Forms;
 using Path = OpenBveApi.Path;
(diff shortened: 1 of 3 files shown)


leezer3:
Thanks.

I've added a version string to converted XML files now too and a different prompt if an older convertor version was detected. 
That ought to help if anything else gets found / fixed.

This won't be perfect if you mix nightly and release builds (release builds will always be assumed to be newer), but may help with other stuff in the future.

Issue #722, opened 2021-12-23
Route map image is't display at the tab change


ginga81:
### Description

When we choose a route, and change 'Map' tab, the route map image is not show.
To show, we have to change the window size.
I think that at the press of 'Map' tab, the route image should be show directly, and not need cnange window size. 
I think that this issue has been from at least 1.6.0.

At the 'Map' tab select:
![Screenshot from 2021-12-23 16-23-02](https://user-images.githubusercontent.com/22734699/147207989-8f7c419f-9558-4bdf-82e7-c652c67eb796.png)

Window size changed:
![Screenshot from 2021-12-23 16-23-14](https://user-images.githubusercontent.com/22734699/147208002-98fa5063-d3a9-4a0f-a9bd-7b46a2855156.png)

### Reproduction

This issue happens at OpenBVE's main program's. so no need any data.

## Route

This issue happens at OpenBVE's main program's. so no need any data.

## Train

This issue happens at OpenBVE's main program's. so no need any data.

## Logs

This issue is not crash so no output any logs.

### Related information 

My PC's spec:
OS:Ubuntu 20.04 LTS 64bit
CPU:AMD® Ryzen 7 2700x eight-core processor × 16
Memory:31.3GB
Graphic:AMD Radeon RX 590 Series (POLARIS10, DRM 3.40.0, 5.11.0-40-generic, LLVM 12.0.0)

leezer3:
Try the build from today.
I can't reproduce this at the minute, but I've added a force redraw on the routemap when the tab changes. This will hopefully give it the poke it needs.

ginga81:
I tested at 20220103's build.
But still not shown.
I think that the trigger is not enable at tab change.
If I run on Wine, this process is works perfectly.
![Screenshot from 2022-01-10 03-01-57](https://user-images.githubusercontent.com/22734699/148694568-dddd08c1-482c-4566-ac63-3b2ba1b2ed55.png)

Kenny-Hui:
Reproducible on the following VM:

**Lubuntu 21.10 (Mono 6.12.0.1xx):**
Something (Context menu, or a window) must be on top of the image, then that portion of the image will be shown.
When you switch to another tab or resize, it will remain empty again.

**Pop OS 20.10 (Mono 6.12.0.182, Latest):**
Image will always be blank no matter what.

Probably something to do with https://github.com/mono/mono/issues/12425
[...]

PR #849, opened 2023-01-04
UI improvements

Kenny-Hui:
This PR improves the Winform main menu (Mostly quirks running under mono), slightly improve the OpenGL Menu and other front facing stuff.
Not sure about merging it just yet.

## Main Menu
- Typo fix: RailDriver conrollers -> RailDriver controllers (How would transifex handle this?)
- The auto binding keyboard/joystick assignment no longer has a red background when capturing keys.
(Red color would usually imply there's an error of some sort?)
image
- "Report Problem" and the "About" Dialog now properly supports resizing.
- "Report Problem" Button in the Report Problem Dialog has been renamed to "Save Bug Report" to better reflect the actual behavior, along with a follow up dialog telling the user to submit the bug report.
image
- Some of the label in options were previously not translatable, this PR should fix it.
- UI element should display better under dark theme in Mono. (Not necessarily prettier however)
image

### Mono!
- The "About" Dialog will now block the main UI thread just like any other dialog that pops up.
(Both for consistency reason, and that mono just doesn't play well with these sort of stuff)
- Textbox is very slow on Mono, resulting in long freeze of the UI. However setting the text after the Form is Shown can result in faster loading (5s -> ~1.5s in here).
A side effect of this is that a form that is not fully initialized will be shown to the user first, this *shouldn't* matter as it only persists for a short period of time.
This change is applied to the About Dialog and the View Log dialog (Which is used in non-Windows OS)

- Listview header won't be rendered in Mono (at least here) unless a Auto Resize Column is triggered when the element is visible. (It probably is rendered, just the width not being set correctly?)

## GL Menu
- `CanLoadTrain` in Train.OpenBve will now return false if it failed to combine path, preventing a crash on the GL menu when browsing to a folder without read permission. (lost+found)
- `maxFittingLines` in Textbox *should* now be correctly calculated, previously parts of the text would not be displayed despite having more space left.
- "options" is now capitalized in the menu, displaying "Options"
- (Comfort feature) The TextBox scrollbar now reflects the portion that is displayed. (No longer just a small square)
image

## Other
- Some message box now shows an icon, which should give the user a better first glance at what is the message about.

## Question
- Now that the project is called OpenBVE (With the capital O), should it be corrected in the Language file as well?
Catalan, Magyar, Chinese (Hong Kong, my fault) uses OpenBVE, while others remain using the lowercase openBVE.
- How does the transifex translation works? I've tried searching "Route default", but there are no such entry in transifex.
Despite that it does exist in the `.xlf` file, does the transifex translation have to be added manually by the project creator for it to appear?
- Is the future of the sim's UI heading towards being OpenGL based (Even if it will take a very long time to replace, or even never), or is it going to be something like gtksharp?

Feedback appreciated (Other tester as well), I've only done a limited amount of testing on Windows.
[bin_debug.zip](https://github.com/leezer3/OpenBVE/files/10349676/bin_debug.zip)


leezer3:
Thanks :)

I'm going to try and look at this properly over the next few days.

In an **ideal** world, the current main menu would be dropped entirely in favour of OpenGL- WinForms is unfortunately a near dead windows technology. There's no support on x64 Apple, and bugfixes on Linux etc. appear to have largely died out as well.

However, it's also clear that the OpenGL main menu is nowhere near the standard of the current one. That's unfortunately IMHO not going to change fundamentally any time soon. 

---

Naming:
Originally, the small o was used by Michelle etc. 
For Japanese / eastern cultural reasons ( @ginga81 ?? ) I'm afraid I never really understood too well, we changed to using a capitalised O, but it's still inconsistant, and in most Western cultures, the two are essentially ideomatically identical. 
Unfortunately, users tend to use the two interchangably too, which basically makes it confused :)

FWIW, if we were using .Net naming conventions, the small o would be correct :P

Kenny-Hui:
Added a little tips at the bottom left in the dependencies tab, and package creation should now be more readable on dark theme.
image

The route preview fix is a dirty hack that just resize the Picturebox by 1px so it will be shown (It won't on base resolution)
image

Not entirely convinced by it (Also a bit blurry), but I think that's as far as we can go into making this thing display without user interference?

leezer3:
Now that's interesting with regards to the picturebox, but I suspect it's just a fluke.

I spent quite a while fiddling with that, and came to the conclusion it's timing related somewhere deep in Mono. The fix that worked for me on one system (above your additions) failed on that of @ginga81 and a separate machine of mine.

ginga81:
If today's daily build fixes #722, I want to test it.
is it applied?

leezer3:
Not yet- You would need to build from this pull request.

Kenny-Hui:
> Original issue with pictureboxes, definitely timing related somewhere as per my notes in the linked Mono issue: #722

I am starting to doubt whether that's the same issue, different system reacts very differently (unfortunately)

On my Mint 21.1 install (Cinnamon) and on an KDE Neon (Plasma 5.26) VM, the route preview will not be shown, but if you resize it then it will be shown. (Both which doesn't line up with the reproduction I've tested in #722, but does line up with @ginga81's described behavior)

This probably won't magically fix this across all system, but should help on certain system.
Although I probably have to do lots of further testing to see how this helps.

Prebuilt binaries: [bin_debug.zip](https://github.com/leezer3/OpenBVE/files/10351290/bin_debug.zip)

ginga81:
I ran this compiled build.
It displayed fine without resizing the window!
it's great!
I think my only remaining problem is that the train icon doesn't appear.
The upper route and folder icons are already displayed.
Screenshot from 2023-01-06 00-19-15

Kenny-Hui:
> I ran this compiled build. It displayed fine without resizing the window! it's great! I think my only remaining problem is that the train icon doesn't appear. The upper route and folder icons are already displayed.

No idea for train icon, I can't reproduce anything here :(

### Font handling
The font "Aharoni" doesn't seems to come with a Regular variants, this message would pop up everytime OpenBVE launches. (Which is annoying, and it would crash the entire menu trying to render the font in the ComboBox because it can't find the "Regular" variants)

image
How should we handle this?

leezer3:
Added that to the blocked list in my other PR.

Might need to block some more of the odder FOD fonts?
https://learn.microsoft.com/en-us/typography/fonts/windows_10_font_list

Kenny-Hui:
"Cambria Math" should be blocked as well
image

Some "junk" font on my Linux install:
MathJax_Script
MathJax_Size1
MathJax_Size2
MathJax_Size3
MathJax_Size4
MathJax_WinIE6
MathJax_WinChrome
MathJax_Vector
MathJax_Vector-Bold
D050000L

leezer3:
Thanks.
I've added a global block for any font containing the string MathJax and a couple more from the same URW base fontset under Linux.

Issue #853, opened 2023-01-12
[Request] Implementing some new .animated functions


BPI-919:
That came to mind that the .b3d files there are these commands:

- TranslateAll
- RotateAll
- ScaleAll
- ColorAll
- EmissiveColorAll
etc.

Of the above commands only the first two have a function that can be mapped at the .animated file level:
- TranslateAll -> TranslateXFunction, TranslateYFunction, TranslateZFunction
- RotateAll -> RotateXFunction, RotateYFunction, RotateZFunction

My idea would be to give other .b3d file level commands a function that can be matched at the .animated file level, similar to what was written above, so:

- ScaleAll -> **ScaleXFunction**, **ScaleYFunction**, **ScaleZFunction**
- ColorAll -> **ColorRFunction** / **ColorRedFunction**, **ColorGFunction** / **ColorGreenFunction**, **ColorBFunction** / **ColorBlueFunction**
- EmissiveColorAll -> **EmissiveColorRFunction** / **EmissiveColorRedFunction**, **EmissiveColorGFunction** / **EmissiveColorGreenFunction**, **EmissiveColorBFunction** / **EmissiveColorBlueFunction**

Personally, I would welcome these innovations to avoid duplication of file content, because if other .b3d files cannot be referenced in .b3d files, hopefully there is another way to avoid this problem, such as introducing new .animated file features.

Suppose we have two .b3d files that differ only in that at one end there is a command like this: "_RotateAll 0, 1, 0, 180_".
Both .b3d files are linked to one .animated file, so:

"
[Object]
States = something_1.b3d

[Object]
States = something_2.b3d
"

Instead, we could solve it by having only the file something_1.b3d, and we would duplicate it two times, after reversing the duplication:

"
[Object]
States = something_1.b3d

[Object]
States = something_1.b3d
RotateYFunction = 3.1416
"

This is also okay, but for example for the EmissiveColorAll command, I can't solve this with the current methods like this:

"
[Object]
States = something_1.b3d

[Object]
States = something_1.b3d
EmissiveColorRFunction = 255
EmissiveColorGFunction = 255
EmissiveColorBFunction = 255
"

Unfortunately, here you need two identical .b3d files with an "_EmissiveColorAll 255, 255, 255_" command at the second end.

If this feature extension were feasible, there would be no need to worry about file duplications.
In case of duplication, it is necessary to modify in both files, if we modify or specify something, or we also reserve unnecessary storage space.

leezer3:
[...]

PR #855, opened 2023-01-12
New: Add scaling functions

leezer3:
This PR adds scaling functions for animated objects. (Much of the plumbing for this was implemented with the GL3 renderer)
https://github.com/leezer3/OpenBVE/issues/853

**ScaleXFunction**
Controls the scale of the object in the X axis.

**ScaleYFunction**
Controls the scale of the object in the Y axis.

**ScaleZFunction**
Controls the scale of the object in the Z axis.

---
### NOTES:
* Changing the direction for these is *not* supported.
* The order of application is **SCALE** ==> **ROTATE** ==> **TRANSLATE** as per standard OpenGL matrix math.


adfriz:
hey leezer3

I tried the daily builds 2023-02-02 on windows 10 this function seems only accept a value (int and float) and when i try writing `ScaleXFunction = value + 0.003` the mesh get squash to center along x axis instead.

I expected to scale the mesh outward by 0.003 every frame on both positive and negative x axis.

leezer3:
That function is going to be a total mess.

Assume ~120 frames per second plus in the viewer with a simple object.
The initial scale value will be 0.
Might take anywhere up to 10 seconds before it irredeemably glitches, and I don't know off the top of my head whether an overflow will wrap back around to zero.

adfriz:
Understandable, im not going to use that function either 😅, thats just an example.

What i already got in my head when see that scale function is "dynamic" train connector, you know that rubber thing.

The mesh get scaled slightly when train accelerate or decelerate, or better the mesh can get tapered when the train on curve track, if curving to the right any vertex that in positive X get scaled inward, and the vertex in negative X get scaled outward. Scaled not based on their origin though (not scaled towards 0,0,0) but scaled along z axis.

PR #868, opened 2023-03-08
Change: Improve DenshaDeGo! Input usb mapping

leezer3:
https://bveworldwide.forumotion.com/t2345-densha-de-go-controller-plugin-not-registering-controller#21523
@MarcRiera - This should solve the reported issue, but your thoughts on this design may be interesting.

Basically switches to using a constructed class and a nice human readable enum to switch controller type.
Also adds some log messages if we encounter an unknown PID from the recognised VID list.

Not sure on the names for the unbalance enum members, you'll have to comment on those :)

### DIFF
--- a/source/InputDevicePlugins/DenshaDeGoInput/Config.cs
+++ b/source/InputDevicePlugins/DenshaDeGoInput/Config.cs
@@ -27,8 +27,6 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Windows.Forms;
 using System.Drawing;
-using System.Timers;
-using OpenTK.Input;
 using OpenBveApi.Interface;
 using Timer = System.Timers.Timer;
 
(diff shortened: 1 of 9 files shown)


leezer3:
Some further thoughts:
* I suspect we could probably just recognise *all* Zuki Inc. PIDs as controllers and see what happens.
* Taito Corp PID 0001 and 0002 are probably Densha controllers we haven't seen yet.
* Hori Inc has a whole bunch of USB fight sticks under thier VID, and so I don't think assuming that anything unknown is a Densha controller is a good idea.

leezer3:
I was looking at it before you posted on the forum that you would :P

Your thoughts seem logical, but I've got none of these things (only a RailDriver), so happy enough to be guided by whatever you think is better.

PR #877, opened 2023-03-20
New: Add option to allow routes to be driven in the reverse direction

leezer3:
Still somewhat of a WIP, needs various camera related glitches sorting.

### Significant Changes:
* Add **Options.ReverseDirection** - If set in a routefile, it may be driven in the reverse direction.
* In the **AbstractTrain** the FrontCarTrackPosition and RearCarTrackPosition have changed signature from a method to a virtual double. This is technically breaking. but I don't think anything will be using these, and it's much more sensible.

### TODO:
* ~~Direction for F / R needs changing in the main program.~~
* ~~lnitial camera position is in car 0, not last car (Oddly, is OK going in / out of train)~~
* ~~Cabview is not being reversed when calling Reverse() function on train.~~
* Needs tag in train.xml to allow reversal of cab views so dual ends work properly.
* ~~Arrival / departure messages glitched- Probably needs the events reversing?~~
* If train overhangs signal at initial station, this generates a passed red signal message. Signalling is probably somewhat broken at the minute.
* ~~AI overshoots stations. Probably distance to the next station related (??)~~
* ?? Flip signalling objects ?? - Unconvinced, is this a developer only option?

### DIFF
--- a/assets/Shaders/default.frag
+++ b/assets/Shaders/default.frag
@@ -19,7 +19,7 @@ out vec4 fragColor;
 
 void main(void)
 {
-	vec4 finalColor = oColor * texture(uTexture, oUv);
+	vec4 finalColor = vec4(oColor.rgb, 1.0) * texture(uTexture, oUv); // NOTE: only want the RGB of the color, A is passed in as part of opacity
 
 	if((uMaterialFlags & 1) == 0 && (uMaterialFlags & 4) == 0)
 	{
--- a/source/LibRender2/Camera/Camera.cs
+++ b/source/LibRender2/Camera/Camera.cs
@@ -334,9 +334,9 @@ namespace LibRender2.Cameras
 		}
 
 		/// <summary>Unconditionally resets the camera</summary>
-		public void Reset()
+		public void Reset(bool ReverseDirection)
 		{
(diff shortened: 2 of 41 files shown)


ginga81:
I built and drove this branch and it was my first experience and I'm thrilled!
I'm really looking forward to the official implementation!

leezer3:
I've merged this into the master branch as an experimental feature.

The signalling logic still needs a complete re-write to account for multiple possible directions, and multiple directional cabs in trains still need looking at. but I think the current mechanics of this work well enough for any further tweaks and bugfixing to be done in the main branch.

Planning to push a new build to the main site, maybe in a week or so.

ginga81:
I have one question, is this only for Rail0, i.e. self-line?
Are you thinking about combining it with OtherRails in the future?
In the case of a double-track route, or even on a single track, the opposite side may be reversed due to a misunderstanding, so is this the first step?
If possible, I would like to incorporate it into the Tohoku Shinkansen and release it...

leezer3:
Eventually with other tracks- It's all building blocks to decouple from the whole Rail0 concept.

Doing this exposed just how much stuff assumed that the direction of travel was positive.
Sound simple doesn't it :)

PR #894, opened 2023-05-03
New: Implement Driver Supervision Device (DSD)

leezer3:
This PR aims to implement a DSD device.
Likely to be XML only feature, although could probably be added to train.dat with relatively little pain.

**Available Intervention Modes:**
* Cuts power
* Applies service brakes
* Applies emergency brakes

**Available Reset Modes:**
* Power handle change
* Brake handle change
* Any handle changed
* Independant DSD key press
* Independant DSD key held

**Available Trigger Modes:**
* DSD is always active.
* DSD is active when moving.
* DSD is active when a direction is set or the train is moving.

### DIFF
--- a/source/ObjectViewer/FunctionScripts.cs
+++ b/source/ObjectViewer/FunctionScripts.cs
@@ -1141,6 +1141,18 @@ namespace ObjectViewer {
 							}
 						} 
 						s++; break;
+					case Instructions.DSD:
+						{
+							if (Train != null && Train.Cars[Train.DriverCar].DSD != null)
+							{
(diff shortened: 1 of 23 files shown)


leezer3:
Fiddle, this never got merged....

It's reasonably complete other than testing and documentation, and should be merged ASAP.

leezer3:
Actually needed a few bits hooking in :/

I'm not an especial fan of the way the updates are handled for controls. It'd be useful to be able to access a collection of all devices on the train requiring control input (this is likely to grow exponentially if we replace most plugin stuff)

PR #417, opened 2020-01-06
Change: unlimited number of tracks for motor sounds (WIP)

s520:
This PR change unlimited number of tracks for motor sounds. (#349)
This feature is not intended to be available in X. It will be available in Y.

I have not tested more than three tracks yet.

PR #896, opened 2023-05-05
Add support for Train.xml to TE2

leezer3:
Manual rebase of https://github.com/leezer3/OpenBVE/pull/417

TE2 hasn't really been touched since the original PR, and was S520's work.
With that in mind, I think this can probably be made to work with the current master, and would be a useful addition, even if some features aren't 100% complete.
I'm also adding XML only features (sanders, DSD etc.) which would benefit from a visual editor.

### TODO:
* New unit attribute needs adding to the main parser. (?? Combine ??)
* Fix build issues, shouldn't be anything too major.
* https://github.com/leezer3/OpenBVE/commit/71bdb397a6455777dc8ce33a3996017497c1b394 needs to be re-applied.


leezer3:
Looking again, this one is too much of a mess.
I've implemented the basic properties into the TE2 export manually.

Will do the same with the rest as I get around to it.
Importing has not yet been done.

PR #903, opened 2023-06-04
Object Viewer 64bit

RakagoBVE:
It's a bit weird that we have 64bit OpenBVE and Route Viewer, so why not a 64bit Object Viewer? It would really help to reduce some loading times when using large textures.


leezer3:
Doesn't work like that I'm afraid.

64-bit would be no faster than 32-bit, and is essentially useless for Object Viewer- All it really allows is for the use of more memory, which is required for some stuff. Frankly, most of this stuff really requires better optimisation, *not* memory, but there you go.....

Loading times should be no different- The only thing that might change slightly is that the main game / route viewer will pre-cache textures to a certain extent.

PR #908, opened 2023-06-30
New: Add GDI independent PNG decoder

leezer3:
This PR starts work on adding a PNG decoder, which is independent of libgdiplus, and is suitable for running on systems lacking the full System.Drawing implementation (e.g. Android)

### Significant API changes:
* Textures are no longer limited to 32-bit RGBA (WIP)

### TODO:
* ~~Add support for the PLTE (color palette) chunk~~
* ~~Add support for the tRNS (color-key transparency) chunk~~
* ~~?? Add support for interlaced files ??~~
* ~~Fix applying color-key transparency to non 32-bit RGBA textures. Probably upconvert them to this, although we could also possibly do something interesting with the shader~~


leezer3:
More work.
The only thing I'm now aware of being broken is interlaced files with palleted colors....

Interlacing isn't really useful for our purposes, but typically there are a few files floating around with this setup, e.g. the route indicator numbers for BRSigs_Open

The API changes also need checking / cleaning up, most were initally quick and dirty hacks to get the thing running.
Could also probably combine some of the palleting code between the PNG and BMP decoders.

leezer3:
Other than checking the API changes, I *think* the decoder should now work as intended for all valid PNG files.

Need to see what the speed is like compared to the GDI one, but shouldn't be significant. 
(Indeed, may be slightly faster, as we're not messing around through an intermediate bitmap)

leezer3:
Speed-
Loading a large [503kb] interlaced file with this decoder takes ~100ms in debug mode
Loading via GDI+ (Win 10) takes ~15ms

Not as good as I'd like there, but this is an absolute worst-case scenario.

A more normal 512x512 texture (palleted, but nothing else) takes ~5ms

**Probably** good enough to merge.

PR #918, opened 2023-08-05
Fix: GL Menu crashing when pressing any key

Kenny-Hui:
Currently pressing any key on the keyboard with the new OpenGL Game Menu would crash because there's no player train on the main menu, this PR should fixes it.

### DIFF
--- a/source/OpenBVE/System/Input/Keyboard.cs
+++ b/source/OpenBVE/System/Input/Keyboard.cs
@@ -14,8 +14,12 @@ namespace OpenBve
 			{
 				//If in kiosk mode, reset the timer and disable AI on keypress
 				MainLoop.kioskModeTimer = 0;
-				TrainManager.PlayerTrain.AI = null;
+				if (TrainManager.PlayerTrain != null)
+				{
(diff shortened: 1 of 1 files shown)


leezer3:
Thanks.

The null check you added when setting the plugin to null is pointless seeing as this isn't being logged or anything, so I've not bothered with it :)

Kenny-Hui:
Hmm are you sure the AI check doesn't matter? If TrainManager.PlayerTrain is null it will still throw an Object reference not set to an instance of an object.

I've tested the new build (2023-08-05) and it seems to crash with Kiosk Mode + GL Menu on.

leezer3:
I *think* I see what you're getting at there actually :/
Forgot about kiosk mode- That's something that must have been there all along. 
The GL menu is a bit of a bodge with the in-game menu structure / input handlers....

See what you think of this:
https://github.com/leezer3/OpenBVE/pull/919
Not really any different to you, just not calling the code path by detecting the GL menu, but actually having it specificed as an interface type should make any required special-casing like this easier.

PR #919, opened 2023-08-05
Change: Actually specificy GL main menu as a different interface type

leezer3:
https://github.com/leezer3/OpenBVE/pull/918

Actually specifies the GL main menu as a different interface type, allowing special handling where required.


source/OpenBVE/Game/Menu/Menu.cs
Kenny-Hui, line 1: Kiosk mode code still gets triggered under GL Menu, if statement inverted?
(PlayerTrain == null ? InterfaceType.GLMainMenu : InterfaceType.Menu

PR #934, opened 2023-08-23
Change: Move cab handles storage to the car

leezer3:
https://github.com/leezer3/OpenBVE/pull/930

This actually appears to be a relatively minor change.
CabHandles are moved to the driver car, and the train handle struct is changed to an accessor.

Can't see any immediate crashes or issues, but not going to merge this at the minute- Needs a way to specifiy different handles on a per-car basis, and then see what happens after that, and what becomes unhappy there when it changes....


leezer3:
Rebased this one to the current master.

Going to test a little more before merging- With the advent of MSTS content and coupling, this one needs to be sorted out.

PR #938, opened 2023-08-29
Fix: HUD Overlay

AmberIsFrozen:
This PR syncs all newly added elements to all size of HUD (Not only default), improve margins and texture on the Large HUD, and fixes a bug where the background texture would be rendered instead of the overlay texture, which reverts the HUD back to the way it used to look.

Before:
image

After:
image


leezer3:
Thanks.

Is there a reason for setting the alpha value of the text to 255 as opposed to 192?
Only thing I'm thinking is that there will be complaints about it having changed.....

Kenny-Hui:
I think having the text being opaque makes the text slightly more legible, although it's just a subtle change and I wouldn't mind reverting for whatever reason.

As for the design, I am thinking to drop the overlay border entirely as it only seems to complicate things in different HUD sizes and looks a bit busy.

I am mainly just bothered by the currently brightened color in the handle and door control, which imo is a bit too bright and clashes with the text color.

leezer3:
Don't disagree in principle.
Thinking sideways a bit, if dropping the border, it ought to be possible to do the overlays in pure GL as opposed to using a texture- All they are is a simple quad after all.

Kenny-Hui:
Probably a good idea, though I am thinking more on expanding the HUD feature instead of dropping the entire customizing the HUD thing, it can come handy at times. (So maybe you specify a subject in interface.cfg and it renders in pure GL, although the interface.cfg format is a mess in of itself)

Or maybe I am just overthinking over a HUD feature...

leezer3:
Yep, definitely a mess.
Only thing with specifying GL in interface.cfg is that would *have* to be GL3 only I'd have thought.

Dropping 1.2 probably ought to be done soon, but meh...

PR #947, opened 2023-09-23
Parser

leezer3:
This PR aims to combine all BVE parsers into a single unified KeyValuePair parser library- Each induvidual parser currently has it's own loop methods to pull out the section headers, split the keys etc.

This causes considerable duplication of code, and as using an entirely new designed internal format (heavily influenced by the MSTS parser library), it'd probably as a side-benefit allow relicening of these parsers to BSD.

Development of the basic idea from https://github.com/leezer3/OpenBVE/pull/544
Experimental, not to be merged without careful consideration.

### Suitable Parsers (?)
* Animated object
* sound.cfg
* extensions.cfg
* panel.cfg
* panel2.cfg

### RISKS:
* Error messages are likely to change with this. Not currently using line numbers in most, but these can be brought back again.
* Introduction of new bugs.
* Check how well some formats are handled.

### FUTURE:
In principle, there's nothing really stopping an XML file from being parsed into this sort of block structure.
A unified parser would in some ways be interesting, but I'm not sure it'd work quite right, as the XML format lets far more properties be set on a per-car basis.

### QUESTIONS:
* Merge with Formats.Msts for a more generic format support library? - Issue there is that the block concepts are differently designed :/


leezer3:
Subject to testing, this is probably now mergable, although I'm still considering chucking this in with the MSTS library as a more generic format support thing.

The XML parsers will probably need a little more work with sub-blocks, but should essentially be able to use most of this code.

I also want to expand this to the RW route parser- Internally converting to CSV is a mess, and this should be quite easily adaptable to do this....

Issue #955, opened 2023-11-01
[Bug] Loading specific route into openbve or route viewer make the program not responding and use too much RAM. 


adfriz:
### Description

as title said.

### Reproduction

just load the route, either in openbve itself or route viewer.

## Route

try load this route: https://www.mediafire.com/file/q9g8ojc55c5ezo7/Yogyakarta-Madiun.rar/file

*my friends found this link, we already test it in newest openbve and route viewer but it make the app not responding.

## Train

no trains.
## Logs

no logs, the app immediately hang and not responding when the route loading progress bar reached 100%.
### Related information 

* Operating system: windows 10
* Openbve: 1.9.2.3 64 bit
* Method of control : keyboard and mouse.

NOTE: my friend said he try load the route with the oldest openbve version he got, he got openbve 1.2.12 and its loaded he said.

here some image i capture when i tried to load the route with newest openbve version.

![a1](https://github.com/leezer3/OpenBVE/assets/76892624/2c581d75-d987-4f56-8845-b2c89827139e)

in my task manager its immediately use too much ram and high cpu usage.
![a2](https://github.com/leezer3/OpenBVE/assets/76892624/c8d1d942-6518-4a27-8317-81e975f76c89)

and here a screenshot from him.

![a3](https://github.com/leezer3/OpenBVE/assets/76892624/8be9bc15-aa9b-46a1-a490-dcf330db8d5d)

not too realistic route in terms of graphics, but im curious what thing that make this route cannot loaded in newest openbve.

leezer3:
So, the primary issue is some quite creatively 'broken' animated objects, which are used as grounds- Large numbers of similar objects:
`
[object]
states=..\Tree\field2.b3d
rotateXfunction=sin[time/3]/2
position=0,-2,29
`

I think the author was attempting to create a swaying grass effect with this.

It will load (eventually), but it's got all sorts of issues, and I'm not sure this is easily fixable.

leezer3:
The PR above gets the loading time back to something more sensible, but I want to do a considerable amount of testing before merging it into the main public branch, as I don't want to break something else.

(Note: I seem to be missing various of the objects with just your link, but this doesn't really matter at this point)

FWIW, I note that this must have been built with an relatively early (pre GL3 I'd say) build from this project, as Michelle's 1.4.3 doesn't load it :)

leezer3:
[...]

Issue #945, opened 2023-09-17
Suggestion: Reduce RouteViewer's reload time


ginga81:
Now, the latest RouteViewer allows you to adjust the drawing distance and has faster drawing speed, making it very convenient.
However, as a result, it takes longer to reload route data using the F5 key.
The Tohoku Shinkansen I am currently building is using 1.8.4.0's RouteViewer, which is the last version that will be before slower.
Using this, reload time only takes about 40 seconds each time.
The current latest version now takes 1 minute and 20 seconds each time.
In order for route data creators to work efficiently during production, they want to shorten reload time as much as possible.
But, even if you increase the draw distance to 300m or 2000m in the latest version, the reload time will not be reduced.
Otherhands, Bve5 has a function to reread the data without reloading the object.
Similarly, at the RouteViewer, which specializes in route drawing, if we implement a function to redraw using the currently loaded object instead of loading object data every time like OpenBVE itself, we can redraw.
When we add or remove the object(s), choice the reread all object or don't, I think it will save a lot of time.
I would like to make a suggestion and hope you will implement it.

ginga81:
https://www.zbx1425.cn/nautilus/quickmap/
I want to reload speed more rapidly at RouteViewer...
This tool is for Bve5, but reload time is faster than one second.

PR #958, opened 2023-11-09
Experimental: Improve forced updates when large numbers of Animated Objects are present

leezer3:
https://github.com/leezer3/OpenBVE/issues/955

The (essential) cause of the above issue is that it uses large numbers of animated objects, all of which contain functions which are costly to execute. 

The game executes a forced update on all world animated objects in the two following circumstances:
* At game start
* After a jump (manual / change ends)

If our world contains thousands of animated objects, each of which contains one or more heavy functions, this can bog down significantly, and cause the slow loading above.

This PR does the following:
* Implements a custom equality operator for animated objects. (This speeds up some aspects of the update)
* On a forced update, only objects placed within 5km of the camera position or having a **TrackFollowerFunction** will be evaluated. 

This brings the loading time of the linked route above to a sensible level, and as a bonus improves loading times for anything using lots of animated objects.

### RISKS:
Actually, I'm not sure, and this is what concerns me a little.
There was obviously a reason for updating all animated objects. but I can't immediately think of one. Any objects present after a jump should be culled before the start of the frame with the standard visibility update routine. 

https://bveworldwide.forumotion.com/t2168-building-a-route-based-on-the-staten-island-railway-integrated-into-the-nyc-subway-system-download-update-10-24-2023#21704

Should be properly tested with the above, as another instance of massive animated function use.....

### DIFF
--- a/assets/Compatibility/RoutePatches/Indonesia.xml
+++ b/assets/Compatibility/RoutePatches/Indonesia.xml
@@ -1,0 +1,11 @@
+<openBVE xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+  <RoutePatches>
+  <!-- Yogyakarta-Madiun -->
+	<Patch>
+		<Hash>FE894F07B945AB4FE30D1805BAC32B44307D2E08D68492C0C47B9F57506DE2DB</Hash>
+		<FileName>Yogyakarta-Madiun.csv</FileName>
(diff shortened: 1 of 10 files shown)


ginga81:
relative
https://github.com/leezer3/OpenBVE/issues/945
Is it possible to check the differences even for non-animated objects and reread only the new objects?
My Tohoku Shinkansen train, which is currently under construction, takes more than a minute every time I read the 460km distance from Tokyo to Morioka by RouteViewer.
The old RouteViewer takes 30 seconds, but it's still faster than the new one, so I'm using the old one now.

Issue #860, opened 2023-01-31
[Question] Untranslated item in Train Editor 2


JunmoreHeavyBox:
### Description

I have found some untranslated items, even though I have added translations to the translation file. There are also untranslated items that are not in the translation file.
Sorry there are a bit too many pictures.

Panel settings section
![驾驶室部分](https://user-images.githubusercontent.com/114210303/215814371-337632af-4795-4625-8a69-3f98a0ab6dda.png)

Sound settings section
![声音引用部分](https://user-images.githubusercontent.com/114210303/215814637-ec545f61-bbbb-4baa-8d5b-468b21b4a869.png)

Some small things
![小地方1](https://user-images.githubusercontent.com/114210303/215814707-e707d3e2-9aeb-4281-ba20-e2fddd114fdd.png)
![小地方2](https://user-images.githubusercontent.com/114210303/215815385-9485a679-1784-4660-be42-ff9247e1a016.png)

Things that are not in the translation file.
1. The Electric Brake in the Delay section of the Car settings tab and the Output button in the Status tab.
![翻译文件中没有1](https://user-images.githubusercontent.com/114210303/215810177-a8ac33d3-8fd2-4337-9c97-63c3bcd4eec0.png)
2. Delay set window of the Car settings tab.
3. The Panel settings tab's Subject window.
4. Touch Element window in the Touch section of the Panel settings tab.
![翻译文件中没有2](https://user-images.githubusercontent.com/114210303/215812595-f16e62a9-aa04-453f-9f72-9779754e90b6.png)
5. Import and Export window.
![翻译文件中没有3](https://github.com/leezer3/OpenBVE/assets/114210303/46810b17-2e5f-414e-8234-abd8e3060d3c)

Only because I was not comfortable with the incomplete translation of Train Editor 2, I did some "Chineseization", mainly by copying existing translations and fixing the machine translation.
I have no intention of contributing to the translation.

### Reproduction

[Test_zh-CN_TrainEditor2.xlf.txt](https://github.com/leezer3/OpenBVE/files/10548240/Test_zh-CN_TrainEditor2.xlf.txt)
This is a sample of my abridged version of the Chinese translation file, please only use it in Train Editor 2.
Using the translations in Train Editor 2 with recognisable differences but all covered should reproduce the problems in the screenshots.

JunmoreHeavyBox:
@leezer3 I've always cared about it, please deal with it.

JunmoreHeavyBox:
[...]

PR #970, opened 2023-11-28
Some cleaning and organising and Chinese language pack

JunmoreHeavyBox:
Finishing work for #860 
- Need to check

#

Update Chinese Language
[Indev Chinese Simplified.zip](https://github.com/leezer3/OpenBVE/files/13481748/Indev.Chinese.Simplified.zip)
- Need to review content

### DIFF
--- a/source/CarXMLConvertor/Convert.TrainDat.cs
+++ b/source/CarXMLConvertor/Convert.TrainDat.cs
@@ -34,6 +34,8 @@ namespace CarXmlConvertor
 		internal static int PowerNotches = 0;
 		internal static int BrakeNotches = 0;
 		internal static double CenterOfGravityHeight = 1.6;
+		internal static double ExposedFrontalArea = -1;
+		internal static double UnexposedFrontalArea = -1;
 		private static MainForm mainForm;
 		internal static List<AccelerationCurve> AccelerationCurves = new List<AccelerationCurve>();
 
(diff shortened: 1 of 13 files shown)


JunmoreHeavyBox:
Urge, I want to audit.

leezer3:
Should be in the build from today- Thanks!

JunmoreHeavyBox:
Found some omissions.

Missed the end tag
https://github.com/leezer3/OpenBVE/blob/26c7b31de5a9956d82b279fb36b6f25e83edf116/assets/Languages/en-US.xlf#L4047

It should be "Centre point"?
https://github.com/leezer3/OpenBVE/blob/26c7b31de5a9956d82b279fb36b6f25e83edf116/assets/Languages/en-US.xlf#L4093

Missing the middle
("panel_settings", "touch", "sound_command")
https://github.com/leezer3/OpenBVE/blob/26c7b31de5a9956d82b279fb36b6f25e83edf116/source/TrainEditor2/Views/FormTouch.cs#L271-L276
("import_export", "train"
, "any_format")
https://github.com/leezer3/OpenBVE/blob/26c7b31de5a9956d82b279fb36b6f25e83edf116/source/TrainEditor2/Views/FormExport.cs#L238

#

By the way the language files carried in this issue are under review?

leezer3:
Technically, neither is probably completely correct for the second- Probably just origin, but I wouldn't worry about that. 

All the others should be OK now.

JunmoreHeavyBox:
Ugh, Please go see the latest commits in my branch, I don't want comment more...

leezer3:
You're right, duplicate...
That's minor though.

A suggestion for you:
If you add a breakpoint here
https://github.com/leezer3/OpenBVE/blob/master/source/OpenBveApi/Interface/Translations/InterfaceStrings.cs#L122
then you'll hit this every time you get a missing string.

leezer3:
Removed the duplicate string, and have sorted out the issues from the save dialog and the fact that the language wasn't being remembered between runs.

JunmoreHeavyBox:
After this job this thing is over.
Go through the entries in the en language file critically, clean up the deprecated entries and organise them to correspond to the translated code being applied.
Work slowly, I expect you to be well rested.

PR #972, opened 2023-12-06
Start re-write of language files

leezer3:
### Description
The current language file loader was written by @S520.
Unfortunately, it's a somewhat monolithic solution, and makes heavy use of loops when finding a string, rather than using a dictionary approach.

This PR is the start of a re-structure / design, so that language files in memory are identical to the XLIFF structure, and use dictionaries to actually access properties.
Semantically, this is probably better.

End-goal of this will be to support a basic translation editor and merge tool- Whilst Transifex is good, I've also got a couple of translators who periodically send updates. 
With this in mind, it'd be very useful to have an editor which automatically populates any missing strings, shows untranslated strings per file etc. etc.

Heavy WIP.


leezer3:
This would seem to be a pretty major peice of work :(
Have done some of the preliminary work, but now the hard part will start.

Unfortunately, I think it *needs* doing.

TLDR:
Translations are stored in the following structure:
Program Group --> (N) sub-groups --> String

Unfortunately, the original design for this uses a single string and a bunch of looping to pull the translation out, rather than nested groups. The underscore is used as a group separator, as well as being valid in string names, which adds to the mess.....

To take an example:
```
buttonselectBox.Items.Add(Translations.GetInterfaceString("denshadego_command_none"));
```

This will change to the following:
```
buttonselectBox.Items.Add(Translations.GetInterfaceString(HostApplication.OpenBve, {"denshadego","command_none"}));
```

Note that the keys are clearly separated, and that the program key is added first, as opposed to a single lump string we need to parse.
The hard part will be altering ~900 string calls and checking that none are broken in doing so....

leezer3:
The only app left to switch to this new access is TE2.
This should be easy, as it actually uses a very similar method of access internally, just combining the parts into a single string call for the API.

Can't remember, but I suspect S520 may well have been working on a lot of this sort of refactor :(

leezer3:
This would appear to work as designed at the minute.
Needs completing and merging before thinking about the translation editor, but that's fundamentally a pretty simple affair.

@MarcRiera - This makes some changes to the DenshaDeGo plugin to support the revised dictionary method of accessing controls. It appears OK, but I don't have one of these (on the list of things to actually buy at some point....)
Can you check that the controls still work with this branch now?

Thanks :)

### TODO:
* Possibly the control dictionary needs to be changed for a sorted dictionary. It appears OK, but should be tested more.
* Convert KeyInfo
* Implement the fallback languages list in XML- This currently is only set to en-US, but for example the Chinese variants would definitely benefit from this.

MarcRiera:
I have tested the DenshaDeGoInput plugin with several kinds of controllers and nothing broke.

Regarding the translation editor, there is [POedit](https://github.com/vslavik/poedit) which is simple but may be enough in case anyone wants to edit translations locally. I have noticed it does not show full ID paths (group+trans-unit), but at least allows editing every single string (Transifex always filters out duplicate trans-unit IDs without considering groups).

For advanced online translation, [Weblate](https://weblate.org) is an open source piece of software widely used by several projects (including CI/CD for merging translations in real time). It can be self-hosted, but there is a free public instance for open source projects: https://hosted.weblate.org/projects/ We could still benefit from online translation and from existing translators from other projects. I have not tested how well it handles XLIFF, but I can run a couple of tests if you are interested.

leezer3:
Poedit has an issue with the current files, IIRC might have been to do with with duplicate names in different groups, but I'd have to go back and dig through- its in one of the original XLF prs.
Don't like the UI much either :)

Some of the issue is that we're a little non-standard in using a single translation file for 3 apps. (5 if I get as far as translations for Object Viewer & Route Viewer)
In hindsight these should be in different files, but restructuring would be backwards incompatible....

leezer3:
The GL menu now looks much nicer when customising a control, but seems to have a problem that it breaks anything with a joystick assignment.

Not sure if this is an artifact of this PR at the minute; need to dig into before merging.
Could also do with higher-res versions of the other icons.

PR #981, opened 2024-01-21
Fix: No translation/text in the Bug Report Dialog

Kenny-Hui:
Fixes a regression from #972, where the translation path from the bug report dialog are mistakenly converted, resulting in no text in the bug report section.
image


leezer3:
Fiddle.
That doesn't open automatically, so missed that one.
I think it demonstrates why in hindsight allowing underscores as a separator & using then as the group idea was a bad thing...

Thanks :)

PR #1016, opened 2024-04-13
Change: Standardize date format in logs

Kenny-Hui:
This PR changes the date format to `yyyy-MM-dd HH:mm:ss` when printed in OpenBVE Logs/Crash Logs, as oppose to the current behavior which varies between system locales.
This should reduce ambiguity, especially relating to places with **mm/dd** vs **dd/mm** (Not that it has happened in a puzzling manner... yet, but seems like a good idea anyway)


leezer3:
Logical enough, if perhaps a little pointless in logs :)

Thanks.

PR #1046, opened 2024-06-03
WIP: Speed up BVE5 map loading

leezer3:
BVE5 route loading is somewhat slower than I'd like at the minute.

Major issues:
1. BVE5_Parsing makes *very* heavy use of reflection, and I suspect went somewhat overboard with the number of types and casts. Future re-design project.....
2. Route.BVE5 has a lot of lists. These need to be inserted into at specific indexes, and are cloned into temp lists, both of which are slow. Need to be converted to ordered lists, and preferably the temp list cloning eliminated entirely.
3. Cycling through the list of statements is slow (as a result of the reflection issues in no 1), and this should be eliminated entirely wherever possible.

### DIFF
--- a/source/Plugins/Route.Bve5/MapParser.cs
+++ b/source/Plugins/Route.Bve5/MapParser.cs
@@ -118,11 +118,7 @@ namespace Route.Bve5
 
 		private static void ConvertToBlock(string FileName, bool PreviewOnly, MapData ParseData, out RouteData RouteData)
 		{
-			RouteData = new RouteData
-			{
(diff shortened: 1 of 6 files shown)


leezer3:
This is now approximately 1/3 faster than the master branch.
I'm not totally happy with the use (abuse?!) of sorted list, but cloning the block list multiple times was painful....

Loading times on Fuji go from ~1min 20s to ~1min
(Route Viewer in debug mode)

I think we can probably get rid of some more itinerations over the statement list, just need to do some digging....

Issue #1023, opened 2024-04-22
will there be a chance for openBVE to finally be supported on MacOS Sanoma on Apple Silicon Based Machines


aairbus016:
is there version fully supported for M1 MacBooks

I compared performances of my old Laptop with a Core I5 10th gen at full performance and the M1 MacBook Air I got both having the same 8gb of ram and my old laptop was faster attuning openBVE.

leezer3:
It's complicated, and largely out of my hands I'm afraid.

Please check that you're running comparible resolution settings (remember that Retina displays have a *much* higher default resolution), and let me have some examples of things that are significantly slower.

To note, I *do not* have access to a M1 Mac.

---

The TLDR:
As you're hopefully aware, the M1 powered series of Apple machines run a completely different CPU architecture to a standard laptop.
This breaks the legacy WinForms main menu, as well as most train plugins, and makes it very hard for me to do anything.
For some very techincal reasons, it's likely to be slower than an equivilant age Windows laptop, but for most content, this is unlikely to be significant. 

Most content for OpenBVE uses 4gb or less of RAM, so it's unlikely a lack of it will cause the problems you're experiencing.

aairbus016:
Thank you for your email

alopix:
I can confirm that openBVE at least runs on macOS Sonoma on an M3 without to many issues. Obviously the menu is very limited and some features fully crash (some options like e.g. the load route shortcut the RouteViewer will crash, so you have to load files via command line etc.).

leezer3:
Thanks.

Object Viewer and Route Viewer use a standard Mono file picker to load stuff, which obviously isn't available on Sonoma. 
(I would suspect that this is another manifestation of the fact that WinForms doesn't work on 64-bit Apple, and has been abandoned by Mono)

I may be able to implement a similar file picker to the main game for the viewers, but further than that it's a bit of an issue.
Unfortunately, writing a GUI is a long, hard task whatever you do it in :)

alopix:
I have been looking into some cross-platform C# UI frameworks yesterday and there always seems to be a catch these days (also, VS Code’s support for C# is appalling which makes trying to contribute to openBVE not fun…). I feel like I have not found a good cross-platform desktop UI lib for any language for that matter (apart from using Qt in C++ or Python). And I am always looking at them given I would prefer not implementing my own stuff over and over again for every desktop platform I would like to support.

leezer3:
[...]

PR #1062, opened 2024-08-29
New: Add GL menus to viewers for platforms not supporting WinForms

leezer3:
WIP implementation of the GL menu to the viewers, in an attempt to improve general UI here on systems not supporting WinForms.

Only currently in Object Viewer, and it doesn't yet do anything, but the main drawing routines etc. now work.

Probably going to do a little re-design in the menu implementation too.

Primarily relevant for OS-X 64-bit (which is most Apple machines now)
https://github.com/leezer3/OpenBVE/issues/1023


leezer3:
Object loading and the basic menu system is now working in Object Viewer.

### TODO:
* Implement options menu.
* Implement error listing. (Probably requires a new menu item type, non-clickable, but with icon)
* Copy to Route Viewer.
* Improve UI in-game- Keys overlay for load object / options should probably go in this case, in favour of a single unified menu key (??)

PR #1064, opened 2024-09-18
Update Linux desktop launcher

marcriera:
The desktop launcher used for Linux still had the old name without the initial caps. I have also translated the comment and generic name to Catalan.

### DIFF
--- a/debian/openbve.desktop
+++ b/debian/openbve.desktop
@@ -1,11 +1,13 @@
 [Desktop Entry]
 Version=1.0
 Type=Application
-Name=openBVE
+Name=OpenBVE
 GenericName=Train simulator
+GenericName[ca]=Simulador ferroviari
 GenericName[de]=Zugsimulation
 Categories=Game;Simulation;
 Comment=Train/railway simulator compatible with 'BVE Trainsim' routes
+Comment[ca]=Simulador ferroviari compatible amb rutes del simulador BVE
 Terminal=false
 TryExec=openbve
 Exec=openbve


leezer3:
Thanks.

Is it worth others, e.g. @ginga81 @Kenny-Hui translating the comment / description too?
These don't seem to be widely translated (and IIRC depend on the WM you're using as to where / if they're displayed), but we do have a reasonable sized non-english community :)

ginga81:
I didn't even know that desktop icons were created in the first place.
I made my own .desktop files in Ubuntu.
I didn't think that installers existed in Linux, and I couldn't find them in the release zip...
My Japanese translation of suggested file modifications
GenericName[ja]=列車運転シミュレーション
Comment[ja]=ライセンスフリー、オープンソースな無料のBve5形式にも対応した列車運転シミュレーション

leezer3:
There's .deb package on the release page, which is generated automatically.
This should work for Debian, Ubuntu and compatables. 

IIRC it adds the icon to the menu under games, not to the desktop by default.

We don't install shortcuts manually or anything, I suppose that's something that the makefile could be altered to do.

ginga81:
I see!
I always unzipped the zip files manually.
I also verified the daily builds manually unzipped, and used official release normally.
I assume it's working fine, too, since no other Linux users have reported any failures.

Kenny-Hui:
I only run the latest nightly build via the archive so I am not aware of the desktop file :P
Translated literally:
```
GenericName[zh_HK]=列車模擬器
GenericName[zh_TW]=列車模擬器
Comment[zh_HK]=與 'BVE Trainsim' 兼容的列車/鐵路模擬器
Comment[zh_TW]=與 'BVE Trainsim' 兼容的列車/鐵路模擬器
```
(zh_CN is deliberately not filled, I am not too familiar with their culture)

leezer3:
Thanks.

I don't know the proportion running the release via deb, but they'll be the less technical users, so it's useful to add these.

PR #1126, opened 2025-03-06
treewide: convert packages.config to PackageReference

lzcunt:
This is a mostly-automated patch generated using `upgrade-assistant`. This is needed because nixpkgs infrastructure uses `dotnet restore` instead of `nuget restore` which does not support the legacy packages.config format.

Draft for now, I opened this PR so CI runs.


lzcunt:
I have no idea why CI fails, I'm not a .NET person. Help is appreciated. I just wanna package this in nixpkgs lol.

Is there any reason why OpenBVE uses mono instead of modern .NET? nixpkgs has better infrastructure for building .NET 6+ packages

leezer3:
The Windows build is failing because the script used to generate version
numbers has been broken somehow.

The Linux build is breaking as its faing to build the main API library
which everything depends on. Not sure why off the top of my head.

Might get to take a look tomorrow, but I'm not installing a niche distro
with funky packaging....

Mono:
'Modern' .Net lacks a lot of things, which the code makes heavy use of.
It'd need the font subsystem completely replacing, as well as loosing the
original main menu to boot.

leezer3:
OK, so doing this has totally broken the csproj files and associated conditionals.

We're using various build conditions to do stuff (specifically generate the version numbers, patch for LBA awareness and some other minor bits) and to move things into the final directories.

You're more than welcome to try and fix this yourself, but the current structures work just fine. 
I don't see any reason why nuget restore shouldn't work on NixOS ( For example, Godot seems to be using it: https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/tools/godot/3/mono/make-deps.nix )

lzcunt:
Godot3 is in fact not using `nuget restore` but instead `dotnet msbuild -t:Restore` which also doesn't seem to support packages.config. Godot4 is using `dotnetCorePackages` which is how .NET Core and .NET 6+ packages are built in nixpkgs. But I should be able to build with `nuget restore` with workarounds so this PR is no longer needed.

PR #1130, opened 2025-03-14
Improve train loading progress bar

Kenny-Hui:
This PR improves the progress bar of the train section in the loading screen, which for some time now cannot reach to 100%:
- Add missing progress increment in PanelAnimatedXmlParser
- Add progress reporting when loading train.xml/extensions.cfg
- Fix `CurrentTrain` counter getting incremented when loading train plugins rather than when loading trains


leezer3:
Thanks!

PR #1163, opened 2025-07-07
Change: Remove Legacy GL1.2 Renderer

leezer3:
This PR removes completely the GL1.2 renderer.

### Rationale:

- The GL4.1 renderer is considerably faster.
- The GL1.2 renderer has no support for complex animations. ( https://github.com/leezer3/OpenBVE/pull/684 )
- The GL1.2 renderer has no support for particles.

### Risks:

- This change considerably narrows the hardware we support. GL4.1 is supported by Intel HD graphics and above, so CPUs from ~2012 onwards, or dedicated graphics cards from approximately the same era.


SlavaArduino12345:
I do not support this change, because that would mean dropping support for toasters (I recently got a perfectly fine 2014ish decent Sony Vaio laptop that I found thrown away, and for some reason it refuses to use the new renderer (I checked both on Windows and on Linux Mint and Route Viewer says that the new renderer isnt available)). Isnt the project about compability, or the compability is for BVE Routes only?

leezer3:
It really depends on what you mean by compatability.

The current version has good compatability with most BVE content. 
Implementing new features (see the top) is difficult enough to do once, but with any graphical feature and the new and old renderers, you're essentially having to implement the same code twice, or choose not to support it. MSTS animations would be unfeasibly slow with the GL1.2 renderer; the new renderer uses the graphics card to perform matrix transforms on thousands of vertices, but we'd have to do this on the CPU with the old renderer.

I don't have a good answer for the 'oldest' machine we should support, but something 11 years old is most likely out of mainstream support with Windows, and pushing it for running one of the heavier Linux desktop environments.

Not merging this yet, and if this goes ahead, I suspect that we'd need to link from the main site the last compatability build.

SlavaArduino12345:
Why is OpenGL 4.3 being used specifically in the first place? And why you/we cant just use 4.0 or even 3.0 instead?

leezer3:
~~I used 4.3 when implementing keyframe animations, for [shader storage buffer support](https://www.khronos.org/opengl/wiki/Shader_Storage_Buffer_Object)~~

~~This PR revises it to 4.1, and uses uniform buffer objects instead, as well as dropping the GL1.2 code.
This is sub-optimal (uniforms are fixed, and we can't do stuff in-shader and have this propagate to the next vertex call), and I haven't yet done any speed testing.~~ 

~~The GL3 message needs updating.~~

Current master incorporates the GL4.1 changes.

leezer3:
Re-done this PR to the current master- Seriously thinking about merging this.

@adfriz

adfriz:
well, the control is yours.

im not against this removal, the last time i test old renderer with my old cpu that has intel hd 3000 with the test route and that fujikyu trains, the results is... somewhat _impractical_ if we call it a gameplay, not even reach stable 20fps on camera idle... its on windowed and on 1366x768 px screen.

 if this merged, some plan come to my mind is 
 1. restructure or cleanup the new renderer code (maybe see how godot engine compability renderer works, AFAIK they limit to GL3.3 for that.)
 2. try to improve the shader to use basic phong shading.
 3. maximize the GL4.1 feature on macos and upgrade to use GL4.3 on linux and windows (to use the compute shader for more performance.)

---
@ginga81 have seen video of my experimental feature showcasing the capability of new renderer to use modern rendering technique like ray marched clouds and atmospheric sky, but the code is very messy and its so heavy to render.

here is the video recorded on windowed mode, using full screen drop it to less tha 30 fps.

https://github.com/user-attachments/assets/d7d7b7c7-cdb8-4966-b9fd-40cd41314c52

ginga81:
Please allow me to comment since spoke to me.
That video was very fresh, interesting, and exciting!

Your comment that it would require a complete overhaul of the rendering engine to make it happen helped me understand why.

I am only 'a comitter', so the rendering engine issue is something that we can only await the decisions and actions of those who can program, but if it is realized, I believe that, like Shadow, this project will remain relevant for decades to come.

My tweet about Shadow received an overwhelming response, exceeding 26,000 views. And that's "in Japan!"
Knowing the situation 10 years ago, I find it hard to believe.
 I think that's amazing.

Once again, my heartfelt thanks to @adfriz and @leezer3.

adfriz:
@leezer3 should we update the shader minimum version to 410 too?

PR #1194, opened 2025-09-20
Fix: Camera Rotation issue & In-game Message not displaying

AmberIsFrozen:
This PR addresses primarily 2 things:
1. In https://github.com/leezer3/OpenBVE/commit/ac6e9576262a94f0177b98f6d359548cb772f1dd, only the keybind for camera left rotation is handled after the refactor, rendering the keys to rotate the camera to the right/up/down/tilting inoperative.
2. The rendering code for message overlay have not been updated since Message Timeout is changed to not be tied to in-game time, rendering any message that didn't add the in-game time invisible. (Bug since v1.11.2.5)
2a. Per above, many messages still include the in-game time when specifying the message timeout (But it appeared to worked fine previously as the rendering code has not been changed), this PR should address most if not all of such cases.

### DIFF
--- a/source/OpenBVE/Game/MessageManager.TextualMessages.cs
+++ b/source/OpenBVE/Game/MessageManager.TextualMessages.cs
@@ -131,11 +131,11 @@ namespace OpenBve
 				{
 					if (Timeout == double.PositiveInfinity)
 					{
-						Timeout = Program.CurrentRoute.SecondsSinceMidnight - 1.0;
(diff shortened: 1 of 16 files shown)


leezer3:
Thanks.
I thought I'd found all the uses of time in message timeouts :/

Controls:
I think that came about because I keep picking changes out of the MSTS branch at the minute. ( https://github.com/leezer3/OpenBVE/pull/684 )

PR #1203, opened 2025-10-15
Viewers fixes

AmberIsFrozen:
- Fix playerTrackDistance & playerTrainDistance throwing errors in ObjectViewer/RouteViewer
- Fix PrimaryKlaxon, SecondaryKlaxon, MusicKlaxon not recognized as a variable in RouteViewer
- Change TFO warning message in RouteViewer to informational message. It's a noteworthy message but given that it's not actionable by the developer and will work properly in the main simulation, I am not certain this should be classified as a warning like any other messages.

N.B. It's interesting to see that the klaxon variables only target the player train as that's not explicitly stated in the documentation (Although I suppose back then there's no real way for non-player train to initiate a horn)

### DIFF
--- a/source/RouteViewer/System/Host.cs
+++ b/source/RouteViewer/System/Host.cs
@@ -562,7 +562,7 @@ namespace RouteViewer
 
 		public override AbstractTrain ParseTrackFollowingObject(string objectPath, string tfoFile)
 		{
-			Interface.AddMessage(MessageType.Warning, false, "Track Following Objects are not shown in Route Viewer. Please test using the main simulation.");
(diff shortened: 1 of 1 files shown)


leezer3:
Yeah :/

The viewers are a mess to be honest, and it's an absolute PITA remembering to update them when something is added. 
I'm currently working on a basic vaccum brake for the MSTS parser (In *theory* this is remarkably simple, after all the physics are only really an air brake in reverse at the end of the day), and this is just showing quite how entangled some of this stuff is.

leezer3:
I've added the missing horn instructions as no-ops in both viewers.
Route Viewer will now use the camera track position for TrackPosition based instructions. 
Object Viewer will now return zero for these.

Have also picked out the change from error to info message. This should sort these.

PR #1223, opened 2026-01-07
Adding Swedish translation

yeager:
This is the first release of the Swedish translation. I am unable to verify what is wrong with it since I can't build OpenBVE either from VS, Mono or Linux. There are too many compiler errors. Maybe someone can verify what is wrong with the file?


yeager:
Yeah, I followed the build guide but it fails. I could try a nightly build tomorrow perhaps

leezer3:
Thanks. will sort that in a second.

(I would have hoped that something that simple would have been OK via machine translation, but obviously not....)

PR #1256, opened 2026-04-07
Implement autoreload for route viewer.

adfriz:
- most of it almost the same like object viewer one
- add a freeze frame for transition if manually reloaded to make it not flicker when autoreload turn on.


ginga81:
I'm sorry for talking about this when I can't program.

How about to get diff check nuget?

DiffPlex (nuget) Apache 2.0 license

The fact that someone as skilled as you is struggling with this means that, we can't simply put, when reading the file, it stores the entire string of text as is, and when it detects a change, is it?
difficult...

PR #1264, opened 2026-04-15
Change: Optimize csv b3d object parser and object viewer autoreload performance

adfriz:
- Implement HashSet face lookup in Plugin.Parser.cs
- Add IO caching for File.Exists results during texture loading.
- Optimize NumberFormats.TrimInside to reduce allocations.
- Implement partial reload in ObjectViewer via UnifiedObject caching.
- Add descriptive code comments for better readability.

### DIFF
--- a/source/ObjectViewer/ProgramS.cs
+++ b/source/ObjectViewer/ProgramS.cs
@@ -41,6 +41,8 @@ namespace ObjectViewer {
 		/// <summary>The number of failed auto-reloads</summary>
         internal static int AutoReloadFailureCounter = 0;
         private static double reloadCheckTimer;
+		/// <summary>The list of loaded objects, cached to allow partial reloads</summary>
+		internal static UnifiedObject[] LoadedObjects = new UnifiedObject[0];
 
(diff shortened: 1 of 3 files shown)


leezer3:
Thanks.

As a small request-
Can you try and preserve the existing comments please.
I've found so many oddities over the years, it's very useful to know exactly *why* a hack has been added.

Other thoughts-
I'm not sure your FileExistsCache will work too well if a file is deleted or added. It's fine for a single load, but especially now we've got auto-reload kicking in, might want revisiting.

adfriz:
> Thanks.
> 
> As a small request- Can you try and preserve the existing comments please. I've found so many oddities over the years, it's very useful to know exactly _why_ a hack has been added.
> 
> Other thoughts- I'm not sure your FileExistsCache will work too well if a file is deleted or added. It's fine for a single load, but especially now we've got auto-reload kicking in, might want revisiting.

ok, ill put the comment back.

it works on local builds but your concern is right, ill do more test about it and i think its better to put this PR as a draft.

PR #1265, opened 2026-04-16
Fix TrainEditor2 startup hang and initialization crash

adfriz:
- Fix infinite loop in configuration parser when encountering unknown sections.
- Add safety guards in AppViewModel to prevent out-of-range access during startup.

Tested on windows 11, local release and debug builds.

as usual i cant test on linux and macos, also im not familiar with this traineditor2 i though it was for editing train.dat files but it only accept .te file on the open file menu.


leezer3:
Thanks.

TE2 can import an existing train.dat based train from the menu, but doesn't open it directly.

Issue #986, opened 2024-02-04
[Suggestion] Multi position object for .animated files


adfriz:
### Description

as the title said.

The current method to write multiple animated object in .animated is like this:

```
[object]
states = chair.obj
position = -0.6,0.1,-4
RotateYDirection = 0, 1, 0
RotateYFunction = some formula

[object]
states = chair.obj
position = -0.6,0.1,-2
RotateYDirection = 0, 1, 0
RotateYFunction = some formula

[object]
states = chair.obj
position = -0.6,0.1,0
RotateYDirection = 0, 1, 0

[object]
states = chair.obj
position = -0.6,0.1,2
RotateYDirection = 0, 1, 0
RotateYFunction = some formula

[object]
states = chair.obj
position = -0.6,0.1,4
RotateYDirection = 0, 1, 0
RotateYFunction = some formula

[object]
states = chair.obj
position = 0.6,0.1,4
RotateYDirection = 0, 1, 0

[object]
states = chair.obj
position = 0.6,0.1,2
RotateYDirection = 0, 1, 0
RotateYFunction = some formula

[object]
states = chair.obj
position = 0.6,0.1,0
RotateYDirection = 0, 1, 0
RotateYFunction = some formula

[object]
states = chair.obj
position = 0.6,0.1,-2
RotateYDirection = 0, 1, 0
RotateYFunction = some formula

[object]
states = chair.obj
position = 0.6,0.1,-4
RotateYDirection = 0, 1, 0
RotateYFunction = some formula
```

The only difference in this code just every object is in different position.

So i propose a new way to simplify the code to be like this:

```
[object]
states = chair.obj
RotateYDirection = 0, 1, 0
RotateYFunction = some formula
position = -0.6,0.1,-4
position = -0.6,0.1,-2
position = -0.6,0.1,0
and so on....
```
or like this, add a separator if the addon creator prefer to use one line.

```
[object]
states = chair.obj
RotateYDirection = 0, 1, 0
RotateYFunction = some formula
position = -0.6,0.1,-4 & -0.6,0.1,-2 & -0.6,0.1,0 and so on....
```

or use it in [include] section for static object to be like this.
```
[include]
chair.obj
position = -0.6,0.1,-4
position = -0.6,0.1,-2
position = -0.6,0.1,0
and so on....
```
```
[include]
chair.obj
position = -0.6,0.1,-4 & -0.6,0.1,-2 & -0.6,0.1,0 and so on....
```
----------------------------------
[...]

PR #1266, opened 2026-04-16
Implement multi position object in .animated file

adfriz:
Now we can write multiple position in single [object] section

before

```
[Object]
states= chair.b3d
position= 0,0,0

[Object]
states= chair.b3d
position= 0,0,2

[Object]
states= chair.b3d
position= 0,0,3

and so on...
```

now

```
[Object]
states= chair.b3d
position= 0,0,0
position= 0,5,0
position= 5,0,0
position= 0,0,-5
```

although it works on object viewer, its obviously need more test and make sure not break other code.

some practical use is player making passenger car that have 2 + 2 row seats with multiple column and want tweak each position a bit. but really any object player wants, like maybe player wants to scatter trees across area and want exact position for each tree.

### DIFF
--- a/source/Plugins/Formats.OpenBve/Block.cs
+++ b/source/Plugins/Formats.OpenBve/Block.cs
@@ -43,7 +43,7 @@ namespace Formats.OpenBve
 	{
 		internal readonly List<Block<T1, T2>> subBlocks;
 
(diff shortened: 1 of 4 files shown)


adfriz:
related to #986

PR #1268, opened 2026-04-20
Fix Shadow visibility and settings reload in Route/Object Viewer

adfriz:
- Update Route Viewer FormOptions to return DialogResult.OK when shadow settings are modified, ensuring the main loop triggers a reload.
- Disable shadow casting on the background cylinder in Route Viewer to sync with main openbve exe and prevent visual artifacts.
- Set DialogResult.OK in Object Viewer options on close to ensure shadow settings are correctly applied immediately.


leezer3:
Ah, thanks, was digging in that direction.

ginga81:
I tried downloading it again, but it doesn't seem to have changed.

PR #1269, opened 2026-04-20
Fix shadow rendering, UI bias controls, and renderer switching crash

adfriz:
- Fix locked shadow bias controls and settings reload logic.
- Fix crash when switching between Old and New renderers.
- Sync background shadow rendering state across viewers.


adfriz:
idk why that bias bug not resolved after shadow PR merged... im pretty sure have done it before...

leezer3:
Might well have been when I based it to get rid of the merge commits :/

leezer3:
I've picked the relevant commit manually.

If you reset your master to the current, this should get us to a relatively clean state.

adfriz:
thanks.

handling git is hard... i think ill keep my master clean and only sync with upstream from now, and ill make new branch for every fix.

PR #1271, opened 2026-04-22
Try Fix apple gl error

adfriz:
@leezer3  maybe this would work?

it successfully build in visual studio on my side but this apple specific so i cant test it.

same as your finding, it make some dummy texture.

Note: i may less responsive start from tomorrow, real life stuff.

### DIFF
--- a/source/LibRender2/BaseRenderer.cs
+++ b/source/LibRender2/BaseRenderer.cs
@@ -247,6 +247,8 @@ namespace LibRender2
 		protected internal Texture _programLogo;
 
 		protected internal Texture whitePixel;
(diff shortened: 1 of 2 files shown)


leezer3:
Thanks!

That seems to be working.
Shadows appear different to the desktop at the minute (for example I've got no shadow from the platform roof on the animated object demonstration route), but that may be because the settings aren't the same between platforms. That *shouldn't* be the case but I need to check everything over a bit more

Other possibility is that the Mac (Hackintosh) has integrated Intel graphics which are messing things up.

Need to do some digging on that front :)

PR #1273, opened 2026-04-24
Implement parallel route loading and add performance stats

adfriz:
Still heavily WIP, from my personal test some route can be 4 seconds faster from around 7 seconds without it.

Need more test for macos and linux.


leezer3:
Something isn't quite right with the threading here- I'm not entirely sure what yet.

Take a (very) large folder of routefiles, and open in the main game.
Then start quickly shifting between them.

At some point, the TextureManager is disposed of by something, and this then chucks exceptions on every further texture it attempts to load. 

I'd also be a little cautious of how large the caches are allowed to get.

PR #1275, opened 2026-04-24
Refactor librenderer2 to modular rendering pipeline and state management

adfriz:
Overhauled LibRender2 to use a multipass pipeline architecture.

- Split rendering logic into specialized passes (Geometry, Sky, Shadow, Overlay).
- Added SceneManager to decouple visibility logic from the renderer.
- Implemented GraphicsDevice for centralized OpenGL state caching.
- Introduced GpuResourceManager for managed resource lifecycles.
- Unified the rendering loop across all application modules.

Although it works locally, it have some bugs like shadow not working, some weird halo effect around texture, ingame gui looks odd, 2d cab looks too bright.

WIP, not tested in macos and linux.

currently only synced to commit d37ec8e from upstream master.


adfriz:
image

some preview

adfriz:
good, the ci build without error.

synced with latest master, the bright bug now gone but, halo effect still persist and now the ingame gui clipped with trains and other close object.... it seems something wrong with the depth test 
image

leezer3:
Without digging too far into what you've done, I'd suggest that you probably want to convert the rest of the SetBlend calls from the base renderer.
(The loading screen, text, and some face parts)

Looks like the blending mode is off to cause the halo.

leezer3:
The cab overlay faces aren't right at the minute- I'm seeing a lot of stuff in 2D (BVE4 / BVE2 era cabs) being glitchy.

adfriz:
Mostly sorted out, still need to check other stuff like bve5 route parser.

image

adfriz:
@leezer3 

on master, shadow currently not affected if using night time texture or its emissive material. do you think i should give player 2 new checkbox in the shadow settings to allow player to choose if shadow cast on faces with night time texture and emissive material?

it should say like:

```
[x] Cast on Night time texture
[x] Cast on emissive material 
```

because some trains content like the fujikyu 6000 above wont cast shadow on the interior because it use night time texture.

leezer3:
Less options is better really.
*probably* cast on nighttime, but not emissive.

I will note the following which you may or may not have seen:
https://github.com/leezer3/OpenBVE/wiki/Errata#lighting-behaviour-with-a-defined-daytime-and-nighttime-texture

See also the linked issue.
Basically, whatever you do, it'll break something.

PR #1276, opened 2026-04-24
Tweak sun light labels and fix azimuth/elevation calculation

adfriz:
i think making it sync with the docs is easier for route creator and player.

@leezer3 ready to check


adfriz:
wait it should be `.lightdirection` not `.directionallight`

adfriz:
@leezer3 ready to check.


btw, is `.lightdirection` command can only declared once in one route file or we can use multiple times in one file with different value?

i plan to add an override checkbox so it can automatically change the parameter in the route file too.

leezer3:
Only once, although that's only because there isn't anything setup to allow changes.

It can however be modified by dynamic lighting.

adfriz:
i think ill postpone that override parameter.

this PR is ready.

leezer3:
Sorted.

PR #1278, opened 2026-04-26
Overhaul mouse input system and look-around camera

adfriz:
- Implement smooth mouse look-around using System.Windows.Forms.Cursor.
- Fix camera spinning and choppiness by bypassing OpenTK event lag.
- Add dynamic sensitivity scaling based on zoom level (FOV).
- Add support for mouse scroll and extra buttons in input processing.
- Update Indonesian and English translations for new mouse controls.
- Add default control mappings for mouse input.

Note: there is a small bug, if we in look around mode (right click mouse) and look backward, anything not loaded like got culled and it need to be triggerd by something like moving the camera position a bit or use arrow key once.

@leezer3 i need your help about that bug.


adfriz:
preview, i set scroll up/down as zoom in/out and we can do interesting stuff like dolly zoom.

*i need to compress this video because github upload limit

https://github.com/user-attachments/assets/04f8ecad-1e91-49aa-a7bf-6f9a68980fa6

leezer3:
Hmm.

The issue with using the System.Windows.Forms.Cursor approach is going to be on platforms other than Windows- Without checking specifically, it's unlikely to work at all on Mac, and Linux may or may not work.

To immediately fix your bug, set the alignment direction, not the alignment directly.
Alternatively, you can call the UpdateViewingDistances method from the renderer.
By setting the aligment directly, you're bypassing the viewing distances update, and so the camera still thinks it's got the same backwards viewing distance as when it's pointing forwards.

adfriz:
ok its done.

@leezer3 please check.

adfriz:
Nice all checks green. I think this feature is done...

@leezer3 can you have a look?

@ginga81 do you want to test this mouse input feature on linux?


How to: 
Go to customize control page.
choose some controls you want to edit.
choose mouse input and set the button and modifier.

My test scenario is to change:
Zoom in : CTRL + SHIFT + Scroll Up
Zoom out: CTRL + SHIFT + Scroll down
Toggle mouse Grab mode: Right click

if can, please also test modifier swap combination, like:

Zoom in : CTRL + SHIFT + Scroll Up
Zoom out: SHIFT + CTRL + Scroll Up

I hope not broken on other platforms...

ginga81:
I built and ran the application.
I was able to map scrolling to the mouse wheel, zoom in and out using Ctrl+Shift+Wheel, and switch viewpoints via right-click using a camera movement style that differs from the previous one.
Also can swap to the wheel settings.

Personally, I felt the wheel speed was very slow when zooming in and out.
Zooming in and out was faster using the keyboard.

adfriz:
You can change it on options page, it's called zoom camera speed.

I think I'll change the default speed maybe 50.

ginga81:
I just discovered this on the options page.
The default was 30, and I set it to the maximum of 100.
But, I actually wish I could set it to about 300 or faster...

PR #1284, opened 2026-05-03
Fix: nearclip and autoreload settings was stacked in object viewer.

adfriz:
image


adfriz:
wait, i forgot, this PR should be in its own branch.

PR #1286, opened 2026-05-03
Implement smooth camera transition in exterior camera

adfriz:
now camera should move smoothly between cars in camera F2 mode, and also to cab view (only smooth if use 3d cab or using panel.animated).

still need a bit of tweak if we click the camera change too fast.

preview 

https://github.com/user-attachments/assets/7f84594f-1113-4f26-816a-f6cf442e906d

### DIFF
--- a/assets/Languages/en-US.xlf
+++ b/assets/Languages/en-US.xlf
@@ -1204,6 +1204,18 @@
 					<trans-unit id="misc">
 						<source>Miscellaneous</source>
 					</trans-unit>
+					<trans-unit id="camera">
+						<source>Camera options</source>
+					</trans-unit>
+					<trans-unit id="camera_interior_transition">
+						<source>Smooth interior transition</source>
+					</trans-unit>
+					<trans-unit id="camera_exterior_transition">
+						<source>Smooth exterior transition</source>
+					</trans-unit>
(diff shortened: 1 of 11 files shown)


leezer3:
I think the easiest way to fix most of the glitches is probably just to add a check as to whether the camera is currently transitioning- see my last commit.

I will say though, that this is probably personal preference as to whether you want it smooth or not. I would add an option, but that seems a little overboard...

adfriz:
Preview.


https://github.com/user-attachments/assets/f63f4478-8a10-4076-bca0-eab5a0a0a336

Dont mind the missing car, i didnt edit the train.dat

@leezer3 can you test it?

if all ok, ill add the checkbox in options page for player to choose to turn on/off and tweak the transition speed.

leezer3:
That's a lot better :)

(I've rebased to master and squashed to get rid of the noise WIP stuff)

adfriz:
wait, i shouldn't do merge branch right?

leezer3:
You don't need to merge in the master unless there's conflicts to resolve, or something new needed from it.

adfriz:
I think this is done, @leezer3 please check.

is the label for camera smoothing checkbox clear enough? or you can tweak it.

PR #1252, opened 2026-04-02
Implement basic directional shadow CSM with PCF

adfriz:
- Add CascadedShadowMap and CascadedShadowCaster to LibRender2.
- Implement shadow depth shaders for depth pass rendering.
- Update default shaders to support shadow sampling and application.
- Add shadow configuration settings (Resolution, Distance, Cascades, Strength) across OpenBVE, RouteViewer, and ObjectViewer.
- Update math library with frustum and matrix helper functions.
- Add localization for shadow settings in English and Indonesian.

Note: 
Only tested locally in release and debug builds on windows.
Need further fix, not guaranteed 100% works on all route and trains.
Need more test for linux and macos.

PR #1288, opened 2026-05-05
Restructure shadow mapping code

adfriz:
Restructured the shadow system into a modular manager-based

what changes:

Shadows Manager: Created `LibRender2.ShadowMapping.Shadows` to encapsulate all CSM functionality (initialization, FBO management, depth pass orchestration, and shader binding).
Decoupling: BaseRenderer.cs was cleaned of several hundred lines of shadow-specific logic. It now delegates to the Shadows manager.

Uniform Renaming: 
Old: `uCascadeFarDist` -> New: `uShadowSplit[N]`
Old: `uCascadeCount` -> New: `uShadowCascadeCount`
Also renamed `CascadeFarDistances` to `SplitDistances` in the logic.

Precision Fix for Large Polygons
Moved the vViewDepth calculation from the vertex shader to the fragment shader. This ensures that large polygons (like the 500m ground or floor mesh in B3D) receive accurate shadow calculations even when their vertices are far outside the shadow frustum, it was a glitch when i test in object viewer if shadow range set to near (150m).

Namespace Fix: Renamed namespace to `LibRender2.ShadowMapping` to prevent conflicts with the `BaseRenderer.Shadows `property.
Also fixed visibility of internal fields like nullDepthMap.

### DIFF
--- a/assets/Shaders/default.frag
+++ b/assets/Shaders/default.frag
@@ -28,26 +28,32 @@ in vec4 oViewPos;
 in vec2 oUv;
 in vec4 oColor;
 in vec4 oLightResult;
-uniform sampler2DShadow uShadowMap0;  // Cascade 0 (near)
-uniform sampler2DShadow uShadowMap1;  // Cascade 1 (mid)
-uniform sampler2DShadow uShadowMap2;  // Cascade 2 (far)
-uniform sampler2DShadow uShadowMap3;  // Cascade 3 (extra far)
-uniform float uCascadeFarDist0;
-uniform float uCascadeFarDist1;
-uniform float uCascadeFarDist2;
-uniform float uCascadeFarDist3;
-uniform float uCascadeBias0;
-uniform float uCascadeBias1;
-uniform float uCascadeBias2;
-uniform float uCascadeBias3;
-uniform float uNormalBias0;
-uniform float uNormalBias1;
-uniform float uNormalBias2;
-uniform float uNormalBias3;
-uniform float uShadowStrength;
-uniform bool  uShadowEnabled;
-uniform int   uCascadeCount;  // 2, 3, or 4
-uniform vec2  uAlphaTest;
+// --- SHADOW MAPPING ---
+uniform bool              uShadowEnabled;
+uniform float             uShadowStrength;
+uniform int               uShadowCascadeCount;
+
+uniform sampler2DShadow   uShadowMap0;
+uniform sampler2DShadow   uShadowMap1;
+uniform sampler2DShadow   uShadowMap2;
+uniform sampler2DShadow   uShadowMap3;
+
+uniform float             uShadowSplit0;      // Boundary where cascade 0 ends and 1 begins
+uniform float             uShadowSplit1;      // Boundary where cascade 1 ends and 2 begins
+uniform float             uShadowSplit2;      // Boundary where cascade 2 ends and 3 begins
+uniform float             uShadowSplit3;      // Final shadow distance boundary
+
+uniform float             uShadowBias0;
+uniform float             uShadowBias1;
+uniform float             uShadowBias2;
+uniform float             uShadowBias3;
+
+uniform float             uShadowNormalBias0;
+uniform float             uShadowNormalBias1;
+uniform float             uShadowNormalBias2;
+uniform float             uShadowNormalBias3;
+
+uniform vec2              uAlphaTest;
 uniform sampler2D uTexture;
 
 struct Light
@@ -66,7 +72,6 @@ in vec4  vPosLightSpace0;
 in vec4  vPosLightSpace1;
 in vec4  vPosLightSpace2;
 in vec4  vPosLightSpace3;
-in float vViewDepth;
 uniform int uMaterialFlags;
 uniform float uBrightness;
 uniform float uOpacity;
@@ -78,8 +83,8 @@ uniform float uFogDensity;
 uniform bool uFogIsLinear;
 out vec4 fragColor;
 
-/// Samples a single cascade with 4-tap PCF via hardware comparison.
-float SampleCascade(sampler2DShadow shadowMap, vec4 posLightSpace, float bias, float normalBias)
+/// Samples a single cascade using hardware PCF.
+float GetCascadeShadowFactor(sampler2DShadow shadowMap, vec4 posLightSpace, float bias, float normalBias)
 {
     vec3 projCoords = posLightSpace.xyz / posLightSpace.w;
     projCoords = projCoords * 0.5 + 0.5;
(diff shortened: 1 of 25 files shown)


adfriz:
good the CI passed, we need more test so its not a regression.

If all ok ill made another PR for implemneting PCSS for better smooth shadow, after this PR approved.

ginga81:
https://github.com/leezer3/OpenBVE/pull/1252#issuecomment-4235604520
I tested it with this build.
The shadow pixels became rougher, but the shadows no longer disappear like in the video.

https://github.com/user-attachments/assets/6cee0b58-e48d-4c1b-81df-78e8f6f88b1b

adfriz:
The shadow no longer disappearing? it means good.

and can you give a screenshot how rough the shadow become? and the shadow settings too

ginga81:
https://github.com/leezer3/OpenBVE/pull/1252#issuecomment-4235604520
shadow-lost.mp4's 0:08, the shadow seems sharpen.
In the April 18th build, it was displayed like this and looked very clear.

Screenshot_20260413_084929
Screenshot_20260413_085720
This build
Screenshot_20260505_125456
Settings
Screenshot_20260505_125700
In the April 18th build, it displayed clearly like this.
This build is at 50%, but the first one was at 70%.
Does changing the strength make it blurry?

adfriz:
strength usually only make it darker or transparent.

please try latest daily builds and compare it with this one, and please make sure to use exact shadow settings

ginga81:
I tested this with the daily build on 20260504.
The settings screen is shown here.
Screenshot_20260505_150846
Screenshot_20260505_150913

adfriz:
@ginga81 try build this one.

i tried to fix this part at the moment, please screenshot again at this place

image

ginga81:
https://github.com/leezer3/OpenBVE/commit/d55365d76d14580125f222916765f5deb13ae507
I built and ran this source code.
Here are the settings and results.
Screenshot_20260505_160945
Screenshot_20260505_161347

adfriz:
i cant find the cause for that currently... ill try take a deep look tomorrow.

in the mean time ill fix other stuff first.

btw @ginga81 is the latest tohoku shinkansen route you use is the v1.00(beta2) zip ?

adfriz:
Commit 52b3b07

fix this : 
image

and disable shadow on glow object

image

ginga81:
Yes, that's correct.
diagram is 'Thank you Max', it's at the end of the platform just before Fukushima.
Please move to 255958m in the route viewer.
X=11.54, Y=2.8

adfriz:
image
what sun direction azimuth and elevation that you use? i want to make it look the same as you first

ginga81:
The latest self-build and daily build's versions screenshot was taken using OpenBVE.
I mentioned the distance and coordinates in RouteViewer to indicate the location you want to see.
I am sorry for any misunderstanding.
Route.LightDirection=30;31

adfriz:
oh sorry its my bad, i was loading the snow version.

adfriz:
this is odd, after that fix and full rebuild i cant open the route viewer, got index out of bound.

maybe ill revert this commit

adfriz:
now it works again, i still not sure at ` float activeBias = bias * (1.0 + biasScale * normalBias); ` the constant should be 1.0 or  not

leezer3:
> now it works again, i still not sure at `float activeBias = bias * (1.0 + biasScale * normalBias);` the constant should be 1.0 or not

I'd strongly suspect you'll find the best value for the bias will depend on all of the other settings.

adfriz:
> I'd strongly suspect you'll find the best value for the bias will depend on all of the other settings.

correct, i was testing same shadow settings with multiple routes and the results is expected... some good some kinda off...


But my other plan is if this PR good enough and accepted, we may open the bias settings as route command for route developer, and let the route file override user shadow bias settings for current game session.

maybe under `with route` namespace.

like:
```
with route
.ShadowBias 0.000050
.ShadowNormalBias 2.00
```
Only bias, no other shadow setting exposed. I dont want scenario like route dev set ultra quality shadow but the player PC is limited or low-end.

leezer3:
Merged this one.

I'd say the appropriate place for anything like that would be the Options namespace.

ginga81:
It might have been implemented already, but if we can only specify it once in Options, it would be difficult to handle things like changing from sunny to cloudy to rainy weather.

Similarly, if we can't control the shadow density on the route side, it would be difficult.

leezer3:
That's a good point.

I think the shadow settings would probably want to live in the dynamic
lighting XML file for any changes.

We'll need to take a look at changing the settings on the fly- I think it
can probably be done, but it'd have to be done at the *end* of a render
pass, which may complicate things a bit.

ginga81:
I'm writing this down as a memo so we don't forget it for the future.
I was heard that they wanted shadows from fluorescent lights and other lighting in the subway.

ginga81:
When I placed a 140obj x 2000m building object using FreeOBJ and enabled shadows, the frame rate dropped from 60fps to 13fps.
When the shadow disable, it resume to 60fps.

This gave me an idea: if there's a way to forcibly disable or intentionally enable shadows in any FreeOBJ or Animated file, it might be possible to adjust settings on the route data side to reduce the FPS load.
For example, how about ShadowOBJ function, like FreeOBJ, to toggle shadows on or off?

The screenshot shows that after adjusting the shadow settings as much as possible to restore the FPS, I achieved 30fps.
FreeOBJ140x2000m

adfriz:
> When I placed a 140obj x 2000m building object using FreeOBJ and enabled shadows, the frame rate dropped from 60fps to 13fps. When the shadow disable, it resume to 60fps.
> 
> This gave me an idea: if there's a way to forcibly disable or intentionally enable shadows in any FreeOBJ or Animated file, it might be possible to adjust settings on the route data side to reduce the FPS load. For example, how about ShadowOBJ function, like FreeOBJ, to toggle shadows on or off?
> 
> The screenshot shows that after adjusting the shadow settings as much as possible to restore the FPS, I achieved 30fps. FreeOBJ140x2000m

Good idea, its already in my plan for shadow feature.

ill put my plan here:

We create a new checkbox in options page in main program, maybe like `Essentials shadow` what it will do? it let the game only allow all except `freeobj` to cast and receive shadow. so trains, rails, grounds, pole, signals etc... will cast and receive shadows.

But what about the `freeobj` ? easy, we already know all the defined `freeobj` from structure namespace right? then we can use that, maybe we can create new namespace like Shadows, and the command is `.Caster `and .`Reciever` the parameter just (freeobj index0, freeobj index1, freeobj index2 and so on...), or we can put it under other namespaces rather than create a new one.

what do you think? @ginga81 @leezer3 

---
Note: this month i think i can't be more responsive as usual, a lot happens. and i already put my other PRs as draft, i think i can't finish those PR.

PR #1292, opened 2026-05-09
Shorten translations for 'Power' and 'Brake' in bahasa indonesia

adfriz:
Some user report that make it hard to read the current power and brake value when playing because how small the UI area for that is.

image

i hope we can use the GL menu soon for drawing that box background and have dynamic width scaling instead using png texture.

### DIFF
--- a/assets/Languages/id-ID.xlf
+++ b/assets/Languages/id-ID.xlf
@@ -1865,11 +1865,11 @@
 					</trans-unit>
 					<trans-unit id="power">
 						<source>Power</source>
-						<target>Tenaga</target>
+						<target>T</target>
 					</trans-unit>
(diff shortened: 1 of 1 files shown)


leezer3:
For the minute, I've just measured the handle strings and used the existing code for growing the box, so this should work without this in today's build.

This is somewhat hacky, but mostly works.

I'd agree that this wants converting to drawing this entirely in shader format really, but trapeziums aren't as easy to produce procedurally as rectangles.

adfriz:
i haven't test that fix, i think ill unshorten that text.


btw i saw you got multiple typos of my name in the Pre release tab 😅

leezer3:
Thanks. fixed the typo.
Copy and pasted one I think...

PR #1294, opened 2026-05-14
Implement modular tooltip system for main launcher

adfriz:
Currently only tested on windows and with english language.

Not all parts got tooltips, i only test it in the interpolation section.

https://github.com/user-attachments/assets/20cbddc4-4f22-46cc-8fe6-cc3918e8a19c

@leezer3 what do you think?
@ginga81 you may want to test this on your linux.

### DIFF
--- a/assets/Languages/en-US.xlf
+++ b/assets/Languages/en-US.xlf
@@ -1421,6 +1421,29 @@
 						<source>UI Scale Factor:</source>
 					</trans-unit>
 				</group>
+				<group id="tooltips">
+					<trans-unit id="interpolation">
+						<source>The texture filtering mode used to smooth textures.</source>
+					</trans-unit>
+					<trans-unit id="anisotropic">
+						<source>Higher levels of anisotropic filtering improve texture quality when viewed at an angle.</source>
+					</trans-unit>
+					<trans-unit id="antialiasing">
+						<source>Reduces jagged edges on polygons.</source>
+					</trans-unit>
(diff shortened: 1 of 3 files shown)


ginga81:
I built and ran it.
Just like in the video, it worked on my Ubuntu 24.04 system.

adfriz:
> I built and ran it. Just like in the video, it worked on my Ubuntu 24.04 system.

thanks for the test 👍

leezer3:
Picked into master, thanks.

Only issue I see is that we've got enough issues with incomplete translations that this will only add to it.

adfriz:
thanks leezer3, i didn't expect this PR merged this soon 😅.

for an incomplete translation, i think i can tweak the tooltips code to make the English tooltips as fallback if the language player use didn't have the tooltip translated. its better than showing nothing while we wait someone to translate it.

maybe next weekend.

---

if im remember, we did use transifex for easier translation right? we may try to integrate it for this. but i dont know how they works and i even forgot how to use it.

leezer3:
Short answer is that we've got a Transifex project, but it's the trouble of actually getting real people to translate the strings that's the issue.

I can do English, and reasonable-ish written French but there's a big difference between casual usuage and actually translating something so it's not unintelligble.
I don't know whether this concept translates too well, but it sums up why I'm not really a fan of AI / machine translations: 
https://en.wikipedia.org/wiki/Chinglish

Issue #1313, opened 2026-05-18
DynamicLight.xml gives wrong shadows vs "static light" (bad positioning) - Possible bug.


GianP-git:
### Description
Testing the "moving sun" using xml file:
- SphericalLightDirection (Azimuth/phi) gives wrong/antithetical results vs Route.LightDirection (works fine) 
- SphericalLightDirection (Elevation/theta) gives wrong/complementary (to 90°) results vs Route.LightDirection

Tool used to get sun position data: https://gml.noaa.gov/grad/solcalc/

Parts of the "With Route," namespace & the xml file hereafter.
**Note**: simulation always starts at 12:23:25 (1st station arrival)

### Reproduction
**Method 1 (fixed sun) *** PERFECT SHADOWS!**
_Ref: Clip_01.jpg_

```php
; ************** New to manage fixed shadows & lighting ***************************
	.AmbientLight 160; 160; 160	; Full light (def.160 all)
	.DirectionalLight 178; 178; 178	; spring/summer noon! (def.160 all)

 ; LightDirection x; x, ; El/theta (90=nadir 0=flat -90=zenit) & invert_Az/phi (0=front 90=R; 180=back; 270=L)
 ; Domo coords: N46° 07' (46.12) - E008° 18' (8.30) - Simulated day: May 15 2026
 ; Domo sun on May 15 at 12:24:00 GMT+2 --- Altitude=60.21 - Azimuth=150.92 (150.92-131+180=199.92°)
 ; Train start direction: 131°
	.LightDirection 60.21; 199.92	; DST sun at ~noon is approaching Sth 
``` 

**Method 2 (moving sun) *** INVERTED SHADOWS (antithetical)** 
_Ref: Clip_02.jpg_           * * * * * * * *   **and TOO LONG** Wrong sun elevation?  (~29°)
```php
; ************** New to manage dynamic shadows & lighting ***************************
 ; LightDirection x; x, ; El/theta (90=zenit -90=nadir 0=flat) & invert_Az/phi (0=back; 90=sx; 180 front; 270=sx)
 ; Domo coords: N46° 07' (46.12) - E008° 18' (8.30) - Simulated day: May 15 2026
 ; Domo sun on May 15 at 04:00:00 GMT+2 --- Altitude=-15.98 - Azimuth=38.49 (38.49-131+180=87.49°)
[...]

PR #1315, opened 2026-05-20
Try fix SphericalLightDirection XML parsing angles and signs (#1313)

adfriz:
for #1313 
there is incorrect shadow directions and angles when using `<SphericalLightDirection>` in XML dynamic lighting.

Added .ToRadians() to degrees parsed from XML before math calculation (for the wrong elevation angle).
Aligned coordinate signs (-dx, -dy, -dz) with Route.LightDirection (for the 180° inverted shadow directions).

I'm not familiar with the xml dynamic light... i hope this fixed it... and didnt broke anything...


leezer3:
Yep. that looks reasonable to me.

I think I must have overlooked the radians when implementing this originally.

PR #1317, opened 2026-05-22
Fix shadow with cartesian coord xml

adfriz:
#1313 

In `DynamicLight.cs` we negate Cartesian `LightDirection` so it stores the vector pointing to the sun, aligning with Spherical coordinates and fixing shadows. 

In `Lighting.cs` we simplify time-block index wrapping logic using modulo and fix cab brightness variable assignment typo.

@leezer3 Please check.

I hope this didn't broke legacy stuff...

### DIFF
--- a/source/LibRender2/Lighting/Lighting.cs
+++ b/source/LibRender2/Lighting/Lighting.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using OpenBveApi.Colors;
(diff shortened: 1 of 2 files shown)


leezer3:
Thanks.

I'll try and have a proper look, but might be a few days.
Bank Holiday weekend here, nuts at work.

It looks reasonable from first glance, but this is a relatively niche thing :)

Issue #1332, opened 2026-06-12
[Discussion] About video parsing/playback as texture.


adfriz:
as title said...

I have done implementing video playback using library, specifically libmpv with their dll file.

although it _works_ and run well locally (only tested with .mp4 in windows), I'm still unsure how this play well in terms of license.

AFAIK, we _could_ just bring the dll as is ? and write the license in the readme as usual?

Note: the code still on my PC maybe i will push it as a PR like always if all clear...

Some preview:

https://github.com/user-attachments/assets/ddfe41f5-9bf0-476f-8f61-576ae8c539a9

---
@leezer3 what do you think?

adfriz:
the object written as usual, using load command.

```
[MeshBuilder]
Vertex 0.0000, 0.6000, -1.0000, 1.0000, 0.0000, 0.0000
Vertex 0.0000, 0.6000, 1.0000, 1.0000, 0.0000, 0.0000
Vertex 0.0000, -0.6000, 1.0000, 1.0000, 0.0000, 0.0000
Vertex 0.0000, -0.6000, -1.0000, 1.0000, 0.0000, 0.0000
Face 0, 1, 2, 3
Load video3.mp4
Coordinates 0, -0.0000, 0.0000
Coordinates 1, 1.0000, 0.0000
Coordinates 2, 1.0000, 1.0000
Coordinates 3, 0.0000, 1.0000
```

ginga81:
If the license can be cleared and implemented, it would be great if it could be played on the interior of the exterior view.
https://www.youtube.com/watch?v=3V9IXHRSU1E&t=7s

I understand that current animated GIFs are designed for around 30 seconds. It would be great if the playback time could be unlimited.

There are people who use it this way, and it would be great to be able to display various movies on displays in trains, stations, and advertisements to create digital signage.

adfriz:
the duration of the video will not get cut, you may try it yourself by following the instructions in #1333 

well... it can be used anywhere, i tried make this as basic as possible. you just need an object like signage and separate the [meshbuilder] for the video part.

let say an signage-object.b3d

```
[meshbuilder]
vertex......
Face....
Load yourtexture.jpg
Coordinates....

[meshbuilder]
vertex......
Face....
Load myvideo.mp4
Coordinates....
```

This _should_ works.

ginga81:
Thank you show me how to use for testing.
If it could be incorporated as an object section within the animated objects, I think it would look like the Crosshatch video. Is that possible?

[...]

PR #1333, opened 2026-06-12
Implement video parsing/playback

adfriz:
Discussion at #1332 
VERY EXPERIMENTAL, ONLY TESTED ON WINDOWS.

- Added mpv_render_context_update flag check to bypass heavy OpenGL renders when no new frame is available.
- Added auto-pause and resume behavior when textures are not rendered/accessed.
- Configured automatic unloading after 20 seconds of inactivity to free GPU memory.
- Added path query parsing (e.g. video.mp4?loop=no) to configure loop settings from b3d/csv files.
- Added strict URL and network protocol checks to block internet endpoints.
- Tuned MPV options (demuxer bounds, fast decoding, skiploopfilter, and thread limits) to prevent CPU/FPS lag.

simple way to use is jut load it like using nomal texture

example : myobject.b3d

```
[MeshBuilder]
Vertex 0.0000, 0.6000, -1.0000, 1.0000, 0.0000, 0.0000
Vertex 0.0000, 0.6000, 1.0000, 1.0000, 0.0000, 0.0000
Vertex 0.0000, -0.6000, 1.0000, 1.0000, 0.0000, 0.0000
Vertex 0.0000, -0.6000, -1.0000, 1.0000, 0.0000, 0.0000
Face 0, 1, 2, 3
Load myvideo.mp4
Coordinates 0, -0.0000, 0.0000
Coordinates 1, 1.0000, 0.0000
Coordinates 2, 1.0000, 1.0000
Coordinates 3, 0.0000, 1.0000
```

By default its looping infinitely, add this paramater in the same filepath as your video file.

Play once: `video.mp4?loop=no` (or `?loop=false` / `?loop=0`)
Loop forever (default): `video.mp4` (or `video.mp4?loop=inf`)
Loop few times (e.g. 3 times): `video.mp4?loop=3`

so its like:

```
[MeshBuilder]
Vertex 0.0000, 0.6000, -1.0000, 1.0000, 0.0000, 0.0000
Vertex 0.0000, 0.6000, 1.0000, 1.0000, 0.0000, 0.0000
Vertex 0.0000, -0.6000, 1.0000, 1.0000, 0.0000, 0.0000
Vertex 0.0000, -0.6000, -1.0000, 1.0000, 0.0000, 0.0000
Face 0, 1, 2, 3
Load myvideo.mp4?loop=no
Coordinates 0, -0.0000, 0.0000
Coordinates 1, 1.0000, 0.0000
Coordinates 2, 1.0000, 1.0000
Coordinates 3, 0.0000, 1.0000
```

---

Note:
1. I only write the code for talking with the dll, for the dll it self you need to compile it from the source https://github.com/mpv-player/mpv or use prebuilt binaries, i use the prebuilt by third party https://sourceforge.net/projects/mpv-player-windows/files/libmpv/ i choose the x64 for 64bit variant.
2. You need to put the `libmpv-2.dll` next to `Tolk.dll `
3. test it, on object viewer is easier.
4. currently i set is as no audio, if you want to have audio, in file `videotexture.cs` line 93 you set that line as comment to enable audio decoder.
5. i limit max video fps playback to 30fps
6. the video may looks stretched, you must tweak your object UV coordinates and it shapes.

@leezer3 if you have time, may take a look?


ginga81:
Screenshot_20260613_201817
Screenshot_20260613_201854
I built and tried to load an 11GB MP4 file, but it didn't play.
It's GB, not 11MB. Since there's no time limit, I tried a larger file.
I've never used b3d before, so I just tried it as is.
I also tried a 400MB file, but that didn't work either.
The codec in the MP4 file I loaded might not have been supported. Does it need to be converted to a supported codec?

adfriz:
please dont use too big video file, try test it with small 720p 10 seconds file.

you need the libmpv .so file, unfortunately im not familiar with linux environment. you need to obtain the libmpv file either compile it from the source or from your package's server.

leezer3:
For Debian / Ubuntu, I think you'll need to first install the libmpv2 package.

Then, edit the OpenBveApi.dll.config file to add the following into the `<Configuration>` block:
```
<dllmap os="linux" dll="libmpv-2.dll" target="libmpv.so.2"/>
```

Not tested at the minute, might try doing that later.

Apple wise, I think you'd have to compile it as a dylib, along with all it's dependancies and add that to the actual app bundle.

I'm not entirely convinced though. This seems like a lot of work (and mess) for a relatively small return, Animated GIF files haven't really been used, and they give much the same result.

ginga81:
I also asked the question because I thought I could play an 11GB file since I heard there was no time limit.

However, as you mentioned, there's a time limit, and the resolution is 720P, and it's less than 30 seconds long, so I thought there are already realize by animated GIFs...
However, if the audio is also played, I think that's an advantage.

adfriz:
> I'm not entirely convinced though. This seems like a lot of work (and mess) for a relatively small return, Animated GIF files haven't really been used, and they give much the same result.

well yeah... it is very time consuming... just assume this PR as working proof of concept.

i tried to narrow it down, so only the video and audio decoder must work if we set video as texture, others feature should be blocked. But in the end, we must compile and tinker with the lib itself and its dependency if we need it for real public release... i don't want us to ship entire 100MB lib...

adfriz:
> I also asked the question because I thought I could play an 11GB file since I heard there was no time limit.
> 
> However, as you mentioned, there's a time limit, and the resolution is 720P, and it's less than 30 seconds long, so I thought there are already realize by animated GIFs... However, if the audio is also played, I think that's an advantage.

i mean... there is _technically_ no limits for size and video length... but if you wanna test, 720p 30fps 10 seconds is good enough.

The audio is supported when i tested it in my side (windows).

PR #1337, opened 2026-06-18
Implement texture compression and downscaling optimizations

adfriz:
WIP, only tested on windows with object viewer, route viewer and main program.

Just realize opengl technically support texture compression such as BC1 / DXT, if we use this this _should_ use less RAM/VRAM

currently not have UI, needs to write manually the setting in the options.cfg or options_ov.cfg or options_rv.cfg under [Quality] sections.

```
compresstextures = true or false
maxtexturesize = 256 or 512 or 1024
```

max texture will downscale only the texture if greater than the resolution specified.
currently i use console logs for estimating the memory savings, AFAIK it wont drop the ram usage immediately, we need to wait the next GC sweep. Also, compressing and downscaling needs time, it will make the loading longer.

### DIFF
--- a/source/LibRender2/Textures/TextureManager.cs
+++ b/source/LibRender2/Textures/TextureManager.cs
@@ -231,6 +231,8 @@ namespace LibRender2.Textures
 			}
 			//Set last access time
 
+			bool compress = renderer.currentOptions.CompressTextures;
+
 			if (texture != null)
 			{
 				handle = texture;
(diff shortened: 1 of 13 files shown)


adfriz:
oops, this must be a PR draft 😅

leezer3:
My initial thoughts would be that this whilst interesting I'm not sure this is really going to give us any decent gains if developers have actually optimised properly.
You're still storing the textures as uncompressed in RAM, and assuming VRAM is available, this is going to do nothing particularly helpful- I suspect the limitation here will be passing the texture over the RAM to VRAM bus.

You could possible compress the in-memory texture to DXT on load and copy the compressed data direct to VRAM, but that would probably come with re-writing the entire texture class.

---

The 'normal' problem trains in terms of RAM / VRAM are NYCTA content, such as those from here:
https://sites.google.com/view/openbvetraindownloads/trains

Essentially, they're an absolute disaster zone in terms of texture sizing (most *way* too big, and not to power of 2 either), total texture counts, and general practices.

PR #1352, opened 2026-07-10
New: Implement replacement CSV B3D object parser

leezer3:
Still in progress.

### Significant Changes:
This PR implements a replacement CSV / B3D object parser, in block format.
It's far cleaner to read, and replaces another chunk of legacy code with unclear licencing with something unambiguously BSD-2 licened.

### TODO:
* ~~Cylinder~~
* ~~Cube~~
* ~~Scale~~
* ~~Shear~~
* ~~Normals~~
* ~~Mirror~~
* ~~Full blending mode calculations not done yet.~~
* ~~Re-implement the hacks- Some *should* not be necessary (e.g. multi-column CSV)~~
* ~~Re-add most error messages.~~

##  Testing:
Marginal at the minute.


adfriz:
A typo in line 356 in csvb3dfile.cs

Should be `v.Z` for the z axis right?

leezer3:
Yeah, definitely. Needed to increment the array index on that line too.

This is currently very much WIP, and is probably liable to explode at any moment, and is something I'm a little reluctant to do with the troubles that the related CFG etc. parsers caused, but it's so much easier to read this way, plus has the advantage of actually getting to to a clean licence state.

source/Plugins/Object.CsvB3d/Plugin.NewParser.cs
adfriz, line 1: Typo in scalevector.Y == 0

leezer3: Thanks

source/Plugins/Formats.OpenBve/CSVB3D/CSVB3DFile.cs
```diff
@@ -0,0 +1,430 @@
...
+		public override bool GetNextTexturePath(string absolutePath, out string tDay, out string tNight)
+		{
+			KeyValuePair<int, KeyValuePair<T2, string[]>> value = Values.Dequeue();
+
+			tDay = value.Value.Value.Length >= 1 && !string.IsNullOrEmpty(value.Value.Value[0]) ? OpenBveApi.Path.CombineFile(absolutePath, value.Value.Value[0].Trim()) : null;
+			tNight = value.Value.Value.Length >= 2 && !string.IsNullOrEmpty(value.Value.Value[1]) ? OpenBveApi.Path.CombineFile(absolutePath, value.Value.Value[1].Trim()) : null;
+
+			bool textureFound = false;
+			
+			if(File.Exists(tDay))
```
adfriz, line 480: I might reading this wrong, but should we add null check? if some obscure reason someone writes:

`Loadtexture, , nighttex.jpg`

leezer3: The current parser doesn't allow setting a nighttime texture without a valid daytime texture. 

FileExists just returns false if passed null, so this matches the current behaviour.

adfriz: Alright thanks.

So I did reading this wrong 😅

leezer3:
This is now in theory feature complete.  I'm now in the process of testing...

leezer3:
I believe this is now mostly tested. I've also split it into it's own root block, as it was adding a bunch too many single use methods to the root block class

Still testing however.

---

I've currently got one further errata note to add as a result of this:
When an argument of a cube size failed to parse to a valid number (but was not empty), the cube size was set to 1.0

e.g.

```
[MeshBuilder]
Cube a,1,1
```

This produced a cube sized as 1,1,1

BVE2 and BVE4 however produce a cube sized 0,1,1
(Flat square face)

Zero size appears more logical to me here.

Issue #1355, opened 2026-07-16
[Suggestion] Allow animated object as object based background.


adfriz:
as title said.

a practical use of this is a fake animated background, essentially just using `TextureShiftXFunction = time * 0.005` for moving the texture overtime.

see this example in object viewer.

There is 3 main way:

1. moving the main texture background overtime (AFAIK plenty of AAA games still use this).
2. Half dome mesh with moving texture, can be used as clouds layer.
3. Flat quad mesh, also can be used as moving clouds and stacked for fake clouds.

https://github.com/user-attachments/assets/1a9f1108-a029-4fe7-bdc8-df86ce2ae5f2

adfriz:
i think i have done it...

https://github.com/user-attachments/assets/1e52836e-6fb2-4766-a4e9-508a32d344db

i will do a PR soon.

adfriz:
@ginga81 

you may want to test this on your machine.

here is the animated object.

[Animated Object as Background.zip](https://github.com/user-attachments/files/30123195/Animated.Object.as.Background.zip)

Please refer to the documentation here : 
https://openbve-project.net/documentation_hugo/en/routes/xml/dynamicbackground.html

https://openbve-project.net/documentation_hugo/en/routes/csv.html#texture

usually just need to add in the Texture namespace like this:

```
With Texture
	.Background(0) bg.xml
```

you can also edit the formula for that moving texture, less value means slower.

ginga81:
I checked using the object viewer.
An error appeared stating that "dome.b3d" could not be found.
It seems it is not included in the zip file.

adfriz:
oh my bad, here is.

[dome.txt](https://github.com/user-attachments/files/30129901/dome.txt)

just rename it to "dome.b3d"

btw, the you dont need to follow my object naming for using object based background, any name is ok.

and the important part only the sky.b3d because its just a sphere object i made that i tweak the UV a bit.

For the sky.b3d you must use 360 degree panoramic image in jpg or png.

you can use CC0 License panoramic image in ambientcg.com or polyhaven.com, but you need to use the tonemapped or convert it yourself from hdr or exr images.

ginga81:
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/41dd77cc-7913-4cd8-bdfc-08ac1a80fc6b" />
I managed to get the animation working and displayed it in RouteViewer!
I was amazed at how incredible it is!
[...]

PR #1356, opened 2026-07-17
Allow animated object as object based background in routes

adfriz:
related to #1355 

@leezer3 ready for review.


leezer3:
Thanks.
Dunno if there was a reason for not allowing animated objects in the first place, but this seems sensible enough.

I've tweaked it slightly, and will merge this in a minute when CI parses.

adfriz:
thanks, i thought i need to wait for #1352 😅

leezer3:
That's near enough finished I think, might merge that after the weekend.

I'm using it whilst testing generally and I'm trying to run back through the routes I fixed object issues in as well, but I'm trying to be cautious before merging the thing; someone will find something I overlooked the moment it goes public....

PR #1361, opened 2026-07-20
New: Convert train.dat parser to block format

leezer3:
WIP...

This PR converts the train.dat parser to block format.

Same basic idea as the other parsers, but in this case is primarily to get this closer to a clean licence state.
Need to investigate further whether we can modify TE2 a bit to use the common plugin when loading train properties; if we can do this it'd probably get rid of plenty of duplicate code.


leezer3:
Some further thoughts / notes:
The TrainDatKey enum is only used for errors (not in the file). Might well be possible to combine this with the TrainXMLKey. Need to look into that.

Errors are currently commented out, need to add the current line property.

Licence headers wise, we could probably now add BSD-2 to this file when it's been cleaned up- There's still some duplicate bits and bobs.
The remaining code is all 'facts' and 'magic numbers', neither of which would meet the bar for attracting copyright, although technically I suppose I'd have to re-write the way it's structured.
I'm reasonably happy that re-doing the actual parsing bit in this case qualifies it as being sufficiently new though....

PR #1364, opened 2026-07-29
Update Net Framework version to 4.8

leezer3:
This PR updates the base supported .Net framework version to 4.8

**Should** be harmless, supported on everything from Windows 7 upwards.
Should also be OK on Mono, but will need testing in that regards before thinking of merging. Train plugins referencing older versions of OpenBveApi should be OK. but this wants testing too.

Might be better to go direct to 4.8, but let's see what this does first.


adfriz:
i think directly to 4.8 _maybe_ better? the newer JIT compiler _could_ make startup faster and maybe other stuff as well.

https://learn.microsoft.com/en-us/dotnet/framework/whats-new/#whats-new-in-net-framework-48

for windows 7 i think this supported at least minimum SP1.

leezer3:
Might do.
4.8 is odd, as 4.8.1 doesn't do Windows 7, despite looking like a minor point release from the version number. 

Trying to step slowly at the minute though and test each step on all platforms; there's enough legacy cruft around I'd prefer to not be trying to revert though multiple framework steps.
It doesn't help either that Debian Forky seems to have broken WinForms on Mono at the minute.

leezer3:
```
DllNotFoundException

openal32.dll assembly:<unknown assembly> type:<unknown type> member:(null)
```

Debian doesn't work at the minute- have tried bumping to 4.8 as well to be one the safe side...
That's a very odd error too. This might just be a further sympton of Forky being broken

leezer3:
This is a nasty little rabbit hole...

The Mono default terminal driver has a hardcoded magic number (which obviously hasn't yet been updated even in Forky)
The 'usual' fix of setting ``TERM=xterm`` appears to need inverting to ``TERM=linux``

This fixes our WinForms and errors to the console. It doesn't fix the OpenAL issues yet though. Still digging on that one....

leezer3:
OK, really odd...

JetBrains Rider is messing up the build output location on Linux, so the .dll.config file wasn't being copied. Copy that across and it seems to load OpenAL now. 

Will try and test under Apple later, but I foresee issues if Debian is being this problematic. Might want to cherry-pick the last commit into master if this is going to sit around for a while...

leezer3:
OK, so that at least works on Apple with no further changes. Rider is failing to copy the .dll.config file here too though.
I'm sure it wasn't doing that last time I tried, but still....

Will try updating this PR to 4.8 later and see if that builds OK on all platforms again.

Other notes:
Finding the reference assemblies appears to be problematic in some cases (Rider again....). We can work around that by adding the reference assemblies package to our Nuget list for all projects.

leezer3:
Now updated to 4.8
This also contains changes to the makefile to fix it on Forky. These will probably also apply to the most recent Ubuntu versions.

PR #1371, opened 2026-08-01
New: Add easing functions to FunctionScript

adfriz:
Adds lerp, smoothstep and the full set of Robert Penner easing functions (sine, quad, cubic, quart, quint, expo, circ, back, elastic and bounce, each in in/out/inout variants) to the function script API for the animated objects.

Need more test, examples and documentation...


adfriz:
https://github.com/user-attachments/assets/4c215d1e-bcda-4c26-bed5-931686a2b35c

example, not a proper object to animate but at least easy enough to understand.

### Full function list

| Function | Example | Effect |
|---|---|---|
| `lerp` | `lerp[a, b, t]` | linear interpolation between two values (t = 0 → a, t = 1 → b) |
| `smoothstep` | `smoothstep[a, b, x]` | Hermite curve; smooth 0→1, ramps only between a and b |
| `easeinsine` | `easeinsine[t]` | starts slow, ends fast |
| `easeoutsine` | `easeoutsine[t]` | starts fast, ends slow |
| `easeinoutsine` | `easeinoutsine[t]` | slow-fast-slow |
| `easeinquad` / `easeoutquad` / `easeinoutquad` | same pattern | quadratic variant |
| `easeincubic` / `easeoutcubic` / `easeinoutcubic` | same pattern | cubic variant (most common) |
| `easeinquart` / `easeoutquart` / `easeinoutquart` | same pattern | quartic variant |
| `easeinquint` / `easeoutquint` / `easeinoutquint` | same pattern | quintic variant |
| `easeinexpo` / `easeoutexpo` / `easeinoutexpo` | same pattern | exponential, very aggressive |
| `easeincirc` / `easeoutcirc` / `easeinoutcirc` | same pattern | circular curve |
| `easeinback` / `easeoutback` / `easeinoutback` | same pattern | overshoots target (pull-back effect) |
| `easeinelastic` / `easeoutelastic` / `easeinoutelastic` | same pattern | elastic wobble |
| `easeinbounce` / `easeoutbounce` / `easeinoutbounce` | same pattern | bounce (e.g. drop-to-ground) |

Example of a bouncy drop:

TranslateYFunction = (1 - easeoutbounce[mod[time, 4] / 4]) * 1.5

### Notes

- Inputs outside `0..1` are not clamped — `back`/`elastic`/`bounce` intentionally exceed the range mid-curve (that's their overshoot).
- For state-driven animations, use `if[...]` with a `value + delta * speed` accumulator; plain `mod[time, x]` resets abruptly (snaps), so wrap it in `min[mod[time, x], x - mod[time, x]]` or a cosine carrier if you need continuous motion.ap it in min[mod[time, x], x - mod[time, x]] or a cosine carrier if you need continuous motion.
- `t` is usually your formula

---

@leezer3 what do you think? maybe i should not include the lerp?

leezer3:
I suspect that lerp is probably actually the most useful of these actually.
slerp might also be useful, but really the limitations of the animated format are going to start hitting when you get really complex- fundamentally at the minute there's no way to chain animation positions.

PR #1374, opened 2026-08-03
Improve startup time

adfriz:
Trying to make the startup faster, especially the cold start.

from my personal test, cold start down to 4~ second. after the JIT and cache done by the OS, could make the second time app startup to be less than 2 second.

it logged to the log txt in C:\Users\username\AppData\Roaming\OpenBve\Settings

some log will be writen after the app close.

---

need more test on mac and linux and other non latin system locale e.g CJK...

and make sure all plugin loaded and working as usual...


adfriz:
this should make the window appear faster for player that use HDD. idk how much faster, i dont have any HDD laying around...

PR #1380, opened 2026-08-08
Fix macos startup regressions

kiwiaraga2000:
I hit two separate startup problems on macOS with 1.14.0.2.

First, the app bundle was missing `OpenTK.dll.config`, so OpenBVE stopped immediately with the "corrupt installation" message. The file is in the repo, but it wasn't ending up in `OpenBVE.app/Contents/Resources`.

After adding the config manually, OpenBVE got further but the OpenGL menu started crashing and relaunching in a loop. Tracing the exception showed:

```text
System.NullReferenceException
at OpenBve.Graphics.Renderers.Touch.PreRender()
```

`Touch.PreRender()` was reading `TrainManager.PlayerTrain` before checking whether the simulation had actually been set up. On the menu there is no player train yet.

This changes that order and adds a null check.

For the macOS package, it also copies `OpenTK.dll.config` into the app bundle explicitly and checks that the file is there before building the DMG.

Tested on an Apple Silicon Mac, macOS 26.5.2, Mono 6.12. With the menu bypassed using `/route` and `/train`, the simulation itself started normally.

### DIFF
--- a/makefile
+++ b/makefile
@@ -201,6 +201,8 @@ $(MAC_BUILD_RESULT): all-release
 
 	@echo $(COLOR_RED)Copying build data into $(COLOR_CYAN)OpenBVE.app$(COLOR_END)
 	@cp -r $(RELEASE_DIR)/* mac/OpenBVE.app/Contents/Resources/
+	@cp source/OpenBVE/OpenTK.dll.config mac/OpenBVE.app/Contents/Resources/OpenTK.dll.config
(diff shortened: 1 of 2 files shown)


leezer3:
Thanks!

I'm not sure why it's all of a sudden stopped copying the .dll.config file (might have been a build agent glitch I suppose...), I don't think anything has been touched with that :/

Touch wise must have missed that in the main game....

PR #1382, opened 2026-08-09
Expose most openbve settings to route viewer and object viewer.

adfriz:
The same option as object viewer, idk if this useful, but i think its not a bad to expose this so its more align with object viewer?

### DIFF
--- a/assets/Languages/ca-ES.xlf
+++ b/assets/Languages/ca-ES.xlf
@@ -1356,6 +1356,12 @@
 						<source>Enabled</source>
 						<target>Habilitada</target>
 					</trans-unit>
+					<trans-unit id="display_fpslimit">
+						<source>FPS Limit:</source>
+					</trans-unit>
+					<trans-unit id="display_fpslimit_unlimited">
+						<source>Unlimited</source>
+					</trans-unit>
 					<trans-unit id="display_window">
 						<source>Window mode</source>
 						<target>Finestra</target>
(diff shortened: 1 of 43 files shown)


adfriz:
Just realize route viewer didnt have that vsync options, so i think its better align all the feature to what the main openbve has.

as example turning vsync off unlock the fps cap to not limited by the display hz.

image

adfriz:
i think i will add some fps limit options to the main game too.

adfriz:
Nice all checks passed.

@leezer3 i tried making the commit not depends on previous commits, if needed you can cherry pick each commit to master.

leezer3:
Was closing this deliberate?

With regards to VSync / FPS limit, not sure
how much real value this has in development tools, but on the other hand if it's easy to expose...

(To note, there is an older branch / PR *somewhere* about to add a viewer mode to the main game; I can't remember offhand the issues with that, but it might be 'better' to redo that?!)

adfriz:
Yes, ill do these commit again in different branch, ill do small rewrite and arrange some of the code so its easier to cherry pick.

viewer to the main game? is like bundling it in the main window launcher by making a button and pointing to the each viewer exe? or like multiple ingame mode?

leezer3:
It was essentially just adding another dropdown to the game mode box, plus fixing a few related issues to do with no player train.

Basically, when all's said and done, the main game & Route Viewer are nearly identical internally.

PR #1385, opened 2026-08-10
Assorted features

adfriz:
This is better version of #1382 

This commit _should_ be able to cherry picked easily.

I think this commit should be open, so if there any small easy to add feature that i can do, i will commit it here.


leezer3:
Thanks, dropped in.

I'll push a fresh build to the main site later, so we've got something with the .dll.config file back again.

PR #1386, opened 2026-08-10
Build issues

leezer3:
This PR is an attempt to fix the fact that the OpenTK.dll.config file seems to have stopped copying across, and some other intermittant Azure build glitches.

Changes:
* Adds an explicit NuGet reference to the NetFramework reference assemblies. (Intermittant build glitch on Azure, JetBrains Rider also seems to need this under Debian...)
* Uses a single copy of OpenTK.dll.config in the LibRender2 project. Sets the CopyAlways flag on this (as opposed to the previous not specified)
* Adds an explicit copy step in the makefile for all branches for the .dll.config file; this ought to be unnecessary, but let's go belt and braces here.


leezer3:
Now appears to be copying OK.

No idea why this decided to throw a wobbly now...

Issue #1373, opened 2026-08-02
Incorrect rotation in v1.14.0.1


mc0406:
### Description

Object is not rotated when there is an extra comma at the end of the rotate command.

### Reproduction

<img width="64" height="64" alt="Image" src="https://github.com/user-attachments/assets/409a873c-ca55-4a25-b97b-f5669e38c82a" />

[Wheel.csv](https://github.com/user-attachments/files/30631034/Wheel.csv)

open the above files in object viewer or the main game. 

<img width="2560" height="1600" alt="Image" src="https://github.com/user-attachments/assets/c4db33f5-9f8e-4d49-8939-473988402c13" />
actual result in object viewer of v1.14.0.1. No error message is generated

<img width="2560" height="1600" alt="Image" src="https://github.com/user-attachments/assets/eedcad06-4634-4429-9213-c47a6bf7fac4" />
correct result in object viewer of v1.13.0.8.

### Related information 

* Operating system: Windows 11
* Method of control: Keyboard

mc0406:
P.S. The WASD keys are not working in object viewer. I believe this is a very very old issue.

mc0406:
Problem persists in v1.14.0.3. Translation is broken as well.

leezer3:
That's odd. I'm sure I checked that object. Will take a look tomorrow

adfriz:
huh, this works on object viewer...

i have tested both debug and release builds of latest master.

<img width="1015" height="633" alt="Image" src="https://github.com/user-attachments/assets/3df0e0b0-8736-4aa7-ad49-4109c0564a60" />

adfriz:
on v1.14.0.3 release build also looks ok on me.

@mc0406 perhaps you use installer version? maybe try uninstalling and reinstalling it could fix it?

mc0406:
Tested the wheel csv object after reinstalling, it is ok.
However, b3d format is not ok. Here is an example

<img width="2560" height="1600" alt="Image" src="https://github.com/user-attachments/assets/67b44586-0531-4b4d-b03c-c2e4f48467dc" />

On the left is the original file, on the right with all trailing commas removed.
It is also weird that the total face number is the same but alpha, opaque face counts are different respectively.

https://drive.google.com/file/d/1Xepl3kBK5Q_Ol3wYEl0d7kbAXrG4SRL5/view?usp=sharing
You may check FULL.b3d and FULL No Comma.b3d inside exterior folder.

[...]

PR #1390, opened 2026-08-17
Fix B3D parser: strip trailing commas from section headers

adfriz:
for #1373


leezer3:
Fixed in master- this is very slightly better than yours:
```
command = command.Split(']')[0].TrimStart('[');
```

Splitting on the **]** means we don't need the TrimEnd.

leezer3:
Also will handle if there's any other junk after the command properly

Issue #1402, opened 2026-08-27
Performance drop on dGPU after running openBVE on integrated graphics


mc0406:
### Description

Performance drop on dGPU after running openBVE on integrated graphics. For example ~50fps drop to ~24fps in the same route section and train.
Reinstalling openBVE and restarting pc do not help.
However, clear the openBVE folder in Appdata\Roaming and setup the setting again may solve the issue.

Occuring in latest daily build.

### Reproduction
Play openBVE with integrated graphics > Close it > Play openBVE with dedicated GPU (No openBVE setting is altered).

I tried the process twice and both result performance drop.

### Related information 

* Operating system: Windows 11
* Method of control: Keyboard
* GPUs: AMD Radeon 780M, RTX 5060 Laptop GPU

leezer3:
That's very odd.

There's nothing in the code which will do this directly- My best guess from the limited information is that there's a setting which is only taking effect after a restart of the game, but this *shouldn't* be the case.

adfriz:
AFAIK, nvidia app or nvidia control panel have a feature for setting the app what gpu it should use, maybe you can set that?

it should also have an option to show "use nvidia gpu for this app" in right click menu if im not mistaken, so it force to use the dedicated gpu

mc0406:
I always use the MUX switch to link Nvidia GPU and the laptop monitor directly when playing openBVE with Nvidia GPU. openBVE is forced to use Nvidia GPU under this mode. Also shown in Nvidia GPU activity indicator.

mc0406:
P.S. @leezer3 I really find a setting which is only taking effect after a restart of the game during the first startup of openBVE after installation.
It is the level of anisotropic filitering. Idk this is intentional or a bug.

soyachoin104:
Hi, I noticed I was also getting a big performance drop in the recent builds and I would like to comment here as this seems to be very similar to what I was experiencing. This occurs for me in latest stable but even when switching to 28th August nightly build I still experience this.

For me, despite using an RTX 3070 with 16GB RAM and an i7-12700H, I have noticed with certain high detail routes and trains, the performance when I switch to exterior view drops to single fps in some sections. This was the experience I used to get several years ago when I had a much lower spec computer but I am surprised that the fps drops so much on routes and trains that I thought were fairly decent performing in versions of the game from early this year and now run like I still use my old potato.

I've attached one of the trains for replication purposes. This train used to run fairly decently before but now barely cracks 5fps in the external view. 

[...]

PR #1407, opened 2026-09-01
Try improve gif decoder

adfriz:
WIP, although it works Im pretty sure we can pull down the RAM usage more...

As mentioned on https://github.com/leezer3/OpenBVE/issues/1402#issuecomment-5478917728

Testing with file LCD_Gif.csv from that issue, in train type 280 folder.

Currently loading that csv file can take fixed 430MB~ RAM

Tested with locally release build object viewer:
This one is latest master with the last commit 21f1bb790af49143301dd6411c6acd316e65ce17

image

Pretty crazy for just one file. it was feels like memory leaks or need more aggressive GC

This one is this PR.

image

No visual difference as far as i see.

---

@leezer3 what do you think? 

Sorry the code a bit messy 😅

### DIFF
--- a/source/LibRender2/BaseRenderer.cs
+++ b/source/LibRender2/BaseRenderer.cs
@@ -447,7 +447,7 @@ namespace LibRender2
 			StaticObjectStates = new List<ObjectState>();
 			DynamicObjectStates = new List<ObjectState>();
 			VisibleObjects = new VisibleObjectLibrary(this);
-			whitePixel = new Texture(new Texture(1, 1, PixelFormat.RGBAlpha, new byte[] {255, 255, 255, 255}, null));
+			whitePixel = new Texture(new Texture(1, 1, PixelFormat.RGBAlpha, new byte[] {255, 255, 255, 255}, (OpenBveApi.Colors.Color24[])null));
 			nullDepthMap = GL.GenTexture();
 			GL.BindTexture(TextureTarget.Texture2D, nullDepthMap);
 			GL.TexImage2D(TextureTarget.Texture2D, 0, PixelInternalFormat.DepthComponent16, 1, 1, 0, OpenTK.Graphics.OpenGL.PixelFormat.DepthComponent, PixelType.Float, IntPtr.Zero);
--- a/source/LibRender2/Objects/ObjectLibrary.cs
+++ b/source/LibRender2/Objects/ObjectLibrary.cs
@@ -126,49 +126,28 @@ namespace LibRender2.Objects
 					}
 					else
 					{
-						if (State.Prototype.Mesh.Materials[face.Material].DaytimeTexture != null && State.Prototype.Mesh.Materials[face.Material].DaytimeTexture.Origin != null)
-						{
-							// Have to load the texture bytes in order to determine transparency type
-							Texture daytimeTexture;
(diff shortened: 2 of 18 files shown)


leezer3:
You're playing with a slightly different approach to the one I was (not pushed)- I was thinking a wrapper around a compression stream.

Either way probably works, the other thing I was wondering about is one of the C# libsquish forks compressing the texture to DXT directly.

leezer3:
sample2

Transparency for this is broken.

sample3

Random color glitches.

---

I *think* they'll download as images and work if Github hasn't converted them to something else, but if not I can zip up.

I'll do some digging in a minute. I suspect this is something to do with the restore more from the last frame.

GIFs are nasty things....

adfriz:
That last 2 commits fix memory usage, now if test in object viewer it use _only_ 320MB~. still a lot though...

adfriz:
i will test that 2 gif maybe next hour or two, got something to do.

leezer3:
320mb is far better than 4gb. I wouldn't worry too much; everything is a tradeoff.

Transparency wise, I think the issue is that it's not in the GIF pallette at all- It's just a case of overwriting the pixels of the transparent component with pure zero on output if the transparent condition matches. 
I think what's happening is that you're mapping to the pallette index at all times. I tried bodging another pallette entry which didn't immediately work; I've probably missed something....

adfriz:
Ok i have done some fixing locally, but there still a bug...

for that stickman the transparent works but there a 1 black frame everytime the gif loops.

for the second gif it still glitching...

image

leezer3:
Hmm.. the original decoder is glitching on the second GIF too...

Stickman- the glitch is probably in the last frame, which is all transparent.

adfriz:
> Hmm.. the original decoder is glitching on the second GIF too...
> 
> Stickman- the glitch is probably in the last frame, which is all transparent.

interesting, seems like that second gif has something obscure than common gif file.

anyway, another bug... if i change the texture filtering the gif will stop playing and wont play again even when i reload the object viewer with F5

adfriz:
oh another one... object viewer seems not immediately flush loaded object if i press delete key for unload, the ram usage still steady around 320MB even after 1 minute.

can you test it on your side?

adfriz:
Finally it done...

i hope there is no another obscure gif...


https://github.com/user-attachments/assets/f6f9382b-c4c5-489b-bcbc-16e0c1c93d20


https://github.com/user-attachments/assets/64832923-88d4-4beb-9cef-a192dab0e18a


https://github.com/user-attachments/assets/e9ac4645-385b-474b-bb74-6b938758e12b


https://github.com/user-attachments/assets/3e660df7-971d-4017-9229-a8dc593615d8

Also i have fix the object viewer disposal when pressing delete key, this should clear the ram usage directly without waiting the auto GC call

adfriz:
@leezer3 I still dont know the _easy_ way for minimizing ram usage more..

about the library idk how hard to implement it but i found another lib BCnEncoderNet that may still "active" development

leezer3:
Merged.

With C# compression at runtime, the issue is always going to be that you're doing two operations, the decode from GIF then encode back into DXT or whatever, which is going to take time, although I suppose we're halfway there with re-encoding into the indexed form. 
This works well enough to be good, although I'll test on the laptops, which may be where the cleverness with subpixels falls over- can always detect that and fall back to the legacy path though.

adfriz:
Thanks.

May merge this one too ? #1396

PR #1408, opened 2026-09-04
Add TriFace, TriFace2 to CSV B3D parser

leezer3:
https://bveworldwide.forumotion.com/t2562-drastic-fps-drop#22526

https://gist.github.com/leezer3/b26f9780a0c740ba8717c37f4a34ee53

The underlying trouble here is in how the exporter structures the model, and ultimately the modelling decisions made by the user.

This is the best way I can think of at the minute to detect a face that has had triangularisation performed on it (exporting a model that has had Blender's Limited Dissolve performed on it also does this).

@adfriz any useful thoughts? 
I suppose we could possibly test for the tri pattern in each set of 6 vertices, but you'll end up running practically the entire itinerator loop on the thing...

The other thing to note is that this is backwards incompatible, which is problematic...
You could I suppose introduce a new extension into the mix to differentiate these files, but either way this feels iffy...

### DIFF
--- a/source/OpenBveApi/Objects/ObjectTypes/StaticObject.cs
+++ b/source/OpenBveApi/Objects/ObjectTypes/StaticObject.cs
@@ -12,7 +12,7 @@ namespace OpenBveApi.Objects
 	public class StaticObject : UnifiedObject
 	{
 		/// <summary>Whether the object is optimized</summary>
-		private bool isOptimized;
+		public bool Optimized;
 		/// <summary>The mesh of the object</summary>
 		public Mesh Mesh;
 		/// <summary>The starting track position, for static objects only.</summary>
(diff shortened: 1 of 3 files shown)


adfriz:
honestly idk which way the best... personally, check for a pattern like the vertex or face maybe the way?

also from my experience with that exporter is kinda mixed...

one way for the 3d modeller that could make parser and the optimizer read it _easily_ is to make sure all objects triangulated before exported, but the way it triangulated in blender sometimes can or cannot trigger the optimizer.

But there also some behavior like if i make a 1 mesh that has a lot of tris, let say 800K the exporter could just write it as 1 giant meshbuilder block. i think that will choke the renderer if the optimizer not triggered.

leezer3:
https://blender.stackexchange.com/questions/69081/limited-dissolved-over-thousands-of-objects

I think we can use the code here in the blender exporter in combination with this.

This will then get Blender to triangulate correctly, *and* importantly the new commands then mark the mesh as tris.

PR #1409, opened 2026-09-04
Render custom fonts in the route map view

zeerooth:
Always bothered me that it wasn't a thing. Also, without the custom font selection I couldn't get Japanese characters to render on the route map in WINE, even with CJK fonts installed. I suppose it could be a bug in how the GenericSansSerif font is selected in .NET? Anyhow, custom font rendering fixes that.

Please let me know if I missed something.

### DIFF
--- a/source/LibRender2/Text/Fonts.cs
+++ b/source/LibRender2/Text/Fonts.cs
@@ -25,6 +25,8 @@ namespace LibRender2.Text
 		/// <summary>Represents the largest sans serif font.</summary>
 		public readonly OpenGlFont EvenLargerFont;
 
+		public readonly FontFamily uiFont;
+
 		private static HostInterface currentHost;
 
 		/// <summary>Gets the next smallest font</summary>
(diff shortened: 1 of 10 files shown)


leezer3:
https://github.com/leezer3/OpenBVE/blob/master/source/OpenBVE/System/Options.cs#L188

It's going to be after the weekend before I get to look at this, but that is the logic for default UI font selection.

Apple needs a different font to get working non Western glyphs too...

zeerooth:
> https://github.com/leezer3/OpenBVE/blob/master/source/OpenBVE/System/Options.cs#L188
> 
> It's going to be after the weekend before I get to look at this, but that is the logic for default UI font selection.
> 
> Apple needs a different font to get working non Western glyphs too...

Yeah that font selection algorithm makes sense as the font rendering works for most of the UI on WINE with Noto Sans CJK JP Regular installed, but the issue is that RouteMap rendering code uses FontFamily.GenericSansSerif. I can't find what's the logic behind GenericSansSerif's font selection but clearly it may choose a font that not necessarily supports the glyphs we need. Besides, it's always a good idea to give the user a choice I think.

zeerooth:
I also changed Timetable's code to render user's custom UI font

leezer3:
Thanks.

I've tweaked it slightly to use the underlying font in the normal font so we don't create another memory object lying around, but that's minor to say the least.

PR #1417, opened 2026-09-12
Implement async load for route objects

adfriz:
Route loading should covers both .csv and .rw . it was spent most of its time decoding object files one by one on the loader thread: every `Structure.*` declaration called `Host.LoadObject()` inline before parsing continued.

This PR collects all `Structure.*` declarations (Rail, Walls, Dikes, Forms, Roofs, Cracks, FreeObj, Beacon, Poles, Ground etc..) during parsing, decodes them concurrently with `Parallel.For`, then commits the results sequentially in file order.

Loading behaviour is unchanged in game, so there should be no object popping in game because it loaded late.

Worker count is automatic: `CPU <= 4 ? CPU - 1 : min(8, CPU)`, reserving one thread for the loading screen on small machines.

Parsing the CSV/RW text is cheap, file I/O + object parsing + texture decode is the actual bottleneck. On high multi-core machines the object phase speeds up roughly with core count (currently capped at 8 logical core not physical core count, although disk/RAM speed may become the limit).

Thread-safety hardening

Shared state that was previously only touched single-threaded is now guarded:

- `HostInterface`: add private cache lock with `Store`/`Report`/`Prune`/`Clear` helpers, the cache hits are cloned outside the lock; cache inserts upsert instead of throwing on duplicates.
- `Host` (main program, ObjectViewer, RouteViewer): failure/missing-file sets updated atomically, messages emitted outside locks.
- `TextureManager`: texture decode happens outside the lookup lock with double-checked registration, all allocator paths and readers (load/unload/stats) are also locked.
- Log message lists (all three apps): writes locked, readers use snapshots.

Testing

- Needs a cold-load run on a heavy route, the first load timing is what we want, route reload through F5 should still faster.
- Not yet tested on Linux/macOS
- Was tested with shinkansen route v1.00 beta EN in route viewer, file `1fine-Hayabusa1v1.00.csv `. The load is around ~11s with reload is ~4s
- Route viewer settings should be minimal, no shadow, texture filtering is bilinear with only 1 level, vsync on, object optimization is on high mode.


adfriz:
wait... this messing with the $include route command...

i got plenty of error rail structure index out of range in `Cycle.Rail` command

well... let me set is as draft...

adfriz:
@ginga81 do you want to try this?

in route viewer try use minimal testing settings as i written above.

adfriz:
alright, i think that one will fix it. although on my test its around ~4s slower... now first load is ~15s

well... at least better than wrong looking route.

adfriz:
Ready for review.

ginga81:
All CPU cores are active, and I feel like loading times have become incredibly fast!
It is running on Linux (Ubuntu 24.04 + X11/KDE).
I haven't tweaked any settings; I'm using my usual configuration, including maximum graphics quality.

I'll open a separate issue for this later, but even when shadows are turned off, the brightness doesn't behave as it used to—it stays bright even when entering a tunnel.
すべてのCPUが動作

adfriz:
ok i have added similar async load for object viewer.

currently only for b3d and csv mesh. for .animated mesh need more test although it works pretty well...

Note: 
i add progress bar option for object viewer, same as in the route viewer. But i also add text based progress in the window name if extensions.cfg is loaded, it should be like `Object viewer - Progress 0/10`

i dont know if that route viewer like progress bar is needed here... maybe i will delete it and leave the window name progress only.

leezer3:
X files are broken somehow,

bug

I'm not entirely certain why, not got time to debug this tonight, but at a probable guess, your material cache is broken- Hitting the last loaded object's materials or something?

[postnumber2_5.zip](https://github.com/user-attachments/files/32168209/postnumber2_5.zip)

Definitely no sign of schwellen4.bmp in there.

bug2

Randomly missing textures, very probably related to the above. These are stable throughout a reload, but if I close Route Viewer and re-open, different missing textures- I'd presume this is because the reload simply hits the cache.

adfriz:
is there a link for that route?

adfriz:
That commits may work...

adfriz:
another fix and ram usage improvement. i hope its working on linux and macos...

leezer3:
That looks better now.

I'll try and test on Mac, but might be a few days.

ginga81:
It seemed to work fine on my Ubuntu 24.04.
ObjectViewer

ginga81:
Screenshot from 2026-09-14 18-30-04
However, the scene redraws when I change direction, and the frame rate subsequently drops to 10 fps.
It redraws the entire scene into the distance every time I change direction, and the frame rate does not recover afterwards.

adfriz:
> Screenshot from 2026-09-14 18-30-04 However, the scene redraws when I change direction, and the frame rate subsequently drops to 10 fps. It redraws the entire scene into the distance every time I change direction, and the frame rate does not recover afterwards.

The opaque faces seems too much, did you didn't set the object optimization mode to high?

ginga81:
I wondered what "object optimization mode" was, so I checked the options and saw it was indeed set to "Low"...
Changing it to "High" restored things to how they were before.
Sorry for not understanding it sooner.
Screenshot from 2026-09-14 20-31-00

adfriz:
alright, the commits from the DDS parts is mostly for BVE5/6 routes. now it should as fast as openbve routes.

There is still odd fps drop bug when turning on shadow feature when loading bve5/6 routes... idk what it cause....

leezer3:
Apple appears to be OK, but I haven't done much testing on this at the minute. 

Viewers seem to be broken on Apple though:
```
System.ApplicationException: ERROR: 0:25: '' :  version '410' is not supported
ERROR: 0:25: '' : syntax error: #version
ERROR: 0:26: 'precision' : syntax error: syntax error

  at LibRender2.Shaders.AbstractShader.LoadShader (System.String shaderSource, OpenTK.Graphics.OpenGL.ShaderType shaderType) [0x0005d] in /Users/christopher/RiderProjects/OpenBVE/source/LibRender2/Shaders/AbstractShader.cs:114 
  at LibRender2.Shaders.AbstractShader..ctor (LibRender2.BaseRenderer renderer, System.String vertexShaderName, System.String fragmentShaderName, System.Boolean isFromStream, System.Boolean fragColor) [0x0005b] in /Users/christopher/RiderProjects/OpenBVE/source/LibRender2/Shaders/AbstractShader.cs:59 
  at LibRender2.Shaders.Shader..ctor (LibRender2.BaseRenderer Renderer, System.String vertexShaderName, System.String fragmentShaderName, System.Boolean isFromStream) [0x00000] in /Users/christopher/RiderProjects/OpenBVE/source/LibRender2/Shaders/Shader.cs:79 
  at LibRender2.BaseRenderer.Initialize () [0x00010] in /Users/christopher/RiderProjects/OpenBVE/source/LibRender2/BaseRenderer.cs:391
```

That's somewhat odd, as it's just the default shader which is used on the main game which works OK :/
Maybe the requested context has got tweaked or something. I'll take a better look on Weds...

adfriz:
That fix might work?

leezer3:
Merged, but I've just found this breaks the Object Viewer auto-reload. Might have to disable async for that?

adfriz:
oh yeah its broken... but only when loading extensions.cfg

i will trace back, seems easy enough to fix

leezer3:
Definitely broken with just a plain animated file. This is what I noticed it with:
```
[Object]
States=bogie_commonwealth.b3d, bogie_commonwealth.b3d, wheelset_solid.b3d
StateFunction = TrackNumber[0] + 1
Position=0,0,0
```

(Added a couple of minor functions this morning, was just using a random couple of files to test them with)

Timing related I'd assume.

adfriz:
another minor bug, auto load not triggered when changing something inside .animated file that referenced in current loaded extensions.cfg, need manual reload.

but changing someting in extensions.cfg works as expected.

i think that bug not a big deal?

adfriz:
can you test with #1421 ?

PR #1419, opened 2026-09-13
Show referenced meshes when train.dat is missing

adfriz:
When loading extensions.cfg if no train.dat present allow it to be loaded. the train cars will stacked at the origin, its better than showing nothing. Also write info message telling player the train.dat missing cause this.

Example train: E5 ver2.00a

http://midnightexpginga81.web.fc2.com/touhoku_shinkansen/download-touhoku.html

### DIFF
--- a/source/ObjectViewer/ProgramS.cs
+++ b/source/ObjectViewer/ProgramS.cs
@@ -392,8 +392,50 @@ namespace ObjectViewer {
 
 					    if (!trainLoaded)
 					    {
-							//As we now attempt to load the train as a whole, the most likely outcome is that the train.dat file is MIA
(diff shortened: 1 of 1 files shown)


ginga81:
I think that as you are right; the "no plugin data" error message alone wouldn't lead we to realize that `Train.dat` is missing.
Since these are distributed as "outer views" E5's, they are designed to be integrated into the base cab data.
That is why there is no `Train.dat` file included.
The setup requires the `Train.dat` file from the F81_tec200's cab data.

leezer3:
X is missing from the static mesh list :)

(Although, looking at animations for the X format is something else somewhere on a todo list....)

PR #1421, opened 2026-09-16
Fix: infinite auto-reload loop and make empty load timer message when nothing loaded in ObjectViewer

adfriz:
CheckFileChanges only queues PendingReload on actual file change instead of every frame while loading, fixing infinite reload with extensions.cfg when auto-reload is on. Skip RefreshObjects when no files are loaded to avoid spurious synchronous load message on startup.


leezer3:
Nope, same issue with my minimal test file.

To be exact. the error list is as follows:
```
Error		Plugin Object.Animated.dll returned unsuccessfully at LoadObject
Error		Plugin Object.DirectX.dll raised the following exception at CanLoadObject:The process cannot access the file 'D:\Program Files\BVE\Train\D1015\Stock\bogie_f.animated' because it is being used by another process.
Error		Plugin Object.Msts.dll raised the following exception at CanLoadObject:The process cannot access the file 'D:\Program Files\BVE\Train\D1015\Stock\bogie_f.animated' because it is being used by another process.
Information		Objects loaded in 44 ms (synchronous) | textures: 0 ms decode (0 calls) / 0 ms upload (0 calls).
```

Object.Animated catches the exception that the file is being used by another process, and returns false (same exception as the others, just doesn't seem to be logged)

None of the others could load the file anyways...

I wonder if it'd be better to add a sleep-retry to things. Couple of hundred milliseconds would probably do it..

leezer3:
OK, so see what you think of the commit I've added above.
Simple 100ms sleep if we catch an exception in CanLoadObject, and then retries,

Can't easily alter the Object Interface (as nautilus has it's own object loader plugin), and adding the same logic to the LoadObject function risks multiple exceptions being added for the same file, e.g. non-matching number of face vertices in the X parser.

adfriz:
Thanks, yeah the retry approach makes sense. 

My commits also fixes some stuff, Msts should no longer throws on locked files and stops leaking its stream, the Animated opens read/shared instead of exclusive, the counters went local so parallel loads can't share them, and Animated LoadObject retries sharing violations only so parse errors will show correctly. I also dropped my strict .x gate to preserve the alias support. 

Could you try again with auto-reload on?

leezer3:
Will do later.

HResult will be unreliable under anything other than Windows though:
https://github.com/dotnet/runtime/issues/25998

adfriz:
The ci pass, idk will that work on real mac.

i will revert it if that not working

PR #1422, opened 2026-09-17
Fix: high CPU when VSync is ON with a stored FPS limit

adfriz:
it was bug in this scenario:

1. vsync on + fps disabled.
2. vsync off + fps set to other like 120.
3. vsync on again + fps now looks disabled but internally still use the last limit and not disabled properly.


leezer3:
Better I think to fix that in our OpenTK fork. (second attempt as first tree had a debug commit left in...)

The linked commit will do that, and once I've got the thing built (absolute pain) and uploaded to NuGet, I'll submit a PR that cab be tested.

leezer3:
https://github.com/leezer3/OpenBVE/pull/1423

PR with updated OpenTK version. Shouldn't have any other effect.

leezer3:
Merged that PR, which should resolve this more cleanly.

PR #1423, opened 2026-09-17
Change: Bump OpenTK version

leezer3:
Closes https://github.com/leezer3/OpenBVE/pull/1422

Adds this change to our OpenTK fork directly, rather than needing to set in-game.


leezer3:
That builds and doesn't appear to explode, when all the csproj files actually get updated....

The underlying change is trivial enough that this by itself should be OK.

Might also want to take https://github.com/opentk/opentk/issues/1329#issuecomment-1506570216 if we're poking in the OpenTK internals.
Should have thought of that before pushing to Nuget, but oh well....

